PR #13, opened 2017-07-20
Correct version for Temporal Table Introduction

scottsauber:
SQL Server 2016 introduced Temporal Tables, but the current documentation is pulling the latest version which is SQL Server 2017.  This conflicts with the "This Topic Applies To" at the top of the article.

image

[Here's what live currently looks like before this change](https://docs.microsoft.com/en-us/sql/relational-databases/tables/temporal-tables)

If you'd rather me put it in a different branch than live, let me know.

Thanks!


craigg-msft:
@scottsauber  thank you!

PR #16, opened 2017-07-21
gramma fix at line 188

kevintan1983:
From: An example of a for the ADO.NET provider ...
To: An example of the ADO.NET provider ...


craigg-msft:
Thank you for your time. Merging...

PR #20, opened 2017-07-25
that should be plural

tugberkugurlu:


craigg-msft:
Thank you @tugberkugurlu for your contribution!

PR #25, opened 2017-07-26
Fixed string_agg syntax

jnm2:
If I'm understanding correctly, this looks like a typo. `string_agg(Foo, ', ') within group (order by Foo)` works, `string_agg(Foo, ', ' within group (order by Foo))` does not.


JennieHubbard:
@jnm2 Thank you for correcting that!
#sign-off

PR #28, opened 2017-07-28
Update data-tier-applications.md

aharpervc:
- fix indentation such that the bullet point isn't treated as a block quote


craigg-msft:
Thank you @aharpervc for your time and contribution.

PR #29, opened 2017-07-28
fix list item in cache-shared-datasets-ssrs.md

NReilingh:
List "header" was formatted as item in previous list


craigg-msft:
Thank you @NReilingh for your contribution.

NReilingh:
@craigg-msft Is there any reliable way of previewing the microsoft's markdown rendering before a commit? That didn't quite do what I expected -- it removed the list item "disc", but didn't de-indent...

Oh wait, now I see. All of these lines end with two spaces, which I think create line breaks instead of paragraphs. That's awful. I wouldn't want to attempt fixing them unless I knew what the output was going to look like, or if there was a microsoft docs markdown style guide somewhere to follow. Can you point me in the right direction?

PR #31, opened 2017-07-31
Update insert-sql-graph.md

seb-dataworks:
Clarified description on how to bulk insert an edge table.


craigg-msft:
Thank you @seb-dataworks.

PR #40, opened 2017-08-05
Update configure-the-recovery-interval-server-configuration-option.md

rabryst:
Looks like a link was inserted and the original holding text wasn't removed (end of first para for "automatic checkpoints" -- not sure why GitHub is suggesting replacing the entire section).


JennieHubbard:
@rabryst Thank you! Good catch!! Interesting that GitHub suggested replacing the whole section. We will merge this to Live later today or tomorrow.

JennieHubbard:
#sign-off

PR #41, opened 2017-08-05
Update sql-server-transaction-log-architecture-and-management-guide.md

rabryst:
Fixing Paul's surname


craigg-msft:
Thank you @rabryst

PR #43, opened 2017-08-08
Correct T-SQL statement for creating the AG

wsmelton:
- Statement for COMPUTER01 had syntax error
 - LISTENER statement was missing alter command and cannot be part of the same batch to create the AG itself.


JennieHubbard:
@wsmelton Thank you for your contribution!
#sign-off

PR #44, opened 2017-08-08
Use Join-Path in Powershell script

kalamajka:
Use Join-Path to produce the file path in the PowerShell script instead of string concatenation.


JennieHubbard:
@kalamajka Thank you for your contribution!

#sign-off

PR #45, opened 2017-08-09
Add MaxActiveReqForAnonymous

ericgu:
We added a new configuration setting for PBIRS named MaxActiveReqForAnonymous. This commit adds documentation for the new setting


barbkess:
@ericgu Thanks for making this change.  We will push it live.

PR #46, opened 2017-08-09
Removing extraneous word from column description

bgribaudo:


barbkess:
@bgribaudo Thanks for fixing this!

PR #48, opened 2017-08-10
clarified where to look for the refresh button.

MarkIannucci:
I didn't realize I was looking for it in the designer; hopefully my additions help.


barbkess:
@MarkIannucci Thank you for pointing this out and fixing it.  We really appreciate it.

PR #49, opened 2017-08-10
Fix small typo

justinsg:
Fixed a small typo in _Getting Started with Database Engine Permissions_:

...the sysadmin fixed server role, and making adding a login to the sysadmin fixed server role...
...the sysadmin fixed server role, and adding a login to the sysadmin fixed server role...


barbkess:
@justinsg Thanks so much for pointing this out and fixing it. We really appreciate it.

PR #50, opened 2017-08-10
Fix Typo in distributed-availability-groups.md

gbrayut:


barbkess:
@gbrayut Thanks so much for fixing this.  We really appreciate it!

PR #52, opened 2017-08-12
Syntax and formatting fixes

krislawton:
Corrected syntax typo in lead. Converted result sets to multiline code elements so they are more aligned and consistent with other Docs pages


barbkess:
@krislawton Thanks for fixing this.  We will push it to live.

PR #53, opened 2017-08-14
Update sql-server-2017-release-notes.md

jonathanpotts:


barbkess:
@jonathanpotts Thanks for making this change.  We will push it to the live site.

PR #57, opened 2017-08-22
Update install-analysis-services.md

lianghong42:
misspelling


rothja:
@lianghong42 thanks for this correction!

PR #58, opened 2017-08-22
fix formated value typo

TomVergote:
FORMATED_VALUE should be FORMATTED_VALUE


rothja:
@TomVergote thanks for this correction! We'll publish it.

PR #62, opened 2017-08-25
Update sys-column-store-segments-transact-sql.md

kfarlee:
Correct an error in the TSQL example, as proposed by Sudin


rothja:
@kfarlee Thanks for the correction.

PR #63, opened 2017-08-25
Update file-locations-for-default-and-named-instances-of-sql-server.md

KennieNP:
Fixed table - configurable/fixed path column was not formatted correct


rothja:
@KennieNP Thanks for the correction!

PR #65, opened 2017-08-28
Improve sample code to indicate a range is created

SvenAelterman:
Incrementing the firewall end IP address by just 1 may create a false impression that they are individual IPs rather than a range. The description wasn't helpful either.
Also removed "Windows" from "Windows Azure": It hasn't been called that in a while I think. 
Also added some spaces between stored procedure parameter values, to be consistent with T-SQL docs (https://docs.microsoft.com/en-us/sql/relational-databases/stored-procedures/specify-parameters)

### DIFF
--- a/docs/relational-databases/system-stored-procedures/sp-set-database-firewall-rule-azure-sql-database.md
+++ b/docs/relational-databases/system-stored-procedures/sp-set-database-firewall-rule-azure-sql-database.md
@@ -56,33 +56,33 @@ sp_set_database_firewall_rule [@name = ] [N]'name'
  The following table demonstrates the supported arguments and options in [!INCLUDE[ssSDS](../../includes/sssds-md.md)].  
   
 > [!NOTE]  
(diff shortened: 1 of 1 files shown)


craigg-msft:
Thank you @SvenAelterman  for the contribution to content.

craigg-msft:
@BYHAM please review.

BYHAM:
Thank you for submitting this. @SvenAelterman Didn't you include these #65 changes in your #66 submission. (Which I have accepted.) If you agree, please cancel this one (#65).

SvenAelterman:
@BYHAM: note quite, there are two different articles, one for server-level and one for database-level. So yes, I made almost the same changes but that's because I believe both articles could be improved by making these changes.

BYHAM:
@SvenAelterman Ah. I didn't spot that right away. Thank you. Yes, I agree with the changes in the second topic, too, and I have accepted them.

BYHAM:
#sign-off

BYHAM:
@craigg-msft Was not a duplicate after all. Just similar.

PR #66, opened 2017-08-28
Update sp-set-firewall-rule-azure-sql-database.md

SvenAelterman:
Updated Permissions section to indicate that an Azure AD user assigned as the admin or member of an Azure AD group assigned as admin can also create and update rules.
Incrementing the firewall end IP address by just 1 may create a false impression that they are individual IPs rather than a range. The description wasn't helpful either.
Also remove curly quotes from parameter values.
Also added some spaces between stored procedure parameter values, to be consistent with T-SQL docs (https://docs.microsoft.com/en-us/sql/relational-databases/stored-procedures/specify-parameters)


craigg-msft:
@SvenAelterman Thank you for the contribution.

craigg-msft:
@BYHAM please review.

BYHAM:
#sign-off

PR #68, opened 2017-08-28
Add that Azure AD principal assigned as server admin can also modify …

SvenAelterman:
…rules

Also fix curly quotes to ` for inline code.


craigg-msft:
Thank you @SvenAelterman . We appreciate your contribution.

PR #72, opened 2017-08-30
Updating table formatting

tlmii:
A couple of the tables had markdown that wasn't rendering properly.


craigg-msft:
Thank you @tlmii for your help with the content.

PR #73, opened 2017-08-30
Update columnstore-indexes-design-guidance.md

GeneRPool:
Changed duplicated word "decision" to "design".


craigg-msft:
Thank you @GeneRPool for the help with the content! good catch 👍

PR #75, opened 2017-08-31
Fix incorrect comment/T-SQL code

LowlyDBA:
Comment and T-SQL enable the job status, but the new job name indicated it was disabled. 

Fixed T-SQL comment and @enabled flag to indicate the job is disabled, matching the new job name.


craigg-msft:
Thank you @LowlyDBA for the help with the topic.

PR #77, opened 2017-08-31
Fix typo

eduardok:
Replaced DB)ID with DB_ID.


craigg-msft:
Thank you @eduardok for the help with the content.

PR #78, opened 2017-09-01
Update sql-server-linux-configure-docker.md

sfrechette:
Section ### Mount a host directory as data volume ->  [!IMPORTANT] 
"...This restriction is specific to the /var/opt/msql directory. " replaced msql with mssql


craigg-msft:
Thank you @sfrechette for the help with the content.

PR #79, opened 2017-09-01
Fix acronym in title

tgerstendoerfer:
SSDT instead of SSRS


craigg-msft:
Thank you @tgerstendoerfer . 1 letter that makes a big difference 👾

PR #80, opened 2017-09-04
Fix highlighted value for $node_id

greglow:
The highlighted value currently says $node_ids which is not correct.


craigg-msft:
Thank you @greglow for the content contribution!

PR #81, opened 2017-09-05
64-bit path and removed version specific number

Matticusau:
Added the path for C:\Program Files (x86).... as that is used for 64-bit installs. Also changed the numeric version specific folder of 130 to ### to be more version agnostic


craigg-msft:
Thank you @Matticusau for the content contribution!

PR #82, opened 2017-09-05
System table naming

greglow:
Even though they are headings here, the system tables are all defined as lower case and probably should be here too.


craigg-msft:
Thank you @greglow for the help.

PR #84, opened 2017-09-05
Updated trace flag 610 description

savjani:
Updated trace flag 610 description to include the changes in SQL Server 2016 where the trace flag is not required.


savjani:
#sign-off

craigg-msft:
Thank you @savjani

PR #87, opened 2017-09-11
SQL authenticated procedure call not permitted

greglow:
Hi Folks, this document says that members of the sysadmin server role can call this procedure. That is not true. It cannot be called by logins authenticated by SQL Server such as the sa login.
Regards,
Greg


craigg-msft:
Thank you @greglow for the content contribution

PR #88, opened 2017-09-11
I noticed some typos and fixed them.

MarkIannucci:
I'm sorry about the wall of pink.  I used GitHub's web based editor to update this file which seems to have caused a line-ending fail.


craigg-msft:
Thank you @MarkIannucci  for the help with the content.

PR #90, opened 2017-09-11
Fix indentation

SvenAelterman:
Fixing indentation on the last concept following a bulleted list


craigg-msft:
Thank you @SvenAelterman for the content help.

PR #96, opened 2017-09-14
make clear that outputsqlerrors option requires an argument

johntiger1:


craigg-msft:
Thank you @johntiger1 for the content help.

PR #99, opened 2017-09-16
Update sys-sp-drop-trusted-assembly-transact-sql.md

greglow:
Suspect this was a copy/paste error


craigg-msft:
Thank you @greglow  for the update. merging

PR #101, opened 2017-09-17
Just fixing capitalization

greglow:
Two of these looked wrong


craigg-msft:
thank you @greglow for the help.  merging.

PR #103, opened 2017-09-17
Use fenced code blocks instead of `quoting`

theunrepentantgeek:
Plus a small number of other Markdown syntax changes


craigg-msft:
Thank you @theunrepentantgeek for the content help!

PR #104, opened 2017-09-18
updated - with_clause text

DanielSSilva:
Provided a (hopefully) clear explanation about the "matches keys" regarding the columns and keys matching. SQL usually is case insensitive, but in this case it's not. I've been debugging why I was receiving nulls, and it was because my object properties start with capital letter.


craigg-msft:
Thank you @DanielSSilva for the help!

PR #109, opened 2017-09-22
Move 'U' and 'V' object types out of 2012+ version

mloskot:
The two markers are also available in SQL Server 2008.
It is unclear why they were listed in the section that applies to versions from SQL Server 2012 through 2017 only.


craigg-msft:
Thank you @mloskot for the help with content.

PR #112, opened 2017-09-25
Change "timestamp" to "rowversion"

SvenAelterman:
That's how it's referred to everywhere else. There is only a brief mention of "timestamp" on the rowversion page.


craigg-msft:
Thank you @SvenAelterman for the help.

PR #115, opened 2017-09-26
Update back-up-the-service-master-key.md

rabryst:
The Service Master Key shouldn't need to be in any user database context, and the comment relating to the password protection seems to have been copied from the code example on backing up the Database Master Key.

(Once again the entire section seems to want updating for a couple lines of example code. Weird.)


craigg-msft:
Thank you @rabryst for the help.  @BYHAM can you please review.

BYHAM:
#sign-off

PR #117, opened 2017-09-26
Fix Query Store stored procedures display in SQL 2016 release notes

scottaddie:
Fix the code fencing markup to address a display issue.


craigg-msft:
Thank you @scottaddie

PR #119, opened 2017-09-26
Update row-level-security.md

rabryst:
"You may define a predicate function that joins with another table..." is repeated in the next line, but without the `SCHEMABINDING` information. It seems like a duplicate sentence.


craigg-msft:
Thank you @rabryst for the help.

PR #122, opened 2017-09-28
Remove reference to selecting a rule in the Default section

SvenAelterman:
Rule has its own section, in the correct place alphabetically.


craigg-msft:
Thank you @SvenAelterman

PR #123, opened 2017-09-28
windows docker image is wrong

options:
windows docker image name has to be changed.

### DIFF
--- a/docs/linux/quickstart-install-connect-docker.md
+++ b/docs/linux/quickstart-install-connect-docker.md
@@ -73,7 +73,7 @@ The following steps increase the memory for Docker for Windows to 4 GB.
     If you are using Docker for Windows, use the following command from an elevated PowerShell command-prompt:
 
     ```PowerShell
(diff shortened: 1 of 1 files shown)


craigg-msft:
Thank you @options   @rothja can you please do a quick review.  Thanks!

rothja:
@options The quickstart is actually using the Linux container on both Linux and Windows for consistency in the walkthrough. I'll look at how we can be more consistent in explaining this decision for the tutorial. I could possibly add a note that explains how to create the windows image instead. But this line is actually as intended here. Let me know if that makes sense, and thank you for the feedback!

craigg-msft:
@rothja Should I close this PR?

rothja:
@craigg-msft I think we should close it, because the changes needed are to explain Windows containers better and to make this topic more clear that it is using a Linux container. But the original change is not needed in a Linux-only container tutorial. Thanks.

craigg-msft:
closing

PR #126, opened 2017-09-28
Update sys-computed-columns-transact-sql.md

sikpigs:
Fixed typo - sys.computer_columns changed to sys.computed_columns.


craigg-msft:
Thank you @sikpigs for the help.

PR #129, opened 2017-09-29
Fix example A and example output to be table

SvenAelterman:
Added [] around CURRENT_TIMESTAMP, otherwise the query syntax is not accepted (by SQL 2017) at least).
Also added some spaces after the commas separating columns.
PS: I only modified example A's output to be in table format. Not sure if it was desired, but it seems to make sense to me to match the output to what the query would return. Otherwise maybe modify the query to have multiple `SELECT` statements?


craigg-msft:
Thank you @SvenAelterman   @BYHAM please review.

SvenAelterman:
@BYHAM Sure, will take me a few hours though

BYHAM:
#sign-off

craigg-msft:
Thank you @BYHAM  and @SvenAelterman

PR #130, opened 2017-09-29
Correct wording for converting float to decimal

SvenAelterman:
Not all numbers with precision > 17 will round to zero, only when the number has all zeroes until the 17th digit (from the left).


craigg-msft:
Thank you @SvenAelterman

PR #135, opened 2017-10-02
Format the "important" box

andresnava:
Update tag so that the box with appropriate styling is used in the "important" section. Currently, box and styling are not being rendered correctly


craigg-msft:
Thank you @andresnava for the updates.

PR #137, opened 2017-10-03
[SQL Server 2017, Graph processing] match the title with the one inside the article

tugberkugurlu:
not sure if that was intentional to keep them separate.


craigg-msft:
Thank you @tugberkugurlu

PR #140, opened 2017-10-04
Fixing broken prev/next links

bgribaudo:


craigg-msft:
Thank you @bgribaudo for the help.

PR #144, opened 2017-10-05
Formatting of partial result sets fixed

garethstephenson:
The latter examples were incorrectly formatted, now fixed.


craigg-msft:
Thank you @garethstephenson for the help.

PR #146, opened 2017-10-05
Update sys-dm-os-sys-info-transact-sql.md

savjani:
Updated Applies to for columns sql_memory_model and sql_memory_model_desc to include SQL 2012 SP4


savjani:
#sign-off

savjani:
#signoff

PR #147, opened 2017-10-06
Fixing syntax specification

bgribaudo:
In the existing specification:

- Some string parameter values require Unicode values but are missing the `N` prefix (e.g. `N'value'`) 
- The square brackets used to indicate optional parameters are incorrect.

This pull request addresses both of these issues.


craigg-msft:
Thank you @bgribaudo  for the help.  @edmacauley can you please do a quick review.

bgribaudo:
@edmacauley, while you are looking at this page, would you be the right person to add a missing argument description? The *Arguments* description list doesn't say anything about `RESULT SETS`.

edmacauley:
#sign-off looks good

PR #149, opened 2017-10-09
Updated code samples to use the t-sql highlighter

garethstephenson:


craigg-msft:
Thank you @garethstephenson for the help.

PR #154, opened 2017-10-11
Increase accuracy of operator precedence

chrisdone:
AND and OR have higher precedence than IN, LIKE, SOME, etc.

Consider the following example which TSQL accepts properly:

```sql
X = Y AND Z IN (1, 2, 3)
```

With the existing table and precedence numbers, this ought to be parsed as:

```sql
((X = Y) AND Z) IN (1, 2, 3)
```

But with my corrected numbers, it ought to be parsed

```sql
(X = Y) AND (Z IN (1, 2, 3))
```

which I believe is correct:

```sql
1> select 'Correct!' where 3 = 3 AND 7 in (7);
2> go

Correct!
(1 row affected)
1> select 'Correct!' where (3 = 3 AND 7) in (7);
2> go
Msg 4145 (severity 15, state 1) from 9d665ba5181d Line 1:
	"An expression of non-boolean type specified in a context where a condition is expected, near ')'."
````

I only noticed this as I'm writing a parser for the TSQL language and using this table tripped me up.


craigg-msft:
Thank you @chrisdone for the help.  @edmacauley please review.

edmacauley:
Hi @chrisdone thanks for writing.  The operator precedence information currently on the site is accurate.  As you noted, AND is parsed before the IN clause.

craigg-msft:
per Ed, closing this PR.

PR #158, opened 2017-10-13
Removes broken legacy "Applies to" section

exchange12rocks:


craigg-msft:
Thank you @exchange12rocks

PR #161, opened 2017-10-15
Expanded definitions and updated for SQL Server 2017

srutzky:
Expanded definitions of "CodePage", "LCID", and "ComparisonStyle" properties so that they aren't a mystery to most people who don't know where to get those definitions. I copied the 4 bitmask values for "ComparisonStyle" from [DATABASEPROPERTYEX]( https://docs.microsoft.com/en-us/sql/t-sql/functions/databasepropertyex-transact-sql) but did not copy the "for example" portion as that seemed extraneous. But, if it is best that it be here for consistency then I can add it in.

Regarding the link used for "LCID": the page has the value in hex which requires the extra step of converting the INT value returned here, but that list is by far the most complete that I could find. In fact, I couldn't find any other list that included entries for values 0x00010407 through 0x00040411.

Regarding the addition to the "Version" info: I confirmed that info on SQL Server 2017 as **3** is returned for the new Japanese collations (i.e. the `Japanese_Bushu_Kakusu_140_` series and the `Japanese_XJIS_140_` series).

### DIFF
--- a/docs/t-sql/functions/collation-functions-collationproperty-transact-sql.md
+++ b/docs/t-sql/functions/collation-functions-collationproperty-transact-sql.md
@@ -45,10 +45,10 @@ Is the property of the collation. *property* is **varchar(128)**, and can be any
   
 |Property name|Description|  
 |---|---|
-|**CodePage**|Non-Unicode code page of the collation.|  
-|**LCID**|Windows LCID of the collation.|  
-|**ComparisonStyle**|Windows comparison style of the collation. Returns 0 for all binary collations .|  
-|**Version**|The version of the collation, derived from the version field of the collation ID. Returns 2, 1, or 0.<br /><br /> Collations with "100" in the name) return 2.<br /><br /> Collations with "90" in the name) return 1.<br /><br /> All other collations return 0.|  
+|**CodePage**|Non-Unicode code page of the collation. Please see [Appendix G DBCS/Unicode Mapping Tables](https://msdn.microsoft.com/en-us/library/cc194886.aspx) and [Appendix H Code Pages](https://msdn.microsoft.com/en-us/library/cc195051.aspx) to translate these values and see their character mappings.|  
+|**LCID**|Windows LCID of the collation. Please see [LCID Structure](https://msdn.microsoft.com/en-us/library/cc233968.aspx) to translate these values (you will need to convert to `VARBINARY` first).|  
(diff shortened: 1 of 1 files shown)


craigg-msft:
Thank you @srutzky for the edits.  @edmacauley please review.

edmacauley:
Thanks @srutzky for expanding these definitions.  Looks great to me.  #sign-off

srutzky:
@edmacauley and @craigg-msft I have a question about links to other documentation. Is there a rule about external references? Such as to Wikipedia or blog posts? (And while not a concern for the moment, if blog posts are ok, is it acceptable to reference my own blog, assuming it is the best reference for the info, or is that considered poor form due to it being self-serving / conflict of interest type of stuff?). But back to the issue at hand: the update I made here linked to MS documentation for Code Page mappings. I just found out while doing some research and trying to use those mappings as a resource, that at least one of them, [Code Page 1257 Windows Baltic Rim](https://msdn.microsoft.com/en-us/library/cc195059.aspx) is wrong in a few places. That documentation is not editable so I cannot fix it. Can I reference [Windows-1257](https://en.wikipedia.org/wiki/Windows-1257) on Wikipedia? In which case maybe it is best to be consistent and use Wikipedia for all of those mappings? On the plus side, there are two Code Pages missing from the MSDN documentation (1258 and one other) and at least [1258 is on Wikipedia](https://en.wikipedia.org/wiki/Windows-1258), hence it is more complete. Thoughts?

Also, on a somewhat related note, I did notice that while I was doing my best to conform to the existing style convention, I had used the "code" formatting and upper-case for the word `VARBINARY` when it seems by looking higher up on the page that bold and lower-case is being used for **varbinary**. If there was nothing to change for the links, should I still submit an edit to make that style consistent?

craigg-msft:
@srutzky Thank you for the question and interest in helping with content.  Linking to other relevant sources is ok as long as there a clear note about the purpose of the link, such as "For a current list of code pages, see xyx".  With that said the links should be for occasional references and support of the core article.  It would not be a good user experience if an article how a lot of external references and the "reading flow" essentially has some clicking back and forth.  As a general rule, blogs are ok if they are current, relevant, and the URL appears like it will be consistent. The problem with some blogs is the link is to the current topic of the blog and next month of next post will make the same URL reference different content. It also depends on who much relevant information is in the blog vs summarizing directly into the topic. A lot of it comes down to making the users learning experience and flow as smooth and click free as possible. Referencing other sources can enhance the overall learning experience but we also want to keep a reasonable balance between that and our topics just being used as a launching point to send users all over the place. 

yes, if you have the time to make that style change to varbinary, It would be great.

I hope this helps and thanks again for your time.

srutzky:
@craigg-msft Thanks for that info and I understand those concerns and goals. When the time comes, I will ask about specific blog posts.

I found something else to update on that page, so I made the varbinary and even _BIN / _BIN2 styles more consistent while I was in there. I submitted it via https://github.com/MicrosoftDocs/sql-docs/pull/175

PR #162, opened 2017-10-16
Correction of path to availability group databases

johlju:
- Changes to remove-a-primary-database-from-an-availability-group-sql-server.md
  - Correction of path to availability group databases


craigg-msft:
Thank you  @johlju

PR #166, opened 2017-10-18
Comment on NEWSEQUENTIALID/UuidCreateSequential byte shuffling

altenstedt:
This pull request will add a comment on NEWSEQUENTIALID that the
result of the function Uuidcreatesequential is not used as-is, some
byte shuffling is performed.  A link to a relevant blog post that
describes the byte shuffling algorithm is also added.

https://blogs.msdn.microsoft.com/dbrowne/2012/07/03/how-to-generate-sequential-guids-for-sql-server-in-net/


craigg-msft:
Thank you @altenstedt for the help.

PR #167, opened 2017-10-19
Updated CTP 1.1 notes regarding filestream

vin-yu:


craigg-msft:
Thank you @vin-yu  / @rothja can you please review.

rothja:
Approved. Thanks, Vin. We might be removing all these CTP improvement notes from the past. We're in discussion with Jeff and Lee and Travis about that. But we should publish this in case that doesn't happen. Also, if you use sql-docs-pr to make your changes, you can just #sign-off directly without going through the public review process. Thanks! #sign-off

PR #168, opened 2017-10-19
Documenting WITH RESULT SETS options

bgribaudo:
Text describing `WITH RESUT SETS` is adapted from the description of that set of options on https://docs.microsoft.com/en-us/sql/t-sql/language-elements/execute-transact-sql.

Change includes removing erroneous statement that `WITH RESULTS SETS` is mandatory if data is returned by an R script. The fact that this is incorrect is demonstrated by the fact that the following query/R script returns results without errors and without a `WITH RESULT SETS` clause:

    EXEC sp_execute_external_script
    @language = N'R',
    @script = N'OutputDataSet = InputDataSet',
    @input_data_1 = N'SELECT 1';

Also, change removes a couple of other remarks about returning results from R scripts that are covered by the new 'WITH RESULT SETS' description.

### DIFF
--- a/docs/relational-databases/system-stored-procedures/sp-execute-external-script-transact-sql.md
+++ b/docs/relational-databases/system-stored-procedures/sp-execute-external-script-transact-sql.md
@@ -67,6 +67,7 @@ sp_execute_external_script
         [ db_name . [ schema_name ] . | schema_name . ]   
         {table_name | view_name | table_valued_function_name }  
     | AS TYPE [ schema_name.]table_type_name  
+    | AS FOR XML  
 }  
 ```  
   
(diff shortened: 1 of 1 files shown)


craigg-msft:
Thank you @bgribaudo for the help.  @edmacauley please review.

edmacauley:
Thanks for your help, looks good to me. #sign-off

bgribaudo:
@edmacauley, I just thought of this:

I added the descriptive items for `WITH RESULT SETS` because syntax for that set of options was included in the original syntax code block. However, isn't `WITH RESULT SETS` actually a part of [`EXECUTE`](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/execute-transact-sql), not something specific to `sp_execute_external_script`?

If so, instead of moving forward with this pull request, would it be more appropriate to discard it and instead remove `WITH RESULT SETS` from this procedure's syntax block? It seems this would match how most stored procedure documentation is written—they usually do not mention `WITH RESULT SETS` in their syntax blocks.

craigg-msft:
@edmacauley please review

edmacauley:
It is actually part of execute, that's true.  It would be great if you don't remind removing it.  EXECUTE does have WITH RESULT SETS pretty well defined.

edmacauley:
@craigg-msft based on discussion it would be appropriate to close the PR at this time.

craigg-msft:
Thank you @bgribaudo  and @edmacauley closing.

bgribaudo:
Thanks @edmacauley & @craigg-msft! I reworked this page to remove `WITH RESULTS SETS` in #169.

PR #176, opened 2017-10-26
Fixing TOC and removing extraneous info from html device info doc

jt000:
- [x] The RS TOC contained multiple items on the same line. 
- [X] Frames are not used in the HTML renderer and are not affected by the AccessibleTablix device info
- [X] SharePoint 2007 doesn't need to be referenced (out of support)


guyinacube:
#sign-off

craigg-msft:
Thanks @jt000

PR #178, opened 2017-10-26
Fixing parameter name typo

bgribaudo:
Comment text on commit is incorrect--this removes unnecessary slashes from a parameter name


craigg-msft:
Thank you @bgribaudo @jeannt please review

PR #179, opened 2017-10-26
Changed Permissions Section

way0utwest:
The original section says sysadmin AND serveradmin (emphasis mine). This is incorrect. Either serveradmin or sysadmin is needed.


craigg-msft:
Thank you @way0utwest

PR #182, opened 2017-10-26
Update output-clause-transact-sql.md to fix indentation

Twizzes:
After the *output_table* list on line 90, the following paragraphs through line 127 were indented at the list level because of the leading space before *column_list* on line 98.


craigg-msft:
Thank you @Twizzes

PR #183, opened 2017-10-27
 fix references to various SSIS versions

csrowell:
We were using a "Current Version" include when we should have been referencing specific versions of SSIS which caused a ton of confusion when the "Current Version" include was updated to "2017".

### DIFF
--- a/docs/includes/ssisversion12-md.md
+++ b/docs/includes/ssisversion12-md.md
@@ -1,0 +1,1 @@
+SQL Server 2014 Integration Services (SSIS)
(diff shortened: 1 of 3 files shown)


craigg-msft:
Thank you @csrowell for the help.  @douglaslMS please review.

douglaslMS:
I approve these changes to files I own. I suggest one further change:

To the statement that was corrected to read "If you created the project in [!INCLUDE[ssISversion12](../../includes/ssisversion12-md.md)]", please add "or later".

PR #184, opened 2017-10-29
Update scm-services-prevent-automatic-startup-of-an-instance.md

janos-laszlo:
Update step 4 for more accuracy.


craigg-msft:
Thank you @janos-laszlo for the help.

PR #188, opened 2017-10-31
little correction of bytes vs. characters

chrgraefe:
This PR corrects the word bytes vs storage in the context of nvarchar(max)


craigg-msft:
Thank you @chrgraefe

PR #194, opened 2017-11-01
Fix repeated sentence

amtwo:
Remove one occurrence of "Top is a logical and physical operator." from being repeated twice.


craigg-msft:
Thank you @amtwo for the help.

PR #197, opened 2017-11-02
Fix bad example for sp_changemergepublication

jeroen-mostert:
As is [documented](https://docs.microsoft.com/sql/relational-databases/system-stored-procedures/sp-changemergepublication-transact-sql), `sp_changemergepublication` only accepts `90RTM` and `100RTM` for `publication_compatibility_level`, and it will give an error on other values. It's neither necessary nor possible to increase the compat level beyond `100RTM`, even for later versions of SQL Server, and implying it is [confuses people](https://stackoverflow.com/q/47076507/4137916).

A possible alternative is to remove this example altogether, as, at present, almost nobody has any need to change the property.


craigg-msft:
Thank you @jeroen-mostert

PR #202, opened 2017-11-02
Unclear large string types can't be used

greglow:
The description of the data type does not explicitly call out nvarchar(max) or varchar(max) which don't work.


craigg-msft:
Thank you @greglow  for the help.

PR #204, opened 2017-11-04
Update alter-database-scoped-configuration-transact-sql.md

Matteo-T:
Removing the "public preview" remark about IDENTITY_CACHE, since we are no longer in public preview :)


craigg-msft:
Thank you @Matteo-T

PR #209, opened 2017-11-08
Update replication-management-objects-concepts.md

SQLSourabh:


craigg-msft:
@SQLSourabh looking at the conflict, the text looks identical to what is there now?  should we close this pr?

SQLSourabh:
yes let's close this.. this might have been a duplicate.

PR #212, opened 2017-11-09
Correct false statements re: "Backslash" and fix title

srutzky:
* Removed false statements related to `\` being a client tool feature and _not_ a feature of T-SQL. The line continuation character is indeed a feature of T-SQL as I proved in the following post: [Backslash (Line Continuation)](https://SqlQuantumLeap.com/2017/10/27/line-continuation-in-t-sql/)
* Renamed from "Backslash" to "Backslash (Line Continuation)" because, well, that's what it is ;-)
* Added example for splitting a binary string

One thing that I did not change was the filename. It is currently **sql-server-utilities-statements-backslash.md**, but it doesn't really make sense to have "sql-server-utilities-statements" as part of the name since it isn't a utility / client statement. However, there are people who have linked to this URL so changing the name would break those links and that seems worse than having a poorly named file. Is there a protocol for making such changes that includes placing a file with the "old" filename here that does a simple redirect to the new location? If not, then I guess we are stuck with the misleading name.


craigg-msft:
Thank you @srutzky .  @MightyPen could you please review.

MightyPen:
Hi @srutzky , [A] We rarely change .md filenames unless there is an extra compelling reason (beyond a nice-to-do). [B] I am not convinced that the thrust of this PR 212 is completely correct.  I read your blog, but the surrounding context of its examples are not entirely clear to me.  When I try the backslash in the context of SQL Operations Studio ("SqlOps"; think sorta approx SSMS), the Transact-SQL interpreter is treating the '\' like a plain character; both on SELECT and on PRINT.
That is why I am not proceeding with this PR 212 at this time.
Thank you.  (MightyPen = GeneMi at MS)

MightyPen:
#hold-off

srutzky:
@MightyPen Hi there. For **(A)** Ok, that's what I had pretty much assumed. Thanks for confirming that. For **(B)**, what exactly did you try? Did you execute code against SQL Server or did you just see how it was perceived via SqlOps IntelliSense? I don't know what you mean by 

> the Transact-SQL interpreter is treating the '\\' like a plain character

The issue here is that the documentation currently states that using a backslash at the end of a line is a function of the client tool, _not_ of the SQL Server Engine. My example code shows that statement to be 100% false. You cannot use the "GO" batch separator in Dynamic SQL as _that_ is only known by client tools. But that is the case with the backslash.

MightyPen:
@srutzky In our new tool named SQL Operations Studio, I ran the following two-line Transact-SQL statement:

print N'albert \
einstein';

The displayed results were the following two lines, Not all the text on one line (and I will try to attach my screenshot image
line-continuation-test-pr-212-srutsky-sqlsos
):

albert \
einstein

FWIWorth, the presence or absence of the N made no difference.
There is seeming inconsistency, because the backslash did have the effect you predict when I changed from PRINT to SELECT.

When the tool context is changed to SSMS, the results are as you predict.
But overall, my test results confirm that the effect of the backslash is affected by which tool is used to submit the T-SQL code: but if the locus of the \ effect was in the T-SQL interpreter, no tool would return a two-line result (I presume).

You might want to test with a C# program that uses ADO.NET, to bypass our tools and their differences?

So, perhaps I might obtain more insight into whether the T-SQL interpreter treats \ as special.  But as of now, I am not comfortable with proceeding with this PR 212.  Nevertheless, I have learned a couple interesting things from working with you on this PR.  Thanks for that too.

srutzky:
@MightyPen Thanks for including that description of your testing and the screen shot. I cannot tell from the screen shot, but just to make sure, is there any space or anything else after the <kbd>\\</kbd> and before the <kbd>return</kbd> on the "albert " line? If so, that will cancel the line continuation.

Yes, I will test via `SqlClient` in C#, but just to mention, what you are seeing could easily be due to a bug in SqlOps that escapes the <kbd>\\</kbd> prior to sending it to SQL Server. And in fact, I just realized that you just found a bug in SqlOps: whether handled in the client tool or in SQL Server, there is no reason that you should not have seen "albert einstein" on a single line if that first line ends with a <kbd>\\</kbd>. And just to be sure about the difference in behavior when switching to `SELECT`, since the `SELECT` can't show a newline and would print on the same line in the cell anyway, how about testing out the binary string example, both using `PRINT` and `SELECT`. If the binary string works in only the `SELECT` then the bug is only with `PRINT`. And, if neither works, then the bug is with the <kbd>\\</kbd> handling in general, which is not without precident:

[SSMS ignores final \r\n / CRLF / Carriage Return + Line Feed in PRINT and RAISERROR](https://connect.microsoft.com/SQLServer/feedback/details/3141883/ssms-ignores-final-r-n-crlf-carriage-return-line-feed-in-print-and-raiserror)

Finally, it just occurred to me that your testing does not impact the legitimacy of the changes that I made. I am not meaning to come off as argumentative, but the current documentation states that line continuation is _not_ a feature of T-SQL / SQL Server. My examples in my blog post prove that entirely incorrect. You should try running the stored procedure example code since it bypasses the client. I suspect that you will see the correct output if you run the stored procedure examples.

Upon proving that line continuation via <kbd>\\</kbd> does indeed work within SQL Server when no client tool is involved (the Dynamic SQL in the stored procedure executes prior to anything being sent back to a client tool to be displayed), then any effect that any client tool has is a completely separate matter, and is actually a bug with that client tool (as you just discovered).

MightyPen:
@srutsky You have convinced me. Thanks for the extensive info you provided, and for your persistence.  I will recommend Approval of this PR 212.

MightyPen:
@craigg-msft Craig, I believe that we Should Approve this PR 212.  (SRutsky responded well to my earlier concerns.)  Thanks.

srutzky:
@MightyPen Thanks, and you are welcome 😄 . Do you have a moment to test the stored procedure example from my post? It shouldn't take long, just copy and paste. I would test myself, but SqlOps isn't going to be available for another few days, I believe.

If this is a bug within SqlOps, then who should file it, you or me? I just looked at MS Connect and don't see a way to select it as a version, though I suppose it would go into the "Tools (SSMS, etc)" category.

srutzky:
@MightyPen and @craigg-msft  To follow up on the suggestion to test with C# / `SqlClient`, I think even if that shows the behavior I expect, one _could_ argue that it is still a client program and hence handles the backslash. So, I came up with a more independent approach and tested via SQL Server Profiler. By looking at the exact T-SQL submitted to SQL Server by SSMS (or SQLCMD, etc), we can determine, without a doubt, where the line continuation is being handled. If handled by the client, then it won't be seen in Profiler, and if seen in Profiler, then definitely not handled by the client. And yes, it did come through into SQL Server so definitely not handled by the client (unlike `GO` which is never seen in Profiler, unless it is followed by an error). I have updated my blog post with this additional test.

craigg-msft:
Thanks you @srutzky and @MightyPen merging

srutzky:
@MightyPen (and @craigg-msft ): Just to confirm this behavior, now that SqlOpS is available for download, I just installed it and tested out the backslash for both character and binary strings. It works as expected for both. It must be that there was, as I had suggested at the time, a space (or some other, non-printable character) to the right of the backslash in MightyPen's test shown in the screen shot above. Putting a space to the right of the backslash is the only way I am able to reproduce the behavior shown in the screenshot above.

So, no bug in SqlOpS 😄 .

PR #215, opened 2017-11-14
Fix typo

mahilleb-msft:


craigg-msft:
Thank you @mahilleb-msft for the help.

PR #219, opened 2017-11-15
Update to STWithin

egranet:
Update to STWithin. The command is case sensitive. Since this is unusual for T-SQL, the documentation should reflect that.


PRmerger:
@egranet : Thanks for your contribution to the SQL documentation! The author, @"BYHAM", has been notified to review your proposed change.

PR #222, opened 2017-11-16
The unit information was missing

greijer:
timeout parameters did not specify unit of measurement (seconds)


craigg-msft:
Thank you @greijer for the help.

PR #227, opened 2017-11-21
Fixing reference to incorrect page; clarifying text

bgribaudo:
Two changes re `generated_always_type`/`generated_always_type_desc`:
- Page incorrectly associates these columns with Always On. Changed this so that the association is with temporal tables.
- Description indicates that these columns describe the type of column. This is ambiguous and conflicts with the meaning of columns `system_type_id` and `user_type_id`. Clarified.


PRMerger3:
@bgribaudo : Thanks for your contribution to the SQL documentation! The author, @edmacauley, has been notified to review your proposed change.

edmacauley:
Good catch and thank you for the update.

#sign-off

PR #279, opened 2017-12-14
Update upgrading-always-on-availability-group-replica-instances.md

allanhirt:
Added a note that mixing versions is not supported in the same AG.

> [!NOTE]  
>  Adding an replica with a later version of SQL Server to an AG is not supported outside of the rolling upgrade process. The only supported way to mix versions and do an upgrade with AGs is to use a distributed AG which is available in SQL Server 2016 or later in Enterprise Edition.


PRMerger3:
@allanhirt : Thanks for your contribution to the SQL documentation! The author, @MikeRayMSFT, has been notified to review your proposed change.

MikeRayMSFT:
#sign-off

MikeRayMSFT:
@allanhirt Thank you for your contribution!

PRMerger3:
Only the author of the article can use the sign off comment to add the ready-to-merge label.

craigg-msft:
Thank you @allanhirt and @MikeRayMSFT

PR #292, opened 2017-12-16
Update unicode-compression-implementation.md

davidbaxterbrowne:
This topic needs to explain the use of Unicode compression in Columnstore objects.  It appears that Unicode compression (or something similar) is in effect for ColumStores.  eg

use tempdb
go
drop table if exists uctest
create table uctest(
id int ,
a varchar(100) default left(replicate(cast(newid() as varchar(100)),10),100),
b nvarchar(100) default left(replicate(cast(newid() as varchar(100)),10),100),
c Nvarchar(max) default left(replicate(cast(newid() as varchar(100)),10),100),
index cci_uctest clustered columnstore

)

go

with q as (
select top 1000000 row_number() over (order by (select null)) rn 
from sys.objects o, sys.columns c, sys.columns c2
)
insert into uctest(id)
select rn from q
option (maxdop 1)

alter table uctest rebuild

select column_id,  sum(row_count) rows, sum(on_disk_size) dixe,   
       sum(row_count)/(sum(on_disk_size)/1024) rows_per_kb
from sys.column_store_segments cs
join sys.partitions p 
  on cs.partition_id = p.partition_id
where p.object_id = object_id('uctest')
group by column_id

### DIFF
--- a/docs/relational-databases/data-compression/unicode-compression-implementation.md
+++ b/docs/relational-databases/data-compression/unicode-compression-implementation.md
@@ -25,13 +25,13 @@ ms.workload: "Inactive"
 # Unicode Compression Implementation
 [!INCLUDE[appliesto-ss-asdb-xxxx-xxx-md](../../includes/appliesto-ss-asdb-xxxx-xxx-md.md)]
 
-  [!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)] uses an implementation of the Standard Compression Scheme for Unicode (SCSU) algorithm to compress Unicode values that are stored in row or page compressed objects. For these compressed objects, Unicode compression is automatic for **nchar(n)** and **nvarchar(n)** columns. The [!INCLUDE[ssDE](../../includes/ssde-md.md)] stores Unicode data as 2 bytes, regardless of locale. This is known as UCS-2 encoding. For some locales, the implementation of SCSU compression in SQL Server can save up to 50 percent in storage space.  
+  [!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)] uses an implementation of the Standard Compression Scheme for Unicode (SCSU) algorithm to compress Unicode values that are stored in row or page or ColumnStore compressed objects. For these compressed objects, Unicode compression is automatic for **nchar(n)** and **nvarchar(n)** columns. The [!INCLUDE[ssDE](../../includes/ssde-md.md)] stores Unicode data as 2 bytes, regardless of locale. This is known as UCS-2 encoding. For some locales, the implementation of SCSU compression in SQL Server can save up to 50 percent in storage space.  
   
 ## Supported Data Types  
  Unicode compression supports the fixed-length **nchar(n)** and **nvarchar(n)** data types. Data values that are stored off row or in **nvarchar(max)** columns are not compressed.  
   
 > [!NOTE]  
->  Unicode compression is not supported for **nvarchar(max)** data even if it is stored in row. However, this data type can still benefit from page compression.  
+>  Unicode compression is not supported for **nvarchar(max)** data even if it is stored in row. However, this data type can still benefit from page compression.  Unicode compression is supported for **nvarchar(max)** columns in ColumnStore column segments.
   
 ## Upgrading from Earlier Versions of SQL Server  
  When a [!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)] database is upgraded to [!INCLUDE[ssCurrent](../../includes/sscurrent-md.md)], Unicode compression–related changes are not made to any database object, compressed or uncompressed. After the database is upgraded, objects are affected as follows:  


PRMerger3:
@davidbaxterbrowne : Thanks for your contribution to the SQL documentation! The author, @BYHAM, has been notified to review your proposed change.

craigg-msft:
@MightyPen can you please review.

MightyPen:
@jovanpop-msft @jodebrui Jos has asked Jovan to assess this Columnstore documentation change.
Jovan, please keep me (MightyPen = GeneMi) in the loop.  I can process the outcome of your assessment.

MightyPen:
Jovan P. returns from vacation on Jan/08 2018.

craigg-msft:
@MightyPen any update?

MightyPen:
@jovanpop-msft @jodebrui Jos has asked Jovan to assess this Columnstore documentation change.
Jovan, please keep me (MightyPen = GeneMi) in the loop. I can process the outcome of your assessment.

srutzky:
@davidbaxterbrowne @craigg-msft @MightyPen @jovanpop-msft @jodebrui : I do not believe that these changes are correct. While it does appear that some form of compression is being done on `NVARCHAR(1 - 4000)` and `NVARCHAR(MAX)` (in-row data; haven't tested off-row), I do not think it is a form of [SCSU](http://www.unicode.org/reports/tr6/) which is the specific implementation that is the subject of this page. In order to determine if SCSU is being used, a variety of languages need to be tested. I tried a variety of scenarios using English, Hebrew, and Tibetan characters, and with both similar and random values. I could not find any indicator to conclusively determine that Unicode Compression is being used in Clustered Columnstore indexes. However, I did find that there are other inaccuracies / omissions on this page (in its current form; not related to this PR) and one or two related pages and was planning on updating them soon.

Testing indicates that the **encoding_type** is `3`, which means:

> 3 = STRING_HASH_BASED - string/binary column with common values in dictionary

according to the documentation for [sys.column_store_segments](https://docs.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-column-store-segments-transact-sql). That pages also states:

> All encodings take advantage of bit-packing and run-length encoding when possible.

It might be fair to note that some form of compression is performed on Clustered Columnstore indexes, but unless someone from the Columnstore team can confirm that it indeed uses SCSU, it should probably just be added to the **Note**, but not specifying that it is "Unicode Compression". Of course, then it would need to be determined if the compression is for in-row data only or also for off-row data.

craigg-msft:
Thank you @srutzky.  Are you saying we should close this PR 
  " I do not believe that these changes are correct. "

and you will open another for the other issues you spotted? 
  " However, I did find that there are other inaccuracies / omissions on this page (in its current form; not related to this PR) and one or two related pages and was planning on updating them soon."

srutzky:
@craigg-msft Hi Craigg. I am hesitant to say for certain that this should be closed because I do not want to be unfair to David since I do not have conclusive proof that it is _not_ Unicode Compression. My concern is that the edits suggested here will have the documentation state that it definitely _is_ Unicode Compression being used in Clustered Columnstore indexes, and I do not see conclusive proof of that either. Unicode Compression is a specific thing, not merely the fact that a Unicode string is being compressed. The note on this page already indicates that in-row values for `NVARCHAR(MAX)` columns _can_ benefit from PAGE COMPRESSION (which is true) which is a dictionary compression, _not_ Unicode Compression. I'm pretty sure that Clustered Columnstore is just the dictionary and bit-packing that the documentation states, and does not involve Unicode Compression.

All of that to say, at the moment, I think if this page is going to be updated to reflect Clustered Columnstore indexes, it should only be the "note" that is updated to include that in-row (haven't tested off-row) data for `NVARCHAR(MAX)` columns can benefit from the compression inherent in Clustered Columnstore indexes. Again, if it is possible to get confirmation from the team in charge of Clustered Columnstore indexes, that would be best.

Regarding the other issues I mentioned, yes, I will do a PR for those once I finalize the testing, but that will likely be a week or two out.

MightyPen:
@craigg-msft Vassilis Papadimos responded to this PR 292 via email. Here is Vassilis's response:

I think the right external answer is that Unicode compression is not used in columnstore indexes, and it doesn’t matter because other aspects of columnstore compression provide much of the benefits of Unicode compression anyway.  So “srutzky” is right.

Strings in columnstore dictionaries are compressed with Huffman coding, so we’re not wasting two bytes for each Unicode character. I don’t think we’ve documented that we use Huffman coding anywhere (it is mentioned obliquely in some third party blog posts such as http://www.nikoport.com/2014/09/09/clustered-columnstore-indexes-part-40-compression-algorithms). Huffman coding gets you much of the benefits of Unicode compression, while it is much more widely applicable. It’s a separate thing from dictionary compression and bit packing (Huffman coding applies to the entries in the dictionary, dictionary compression and bit packing apply to the columnstore segments referencing a dictionary). Bu I really don’t think we should get into this level of detail.

Vassilis.

MightyPen:
@craigg-msft  This PR 292 should be Closed.
Vassilis and SRutzky are not confident about this change.  Jovan P. has not responded.

craigg-msft:
Thank you @srutzky for all of your time.

PR #295, opened 2017-12-19
Update note on state

accjon:
Note on State says a non-null in collation_name means that the database is ready to accept a connection. Just checked tested this on a database being restored. Collation_name was populated before the database was available for connection and while state was still restoring.


PRMerger3:
@accjon : Thanks for your contribution to the SQL documentation! The author, @edmacauley, has been notified to review your proposed change.

edmacauley:
Thanks for your contribution  #sign-off

PRMerger3:
Only the author of the article can use the sign off comment to add the ready-to-merge label.

PR #299, opened 2017-12-22
Update docs to use lsb_release rather than hard coded distro

EvanCarroll:
Now we're not 16.04 specific.

### DIFF
--- a/docs/linux/quickstart-install-connect-ubuntu.md
+++ b/docs/linux/quickstart-install-connect-ubuntu.md
@@ -20,14 +20,14 @@ ms.workload: "Active"
 
 [!INCLUDE[tsql-appliesto-sslinux-only](../includes/tsql-appliesto-sslinux-only.md)]
 
-In this quick start tutorial, you first install SQL Server 2017 on Ubuntu 16.04. Then connect with **sqlcmd** to create your first database and run queries.
+In this quick start tutorial, you first install SQL Server 2017 on Ubuntu. Then connect with **sqlcmd** to create your first database and run queries.
 
 > [!TIP]
 > This tutorial requires user input and an internet connection. If you are interested in the [unattended](sql-server-linux-setup.md#unattended) or [offline](sql-server-linux-setup.md#offline) installation procedures, see [Installation guidance for SQL Server on Linux](sql-server-linux-setup.md).
 
 ## Prerequisites
 
-You must have a Ubuntu 16.04 machine with **at least 2 GB** of memory.
+You must have a Ubuntu machine with **at least 2 GB** of memory.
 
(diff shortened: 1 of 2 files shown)


PRMerger3:
@EvanCarroll : Thanks for your contribution to the SQL documentation! The author, @edmacauley, has been notified to review your proposed change.

EvanCarroll:
Actually do not commit this. It's not currently working because Microsoft doesn't package SQL Server for 17.04. Lol.

> When there is a convention.. **BREAK IT.** - Microsoft.

EvanCarroll:
There we go, now it should be fixed

edmacauley:
Proposed changes to the automated install script don't work.  Two examples: the script must still be run with sudo and the sql server conf line fails as mssql is not in the sudoers list.  This is on a stock Ubuntu 16.04.  Current live script does work, in the author's opinion, changes to sample-unattended-install-ubuntu.md should not be merged.

Pinging @rothja as he's the author of the other article in this PR.

edmacauley:
I do see what you mean by running sqlcmd as root, let me see if I can get a working version of the script, however.

rothja:
@EvanCarroll Thanks for the detailed suggestions you've provided in these files. Ed and I are reviewing your changes, and we'll probably have a few tweaks to your recommendations before publishing. I have a few questions for you. 

First, I see you're adding a new repository that takes the output from lsb_release. So you have that there to anticipate Microsoft creating new binaries specifically for the verison you currently running that is greater than 16.04?

Second, is your only concern about running the script as root that we do some things as root that could be avoided, such as launching sqlcmd? Otherwise, most of the actions require sudo during the installation process anyway, right?

I think you're removing some of the apt-get update calls. I believe those are necessary to refresh the package lists before attempting to install. If that is not the case, let me know.

We might have other questions as we go through your changes. Thanks, again for helping to make the content better.

EvanCarroll:
> mssql is not in the sudoers list. 

We're not running sudo as mssql, you run the script as any user. It runs the system utilities required as root, and the server and the client *as* `mssql`. @edmacauley

edmacauley:
Hi Thanks again for these changes.  Right now we're going to need to not make most of your proposed changes.  I have incorporated your changes w/ sudo and they should be published live soon.  We cannot change the version language or repos at this time, we only officially support Ubuntu 16.04 right now.  The product team will need to review before we can make those changes and I don't have a timeframe for that right now.
#please-close

rothja:
@EvanCarroll One of the problems we have at the moment is that many people are out for the holidays and we'll need to get some consensus on some of your suggestions. I'm out for the next week and a half as well. But I'll put this on my calendar to keep investigating your suggestions and talking to others about your repository recommendations. Thanks, again.

EvanCarroll:
A couple of things, `mssql-tools` is i only in 16.04, and the method used in the dotnet distros of writing to files in sources.d rather than using apt-add-repository is better imho.

PR #302, opened 2017-12-26
JSON data only applies to SQL Server 2016 and up

wsmelton:
As is this page shows that JSON data is available in any version of SQL Server, when it only applies to SQL Server 2016 and above.

Updated the header to use the `appliesto-ss2016-asdb-xxxx-xxx-md.md` include which shows the appropriate value for SQL Server.


PRMerger3:
@wsmelton : Thanks for your contribution to the SQL documentation! The author, @douglaslMS, has been notified to review your proposed change.

douglaslMS:
@wsmelton Thank you for noticing this.

douglaslMS:
#sign-off

PR #314, opened 2018-01-03
Update to include returning results with Result Sets

davidbaxterbrowne:
This topic omits the most common and important way to return data from a stored procedure, and the only way to return tabular data.

### DIFF
--- a/docs/relational-databases/stored-procedures/return-data-from-a-stored-procedure.md
+++ b/docs/relational-databases/stored-procedures/return-data-from-a-stored-procedure.md
@@ -26,7 +26,34 @@ ms.workload: "Active"
 [!INCLUDE[appliesto-ss-asdb-asdw-pdw-md](../../includes/appliesto-ss-asdb-asdw-pdw-md.md)]
  > For content related to previous versions of SQL Server, see [Return Data from a Stored Procedure](https://msdn.microsoft.com/en-US/library/ms188655(SQL.120).aspx).
 
-  There are two ways of returning result sets or data from a procedure to a calling program: output parameters and return codes. This topic provides information on both approaches.  
+  There are three ways of returning data from a procedure to a calling program: result sets, output parameters, and return codes. This topic provides information on the three approaches.  
+  
+  ## Returning Data Using Result Sets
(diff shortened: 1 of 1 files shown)


PRMerger3:
@davidbaxterbrowne : Thanks for your contribution to the SQL documentation! The author, @BYHAM, has been notified to review your proposed change.

davidbaxterbrowne:
We may want to additionally document the mechanism for result sets to be sent to the client.  I simply said that small results are spooled and sent to the client, while large results must be consumed by the client for execution to proceed.  This is observed behavior, but not the underlying mechanism.  I suspect that each session has a network buffer of some size, and results that fit in the network buffer won't stop execution, but I don't actually know the internals of how this works.

craigg-msft:
@edmacauley can you please review.

craigg-msft:
@edmacauley can you please review

docs/relational-databases/stored-procedures/return-data-from-a-stored-procedure.md
```diff
@@ -26,7 +26,34 @@ ms.workload: "Active"
...
 
-  There are two ways of returning result sets or data from a procedure to a calling program: output parameters and return codes. This topic provides information on both approaches.  
+  There are three ways of returning data from a procedure to a calling program: result sets, output parameters, and return codes. This topic provides information on the three approaches.  
+  
+  ## Returning Data Using Result Sets
+ If you include a SELECT statement in the body of a stored procedure (but not a SELECT ... INTO or INSERT ... SELECT), the rows specified by the SELECT statement will be sent directly to the client.  For large result sets the stored procedure execution will not continue to the next statement until the result set has been completely sent to the client.  For small result sets the results will be spooled for return to the client and execution will continue.  If multiple such SELECT statements are run during the exeuction of the stored proceudre, multiple result sets will be sent to the client.  This behavior also applies to nested TSQL batches, nested stored procedures and top-level TSQL batches.
```
srutzky: 1. By convention ( [sp_executesql](https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-executesql-transact-sql) and [EXECUTE](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/execute-transact-sql) ), "TSQL" should be fully written out as "Transact-SQL".

1. Might be helpful to hint at the meaning of "nested Transact-SQL batches", something like:

    > nested Transact-SQL batches (i.e. dynamic SQL, EXECUTE, sp_executesql, etc.)

1. While it is a bad practice and there is even a server configuration option to prevent it, would it make sense to add "triggers" to the "This behavior also applies to" list? I only ask since it is still possible to do and people do it since it isn't prevented by default. Mentioning it could include a warning / caution that it is deprecated behavior that will be removed in a future version and can presently be disabled via sp_configure.

edmacauley:
Looks good, sorry for the delay #sign-off

PRMerger13:
Only the author, @rickbyh, can sign off on this pull request.

edmacauley:
@craigg-msft can you please merge? thanks.

craigg-msft:
Thank you @davidbaxterbrowne and @edmacauley

PR #317, opened 2018-01-04
Clarified example in SET ROWCOUNT

greglow:
This example was confusing to readers as it showed two different queries, one with a count, and one without. It would be far better for it to be the same query in each case, with SET ROWCOUNT 4 as the only difference.

### DIFF
--- a/docs/t-sql/statements/set-rowcount-transact-sql.md
+++ b/docs/t-sql/statements/set-rowcount-transact-sql.md
@@ -69,12 +69,13 @@ SET ROWCOUNT { number | @number_var }
  Requires membership in the public role.  
   
 ## Examples  
- SET ROWCOUNT stops processing after the specified number of rows. In the following example, note that over 500 rows meet the criteria of `Quantity` less than `300`. However, after applying SET ROWCOUNT, you can see that not all rows were returned.  
+ SET ROWCOUNT stops processing after the specified number of rows. In the following example, note that this query returns 537 rows that meet the criteria of `Quantity` less than `300`. However, after applying SET ROWCOUNT, you can see that not all rows were returned.  
   
 ```  
 USE AdventureWorks2012;  
 GO  
-SELECT count(*) AS Count  
+
+SELECT *
 FROM Production.ProductInventory  
 WHERE Quantity < 300;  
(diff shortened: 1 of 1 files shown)


PRMerger3:
@greglow : Thanks for your contribution to the SQL documentation! The author, @edmacauley, has been notified to review your proposed change.

edmacauley:
Thanks for your proposed change @greglow.  I don't believe it's needed at this time.

edmacauley:
Do you think that it would be clear to change both examples to use count(*)? We just don't want an example that returns so many results.

greglow:
Actually no. Because then the ROWCOUNT doesn't apply to the rows counted,
only to the total row. I think having counts confuses these examples. The
reader won't get that the ROWCOUNT applies to the rows returned, not to the
number of rows counted.



I think you need two simple example, both with the same query, and the only
difference is the use of ROWCOUNT. Otherwise, the examples don't actually
show what's intended.



Perhaps find a query that returns 4 rows, then use the same query with SET
ROWCOUNT 1 (or 2).



Regards,



Greg



Dr Greg Low



1300SQLSQL (1300 775 775) office | +61 419201410 mobile│ +61 3 8676 4913 fax

SQL Down Under | Web: www.sqldownunder.com |http://greglow.me



*From:* Ed Macauley [mailto:notifications@github.com]
*Sent:* Thursday, 18 January 2018 9:39 AM
*To:* MicrosoftDocs/sql-docs <sql-docs@noreply.github.com>
*Cc:* Greg Low <greg@greglow.com>; Mention <mention@noreply.github.com>
*Subject:* Re: [MicrosoftDocs/sql-docs] Clarified example in SET ROWCOUNT
(#317)



Do you think that it would be clear to change both examples to use
count(*)? We just don't want an example that returns so many results.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/MicrosoftDocs/sql-docs/pull/317#issuecomment-358473534>,
or mute the thread

edmacauley:
@greglow, thank you again, I think we'll leave the topic as it is right now but keep an eye on it.  @craigg-msft can you #please-close thanks

PR #318, opened 2018-01-06
Fix unmatched parenthesis in code block

TobiasBengtsson:
An inline code block contains an unmatched parenthesis, and the parenthesis is instead located outside of the code block (probably due to a typo). This commit moves the parenthesis to inside the inline code block.


edmacauley:
Thank you for catching that typo!  #sign-off

PRMerger3:
@TobiasBengtsson : Thanks for your contribution to the SQL documentation! The author, @edmacauley, has been notified to review your proposed change.

PR #339, opened 2018-01-11
Add context to log location warning

johnmart82:
Adding context about when the corosync.conf file is created as the warning is placed ahead of the sample code to create the cluster, which in turn creates the conf file. The flow of the document means that anyone reading it might try to locate the conf file ahead of it existing. This addition of detail about when the file is created will prevent that.


PRMerger3:
@johnmart82 : Thanks for your contribution to the SQL documentation! The author, @MikeRayMSFT, has been notified to review your proposed change.

PR #346, opened 2018-01-16
Improve "See also" sections for string functions

srutzky:
1. Added relevant function links to **See also** sections. "Relevant" is based on a general operation, such as: converting the value between types, modifying the contents of the string, taking sections of the string, finding something within the string, generating strings, determine  a character value or create a character from a value, etc. The idea is to increase the chances that people can find something helpful that they might not have known the name of.

1. As much as possible made **See also** sections consistent with actual functions listed first, then generic function list pages, etc after.

1. Changed link title encoding where necessary to be consistent such that `(` and `)` are HTML-encoded as `&#40;` and `&#41;`, respectively (instead of as just parenthesis).

1. Added `STRING_AGG` to **Aggregate Functions** page (new as of SQL Server 2017). It is a valid aggregate function and behaves consistently with the definition on that page as the two tests below show: a) it works in a `HAVING` clause, and b) it ignores `NULL`s.

    ``` sql
    SELECT tmp.[ID], STRING_AGG(tmp.[name], '~')
    FROM  (VALUES (1, 'a'), (2, 'gg'), (1, 'b'), (2, 'hh'), (3, 'zzz')) tmp (ID, [name])
    GROUP BY tmp.[ID]
    HAVING STRING_AGG(tmp.[name], '~') = 'hh~gg';
    ```
    returns:

    ```
    2	hh~gg
    ```

   And:

    ``` sql
    SELECT tmp.[ID], STRING_AGG(tmp.[name], '~')
    FROM  (VALUES (1, 'a'), (2, 'zzz'), (1, 'b'), (2, NULL), (2, 'gg')) tmp (ID, [name])
    GROUP BY tmp.[ID];
    ```
    returns:

    ```
    1	b~a
    2	gg~zzz
    ```


PRMerger3:
@srutzky : Thanks for your contribution to the SQL documentation! The author, @edmacauley, has been notified to review your proposed change.

edmacauley:
Thanks for these changes, @srutzky #sign-off

PRMerger20:
Only the author, @edmaca, can sign off on this pull request.

edmacauley:
@craigg-msft can you merge?

craigg-msft:
Thank you @srutzky and @edmacauley

srutzky:
@edmacauley and @craigg-msft : you are welcome.

PR #349, opened 2018-01-17
Added line break after "Role" for consistency

jadarnel27:
"Role" was the only item in the "Terms and Definitions" list that didn't have a line break before its definition.


PRMerger3:
@jadarnel27 : Thanks for your contribution to the SQL documentation! The author, @MikeRayMSFT, has been notified to review your proposed change.

PR #354, opened 2018-01-19
A note about SQL Server password policy which cause docker to stop suddenly

mirsaeedi:
Many of the juniors follow this guideline and most of them set a simple password such as "123" for their tutorial to set up their docker container. However, they see after some seconds docker stops working. The reason behind is SQL Server password policy. But, since they run the command with -d they can't see the log and are not able figure out what is the reason for their headache.

### DIFF
--- a/docs/linux/quickstart-install-connect-docker.md
+++ b/docs/linux/quickstart-install-connect-docker.md
@@ -63,6 +63,9 @@ This image consists of SQL Server running on Linux based on Ubuntu 16.04. It can
    ```
 
    > [!NOTE]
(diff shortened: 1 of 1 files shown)


PRMerger10:
@mirsaeedi : Thanks for your contribution to the Azure documentation! The author, @rothja, has been notified to review your proposed change.

rothja:
@mirsaeedi Thank you for the suggestion. This is a good thing to highlight, since it causes a lot of painful troubleshooting for someone new to SQL containers. Your write-up is very thorough and helpful. Thanks, again. #sign-off

PR #356, opened 2018-01-19
Code correction so the example actually works.

sterlingpetersen:


PRMerger7:
@sterlingpetersen : Thanks for your contribution to the Azure documentation! The author, @jeannt, has been notified to review your proposed change.

jeannt:
@sterlingpetersen Thanks; good catch!

jeannt:
#sign-off

PR #364, opened 2018-01-23
Update memory-properties.md

eugeneniemand:
I spent hours troubleshooting performance issues and cube processing errors due to memory pressure. I looked at perfmon counters and they always stuck at 16GB regardless of how much free memory (50GB in my case) I had on the server. Whenever I googled "16gb limit ssas" and all sorts of variations it all point back to the above documentation. No where in this is there any mention of limits based on the edition of the sql server. Only after someone in the community pointed me to this [page](https://docs.microsoft.com/en-us/sql/sql-server/editions-and-components-of-sql-server-2016#Cross-BoxScaleLimits) did I realise where the problem lies. I propose the above change or somehow mention and link to the link provided.

### DIFF
--- a/docs/analysis-services/server-properties/memory-properties.md
+++ b/docs/analysis-services/server-properties/memory-properties.md
@@ -61,9 +61,9 @@ The following properties apply to both tabular and multidimensional modes unless
 
  Values between 1 and 100 represent percentages of **Total Physical Memory** or **Virtual Address Space**, whichever is less. Values over 100 represent memory limits in bytes.
  
- **Note:** Maximum memory utilized per instance of Analysis Services	
- * Enterprise Edition - Operating System Maximum	
- * Standard Edition - Tabular: 16 GB or MOLAP: 64 GB
+**Note:** Maximum memory utilized per instance of Analysis Services is dependant on the SQL Server Edition
+- [SQL Server 2016](https://docs.microsoft.com/en-us/sql/sql-server/editions-and-components-of-sql-server-2016#Cross-BoxScaleLimits)
+- [SQL Server 2017](https://docs.microsoft.com/en-us/sql/sql-server/editions-and-components-of-sql-server-2017#Cross-BoxScaleLimits)
   
  **LowMemoryLimit**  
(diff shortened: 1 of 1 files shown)


PRMerger8:
@eugeneniemand : Thanks for your contribution to the Azure documentation! The author, @Minewiskan, has been notified to review your proposed change.

johlju:
@eugeneniemand I suggestion. Could we reference back to the page https://docs.microsoft.com/en-us/sql/sql-server/editions-and-components-of-sql-server-2016#Cross-BoxScaleLimits instead of duplicating the information. If a change is made on the first page, this page might be forgotten.

eugeneniemand:
@johlju This was my initial instinct, however that is a version specific (SQL 2016) page and there is another for 2017. Unless we change it to something like:

**Note:** Maximum memory utilized per instance of Analysis Services
- [SQL Server 2016](https://docs.microsoft.com/en-us/sql/sql-server/editions-and-components-of-sql-server-2016#Cross-BoxScaleLimits)
- [SQL Server 2017](https://docs.microsoft.com/en-us/sql/sql-server/editions-and-components-of-sql-server-2017#Cross-BoxScaleLimits)

PRMerger12:
@eugeneniemand : Thanks for your contribution to the Azure documentation! The author, @Minewiskan, has been notified to review your proposed change.

eugeneniemand:
@johlju This was my initial instinct, however that is a version specific (SQL 2016) page and there is another for 2017. Unless we change it to something like the last commit

johlju:
@eugeneniemand the 2017 page links back to the older version of the page (at the top), but I understand your point since you link to the header and the reader could miss that the page is for the wrong version. So your change is much better, and looks good to me.

Minewiskan:
@eugeneniemand @johlju I made changes to this article yesterday as part of a different PR. The sentence and link at the top should help direct readers to the editions and supported features article. We generally like to minimize having something like memory limits shown in multiple places. We like to have a single source of truth when it comes to something that can change with versions or editions. When linking to another article that may have more than one version, we try to link to the latest version. We also try to put that link where it's most likely to be read.  I'll close close this PR without signing off, but that doesn't mean I didn't agree with your suggested change. I just rolled a slightly different solution into a different PR. Thank you for contributing!

Minewiskan:
#hold-off

Minewiskan:
#please-close

PR #366, opened 2018-01-24
Execute against non-dropped databases

rsanderson2350:
Executing msdb.managed_backup.sp_backup_config_basic against an @database_name that has been dropped does not update dropped database configuration, only non-dropped database configuration records are updated.


PRMerger17:
@rsanderson2350 : Thanks for your contribution to the Azure documentation! The author, @MikeRayMSFT, has been notified to review your proposed change.

PR #396, opened 2018-01-30
Included a stronger recommendation to leave this disabled

greglow:
The current article discusses how to change the option but gives no hint on why it is disabled by default, and why it should normally remain that way. I've added content to explain that, and to suggest one of the limited situations where the option might need to be enabled.


PRMerger20:
@greglow : Thanks for your contribution to the Azure documentation! The author, @MikeRayMSFT, has been notified to review your proposed change.

MikeRayMSFT:
#sign-off

PR #408, opened 2018-02-10
Remove sudo from docker commands in bash

charlvictor:
Preferably docker commands should not be run with sudo commands.  On MacOS sudo is definitely not required.


PRMerger20:
@charlvictor : Thanks for your contribution to the Azure documentation! The author, @rothja, has been notified to review your proposed change.

rothja:
@charlvictor Thanks for the feedback. In our tests on Linux, the sudo was required for docker commands to work properly. Are you working with Linux too, or do you think that this is a MacOS variation? Also, just out of curiosity, does it give you an error on MacOS when you use sudo? I haven't testsed recently on Mac, so I appreciate the feedback. Thanks!

charlvictor:
@rothja I have not tested on Linux, but I did some reading and it seems by default you will need to use sudo, but there are some post installation documentation on the docker site to guide you how set it up that you don't have to run it with sudo:
https://docs.docker.com/install/linux/linux-postinstall/

On MacOS, Running the docker commands with sudo will not give any error, but generally you don't want to be running commands with sudo or as root if you don't have too. I have a Docker for Mac installation and by default you don't have to sudo your docker commands.

Maybe I can update the PR to add a side note in the docs to point to this URL in case users don't want to run the commands with sudo?

rothja:
@charlvictor Thanks to the link to that docker article. I was not aware of that option myself. I think adding a note linking users to this location if they want to avoid sudo in their docker commands would be great. For simplicity in the article, I'd like to leave the sudo in the bash examples though. I hope that is not too inconvenient for MacOS users.

PRMerger17:
@charlvictor : Thanks for your contribution to the Azure documentation! The author, @rothja, has been notified to review your proposed change.

PRMerger16:
@charlvictor : Thanks for your contribution to the Azure documentation! The author, @rothja, has been notified to review your proposed change.

rothja:
@charlvictor Thanks again for your comment. I made the changes we discussed in your branch. I hope you do not mind me doing this. I think this reflects what we decided to do. I'm going to sign off on this now to get the changes merged in. Note that there are other changes in the topic to just reflect the latest state of the topic in the repo (so we don't regress any other changes made since you opened this PR). Thanks!

rothja:
#sign-off

charlvictor:
@rothja 👍🏻

PR #412, opened 2018-02-14
Fix formatting syntax from YYYY, DD to yyyy, dd

franciscocha:


PRMerger9:
@franciscocha : Thanks for your contribution to the Azure documentation! The author, @edmacauley, has been notified to review your proposed change.

franciscocha:
@edmacauley Do you even understand the issue I'm trying to solve here? SQL doesn't understand YYYY as an actual format it is yyyy. Every other database engine that uses this type of formatting syntax correctly documents its usage. Same with DD its dd. This type of documentation provides confusion when one is trying to format dates which is what I was trying to do. As I stated before please try to actually understand the problem then just being defensive about the actual solution.

franciscocha:
What's the point of asking others to contribute when a simple change can't even be approved?

franciscocha:
Examples of what useful documentation does its job by actually stating the format necessary to achieve the desired output.

https://dev.mysql.com/doc/refman/5.5/en/date-and-time-functions.html#function_date-format

https://docs.mongodb.com/manual/reference/method/Date/

franciscocha:
@edmacauley no response thanks for the lack of communication.

PR #419, opened 2018-02-17
Add Sequel to the list of ORMs

JesseChavez:
This gem is a full feature ORM and more, It will allow you to connect to SQL server via ADO, ODBC, JDBC, and TinyTds.

PR #430, opened 2018-02-23
Revert "Add Sequel to the list of ORMs"

craigg-msft:
Reverts MicrosoftDocs/sql-docs#419
my fault for not communicating better but we do want to limit the number of links.  there are lots of great resources but we cannot list them all for a variety of reasons including we are on the hook to be sure they stay fresh.  So a rough bar is those that are aligned with general/central organizations..


PRMerger20:
@craigg-msft : Thanks for your contribution to the Azure documentation! The author, @MightyPen, has been notified to review your proposed change.

craigg-msft:
@MightyPen no need to review.

PR #437, opened 2018-03-01
Fix code block syntax name for SQL blocks

prasadu-microsoft:
SQL Code blocks were missing the 'sql' language tag and were therefore being generated as generic code elements instead of highlighted SQL.


PRMerger7:
@prasadu-microsoft : Thanks for your contribution to the Azure documentation! The author, @edmacauley, has been notified to review your proposed change.

edmacauley:
@prasadu-microsoft thank you for that edit, great catch.  #sign-off

PR #442, opened 2018-03-02
technical inaccuracy in when uncommittable trans will be rolled back

NReilingh:
If a transaction becomes uncommittable in a nested batch (using sp_executesql), the transaction will NOT be rolled back when this batch finishes. The transaction will be rolled back when the batch that called sp_executesql finishes. So it is incorrect to say "When a batch finishes running". I am not sure if "outermost batch" is the accepted terminology for this, but it seems reasonable.

(Sorry if this diff is difficult to read — only changed three words after the double `<br />`. Should have used semantic linefeeds :-)


PRMerger7:
@NReilingh : Thanks for your contribution to the Azure documentation! The author, @edmacauley, has been notified to review your proposed change.

edmacauley:
Thank you @NReilingh for your edit and pointing out where it changed, was easy to read.  Looks good to me.  #sign-off

PR #445, opened 2018-03-04
fix the order of the sections

yoichinak:


edmacauley:
thank you @yoichinak, #sign-off

PR #489, opened 2018-03-29
APRC data not persisted on restart

tboggiano:
Forced plans are no longer forced that were auto forced and recommendations no longer exists in DMV.


PRMerger12:
@tboggiano : Thanks for your contribution! The author, @jovanpop-msft, has been notified to review your proposed change.

craigg-msft:
Thank you @tboggiano

PR #536, opened 2018-04-12
Clarify DBCC EMPTYFILE behavior

way0utwest:
When the operation is cancelled, the file used with EMPTYFILE is not marked as read-only and can still have new a data added. Clarified this.


PRMerger7:
@way0utwest : Thanks for your contribution! The author, @barbkess, has been notified to review your proposed change.

PR #538, opened 2018-04-13
Added Note on Transactional Replication and Distributed AGs

SeanGallardy:
Distributed AG Forwarders cannot be setup as a publisher in transactional replication in SQL16/17. This was confirmed by the product group but not in the documentation. Added a note expressly calling this out. I believe this is one of the better places for the information as the Distributed AG information is just a single page and doesn't' have a "non-supported" features area.


PRMerger13:
@SeanGallardy : Thanks for your contribution! The author, @allanhirt, has been notified to review your proposed change.

allanhirt:
I'm fine with this change.

PR #541, opened 2018-04-16
Update char-transact-sql.md

fbsolo:
Text revisions to tighten and optimize the reading flow of the material.


craigg-msft:
Thank you @fbsolo

PR #552, opened 2018-04-18
changed wording on autoindent setting

aldonetti:
as the previous text seemed to suggest the property to act on wrapping rather than indenting


PRMerger6:
@aldonetti : Thanks for your contribution! The author, @yualan, has been notified to review your proposed change.

craigg-msft:
Thank you @aldonetti for the help.  @yualan fyi.

PR #554, opened 2018-04-18
Removed misleading and incorrect sentence

way0utwest:
"remote access" is indeed an obscure feature, and I
don't if it still possible to define a "remote server" à la SQL 6.5, but
probably not. For a execution of a stored procedure over a linked server,
the setting "remote access" has no meaning, but you need to set the server
option "rpc out". Which also is required for EXEC AT, if memory serves.

In any case, the sentence should just be removed.


PRMerger19:
@way0utwest : Thanks for your contribution! The author, @MikeRayMSFT, has been notified to review your proposed change.

MikeRayMSFT:
@way0utwest - thanks for your submission!

MikeRayMSFT:
#sign-off

PR #621, opened 2018-05-06
Update quickstart-install-connect-docker.md

bmarshalsea:
The docker run scripts setup via -e "MSSQL_SA_PASSWORD", this does not appear to work, and once changed to SA_PASSWORD fixes the problem. If there's a reason for MSSQL_SA_PASSWORD here then this needs to be explained.
(Note: this works for the microsoft/mssql-server-windows-developer image.

### DIFF
--- a/docs/linux/quickstart-install-connect-docker.md
+++ b/docs/linux/quickstart-install-connect-docker.md
@@ -51,13 +51,13 @@ This image consists of SQL Server running on Linux based on Ubuntu 16.04. It can
 1. To run the container image with Docker, you can use the following command from a bash shell (Linux/macOS) or elevated PowerShell command prompt.
 
    ```bash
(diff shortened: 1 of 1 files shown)


PRMerger14:
@bmarshalsea : Thanks for your contribution! The author, @rothja, has been notified to review your proposed change.

wsmelton:
I can confirm this one, the docker build for the image based on the PR history was changed in the `mssql-docker` repository in this PR: Microsoft/mssql-docker#113

rothja:
@wsmelton Thanks for testing this and finding the problem. I'm not sure why the MSSQL_* version of the environment variable doesn't work, because all other instances changed the older SA_PASSWORD to the newer form of MSSQL_SA_PASSWORD. However, I'm going to sign off on this and then investigate, because I know for a fact that SA_PASSWORD is still recognized and should work (also you verified it on your side). Thanks for reporting this and making the change. 

#sign-off

wsmelton:
I do recall it used to have the MSSQL prefix and noted it on the docker hub page. But that had since been updated when you check it now.

PR #629, opened 2018-05-08
Removing inaccurate content

sahaj08:
CETAS only exports data to hadoop and wasb, does not import data. Looks like someone copy-pasted this generic polybase content into this doc. And this statement is not supported in SQL Server, so removing the link for getting started with polybase doc, since that's specific to sql server.


PRMerger10:
@sahaj08 : Thanks for your contribution! The author, @edmacauley, has been notified to review your proposed change.

edmacauley:
Thanks @sahaj08 for catching that.  -- You're probably right about how it was created.  #sign-off

PR #630, opened 2018-05-09
Seems the type is not right

vitorfariatomaz:
Seems the type is not right


PRMerger12:
@vitorfariatomaz : Thanks for your contribution! The author, @edmacauley, has been notified to review your proposed change.

edmacauley:
@vitorfariatomaz thank you for that catch.  I noticed that many of the examples have similar problems and have created a task to fix them all.  #sign-off

PR #631, opened 2018-05-10
Update misleading statements about TRUNCATE TABLE

sqlsauce:
The document says that TRUNCATE TABLE is non-logged, but that is incorrect and misleads the reader in a big way. It does not log the removal of individual rows, but the operation (removing pages) is captured in the transaction log. That is an important distinction because readers of this document may also be considering log shipping and will conclude that TRUNCATE TABLE is not compatible with that. Secondly, the document states that TRUNCATE TABLE does not fire triggers. While it does not fire DML triggers, which are used to detect changes for merge replication, it does fire DDL triggers.

### DIFF
--- a/docs/relational-databases/replication/administration/frequently-asked-questions-for-replication-administrators.md
+++ b/docs/relational-databases/replication/administration/frequently-asked-questions-for-replication-administrators.md
@@ -159,7 +159,7 @@ manager: craigg
 ## Database Maintenance  
   
 ### Why can't I run TRUNCATE TABLE on a published table?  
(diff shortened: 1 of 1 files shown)


PRMerger10:
@sqlsauce : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
@sqlsauce Thank you for your SQL Content contribution! You are correct, thanks for catching this :) #sign-off

MashaMSFT:
@craigg-msft per our discussion, I am merging this.

sqlsauce:
Great! That was my first contribution to the sql-docs repo  :-)

MashaMSFT:
@sqlsauce Thanks again for your contribution! As you probably already know,  your change is now live :) If you find anything else that needs to be changed, please let us know!

PR #636, opened 2018-05-14
remove wrong pointer dereference for 64-bit SQLSetXxx

cladisch:
SQLSetDescField/SQLSetStmtAttr/SQLSetStmtOption pass the value by value, not by reference, so the description must say "ValuePtr" instead of "*ValuePtr". For values that are pointers (SQL_xxx_PTR), the description could be misinterpreted as the pointed-to value having a 64-bit type.

### DIFF
--- a/docs/odbc/reference/odbc-64-bit-information.md
+++ b/docs/odbc/reference/odbc-64-bit-information.md
@@ -392,7 +392,7 @@ typedef SQLULEN BOOKMARK;
   
  **SQLSetDescField**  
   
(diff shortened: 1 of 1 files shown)


PRMerger9:
@cladisch : Thanks for your contribution! The author, @MightyPen, has been notified to review your proposed change.

MightyPen:
#sign-off 
After our review, we like this improvement.  Thanks!

MightyPen:
@cladisch @David-Engel cladisch, David Engel checked, and double-checked with the Dev team, and this PR 636 looks like a nice improvement.  Thank you for submitting it!

PR #651, opened 2018-05-21
Clarify default return value for property

way0utwest:
The new LastGoodCheckDBTime property returns a 1900-01-01 at midnight if checkdb hasn't been run on a database. This is different from a NULL value, which would be returned if the property were incorrect or the version didn't support this.


PRMerger9:
@way0utwest : Thanks for your contribution! The author, @edmacauley, has been notified to review your proposed change.

edmacauley:
Thank you @way0utwest for that edit.  #sign-off

PRMerger19:
@way0utwest : Thanks for your contribution! The author, @edmacauley, has been notified to review your proposed change.

PR #654, opened 2018-05-24
Incorrect result set column name

stevenb-mcgrath:
The result set section lists a column named "ordinal_position" when the actual column name is "column_ordinal" at least as of 14.0.3025.34


PRMerger14:
@stevenb-mcgrath : Thanks for your contribution! The author, @edmacauley, has been notified to review your proposed change.

edmacauley:
Thank you for catching that #sign-off

PR #657, opened 2018-05-25
Corrected and clarified ANSI_NULLS remarks

NReilingh:
Fixed a number of errors and omissions related to SET ANSI_NULLS and boolean data types in "Comparison Operators" and "Equals".

Resolves issues noticed on SO in 2012: https://stackoverflow.com/q/9707111/152891

### DIFF
--- a/docs/t-sql/language-elements/comparison-operators-transact-sql.md
+++ b/docs/t-sql/language-elements/comparison-operators-transact-sql.md
@@ -48,7 +48,7 @@ manager: craigg
   
  Unlike other [!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)] data types, a **Boolean** data type cannot be specified as the data type of a table column or variable, and cannot be returned in a result set.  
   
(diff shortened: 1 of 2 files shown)


ktoliver:
@douglaslMS - Would you review the proposed changes? Thank you.

douglaslMS:
@ktoliver see my comments above.

NReilingh:
@douglaslMS Thanks for the note about future tense — I've added a commit that corrects this in my additions.

I actually started by writing "comparing NULL to NULL" (commit 1630db9) but felt that it was ambiguous as to what kind of comparison was occurring. (For example, if you were doing a "not equals" comparison, the statement would be untrue.) If you are satisfied that any comparisons referred to on the "Equals" page are assumed to be equivalency comparisons, I have no objections to changing that wording.

douglaslMS:
@NReilingh Nick, thank you for the corrections. I could tell that you were trying to be extra clear. But yes, on the Equals page, I think we can assume we're talking about equivalency comparisons, and go with "comparing null to null."

douglaslMS:
#sign-off

PRMerger6:
@NReilingh : Thanks for your contribution! The author, @douglaslMS, has been notified to review your proposed change.

PR #658, opened 2018-05-25
add comparison table

NReilingh:
I feel like the overall T-SQL documentation on boolean expressions involving NULLs has been crying out for a very clear example of how NULLs work under the ANSI_NULLS setting. I've created a table that I think illustrates this well, and put it where I think it would be most appropriate, but it may be better suited to a different location.


PRMerger9:
@NReilingh : Thanks for your contribution! The author, @edmacauley, has been notified to review your proposed change.

edmacauley:
@NReilingh Thanks for your contribution, great idea.  #sign-off

PR #690, opened 2018-05-27
[New Content] Adding Windows configuration for Read-Scale AG

wsmelton:
Adding documentation around using SQL Server 2017 feature for read-scale Availability Group between Windows replicas. Only initial version was strictly for Windows to Linux, this covers Windows to Windows.

I found a few things not possible that I think even the Linux version is wrong. Primarily it makes mention us a SQL Server Express edition replica for configuration only replica. However you cannot enable HADR on Express Edition so I'm curious to know how that is even possible.

I have tested building out the Windows to Windows read-scale AG for the purpose of a customer that wanted to test. There are various options for the endpoint authentication much like the normal HA configuration with AGs so I tried to word that the best I could to make it clear.

I followed the same pattern I saw with the Linux version of read-scale using the include files, hope that was the proper method. Please review and let me know what needs to be updated or clarified.


MikeRayMSFT:
@wsmelton this fresh content is fantastic and very relevant. Thanks for your work bringing this together. Reviewing.

MikeRayMSFT:
Thank you @wsmelton - #sign-off

PR #692, opened 2018-05-27
Fix [!IMPORTANT] note inside code block

divega:
`> [!IMPORTANT]` does not work inside code blocks. This is an attempt to fix it by breaking the code block into three parts and converting syntax sub-sections from comments inside the code block into level-3 subtitles. 

Another option would be to not use `> [!IMPORTANT]`


PRMerger12:
@divega : Thanks for your contribution! The author, @edmacauley, has been notified to review your proposed change.

edmacauley:
Thanks @divega for the fix #sign-off

PR #693, opened 2018-05-28
Update statements.md

osfancy:
Adding UPDATE as a DML operation which modifies rows.


edmacauley:
looks fine to me #sign-off

PR #697, opened 2018-05-28
Rearrange acceptable data types for accessibility

dangowans:
There are quite a few places throughout the documentation where tables are used to display list content in two columns.  This is one of those places.  Using tables for layout purposes is a well known faux-pas.

I still used a table here, but tried to categorize the content a bit so that the table makes a bit more sense and isn't just noise to a screen reader.  Might not be perfect, but I think it's a start.

Should tables like this one be reviewed and converted to lists?  Should the tables be restructured to increase their accessibility?


PRMerger18:
@dangowans : Thanks for your contribution! The author, @edmacauley, has been notified to review your proposed change.

edmacauley:
#sign-off

PR #698, opened 2018-05-28
Underline "underline"

dangowans:
Currently styled as bold.  If the convention is for an underline, it should be underlined.


douglaslMS:
#sign-off

PR #699, opened 2018-05-28
Move "bit" under "numeric"

dangowans:
The definition of the "bit" data type is "An integer data type that can take a value of 1, 0, or NULL. "
When  a bit value is passed to the "isnumeric" function, 1 is returned.
The bit page was kind of floating in the table of contents, looking for a home.


barbkess:
@dangowans : thanks!  

#sign-off

PR #707, opened 2018-05-31
Update: Adding bit link

dangowans:
Since moving the "bit" page was approved, "in this section" should be updated as well.


GitHubber17:
@edmacauley - Would you review the proposed changes? Thanks

edmacauley:
#sign-off

PR #734, opened 2018-06-04
Explain trick behind using 3 separate EXEC calls

jzabroski:
This tripped me up and was very hard to figure out why I was getting this error when I tried porting this logic to C# for use with FluentMigrator - I'm trying to create a generic TestProfile that runs a FluentMigrator Profile to ensure there is no data in a test database. Other DB tools like Respawn have similar issues: https://github.com/jbogard/Respawn/issues/33


PRMerger18:
@jzabroski : Thanks for your contribution! The author, @CarlRabeler, has been notified to review your proposed change.

CarlRabeler:
#sign-off

PR #749, opened 2018-06-06
Update bulk-insert-transact-sql.md

raghavt-msft:


PRMerger12:
@raghavt-msft : Thanks for your contribution! The author, @edmacauley, has been notified to review your proposed change.

rothja:
@raghavt-msft Thanks for adding this example. Sorry for the long delay in signing off on this change. I'm doing that now and it should publish soon. Thank you!

raghavt-msft:
Hi Jason,

I don’t see the update on the docs yet - https://docs.microsoft.com/en-us/sql/t-sql/statements/bulk-insert-transact-sql?view=sql-server-2017 . Am I missing something or should I try after some time?

Thanks,
TK

rothja:
@raghavt-msft So, the process is not immediate, but it should be live in about 24 hours or less. What happens is that these public repo PRs get pulled into our private repository once a day. Then they get published on one of the twice-a-day publishing windows. So depending on when the change is merged, it could be a few hours to a day before the change published live. Sorry for the confusion/delay, but this will get published by tomorrow. If not, please contact me again and I'll investigate.

PR #751, opened 2018-06-06
In correct information existing powerpivot model

rajukum:
If the database exists and we try to restore we would get an error. “Object cannot be restored because the object already exists.  Please select a different database name.. There was Defect 743287:Remove Override checkbox in restore from PowerPivot filled to change the behavior. But the documentation suggest we can override an existing database which is not true.


PRMerger10:
@rajukum : Thanks for your contribution! The author, @minewiskan, has been notified to review your proposed change.

Minewiskan:
@rajukum Thanks for contributing!
#sign-off

PR #753, opened 2018-06-06
Update register-a-service-principal-name-spn-for-a-report-server.md

vanto28:
I removed any references to a port requirement for an HTTP SPN. This has been proven to be false, as adding a port to an HTTP SPN would make it unique, and in the case of Reporting Services, it will be ignored. In the Network traces, you can see that the requested SPN is without the port. If the HTTP SPN is added with a port, the error that will come back would be a missing SPN, or go through NTLM as it's not able to find the correct SPN.

I also changed #7 to reflect "Kerberos" instead of "NTLM"


PRMerger12:
@vanto28 : Thanks for your contribution! The author, @markingmyname, has been notified to review your proposed change.

markingmyname:
#sign-off

PR #762, opened 2018-06-08
Fixed case of `Definition` -> `definition`

leidegre:


craigg-msft:
Thank you @leidegre .

PR #788, opened 2018-06-14
Decryption by Password definition

dbaduck:
I believe there should be a callout that this clause is not necessary if the certificate is encrypted by the master key.


PRMerger10:
@dbaduck : Thanks for your contribution! The author, @edmacauley, has been notified to review your proposed change.

craigg-msft:
Thank you @fbsolo

PR #813, opened 2018-06-19
The option ERRORFILE_DATASOURCE is incorrect, it should be ERRORFILE_…

raghavt-msft:
…DATA_SOURCE


PRMerger15:
@raghavt-msft : Thanks for your contribution! The author, @edmacauley, has been notified to review your proposed change.

craigg-msft:
Thank you @raghavt-msft

PR #817, opened 2018-06-20
Remove the "constant" restriction

jeroen-mostert:
Since at least SQL Server 2012, `NTILE` supports arbitrary expressions in the argument, not merely constants. If the behavior was different in 2008/2008R2, that should be noted (I have no instances to test with). The only restriction is that the argument cannot refer to columns in the FROM resultset. (It may be worth mentioning that separately in the remarks.)


PRMerger15:
@jeroen-mostert : Thanks for your contribution! The author, @edmacauley, has been notified to review your proposed change.

craigg-msft:
@jeroen-mostert thank you

PR #821, opened 2018-06-25
Add default for HeapTypeForObjects

matt40k:
Add the default option for HeapTypeForObjects

### DIFF
--- a/docs/analysis-services/server-properties/memory-properties.md
+++ b/docs/analysis-services/server-properties/memory-properties.md
@@ -98,6 +98,7 @@ When set to 1, processing is less likely to fail due to memory constraints becau
   
    Setting | Description
 --------|------------
+**-1** | (default) Automatic. The engine will decide which one to use.
 **0** | Windows LFH heap.
 **1** | Analysis Services slot allocator.
 **3** | Each object has its own Analysis Services Heap.


craigg-msft:
@Minewiskan please review

Minewiskan:
@matt40k Yes. This one should be included. Thanks for contributing, Matt.
#sign-off

amirchandani:
VertipaqPagingPolicy should say:
0 --> (default for Azure AS)
1 --> (default for Analysis Services)


In description, Azure AS should mention that VPP=1 should not be used due to the hard job limits on the AS process in the service. These hard limits can cause out of memory errors when paging is enabled, and therefore customers should stay with VPP=0 in AAS.

matt40k:
Hey @amirchandani!
This isn't the right place to raise this bug as this was for a different bug and it's for "pulling" the fix. 
You should look at creating a new pull request - info here https://docs.microsoft.com/en-gb/sql/sql-server/sql-server-docs-contribute?view=sql-server-2017

amirchandani:
Thanks @matt40k, my mistake - I think @Minewiskan is taking care of this issue...

Minewiskan:
@matt40k @amirchandani I'm a bit confused by which fix is which. Can you both take a look at https://docs.microsoft.com/en-us/sql/analysis-services/server-properties/memory-properties and make sure what you're asking for is shown?  
Thanks

matt40k:
Looks good to me 👍

PR #824, opened 2018-06-26
Update while-transact-sql.md

nardnob:
Removed line 123 "SELECT MAX ( ListPrice) FROM dbo.DimProduct"


PRMerger14:
@nardnob : Thanks for your contribution! The author, @douglaslMS, has been notified to review your proposed change.

douglaslMS:
@nardnob I see why you might consider the line that you removed unnecessary. It isn't critical, you're right, but I think it makes sense there for someone testing and watching the sequence of steps:

1. Change the price.
2. Show me what the maximum price is after my changes.
3. If the maximum price is greater than X, break out of the loop.

So I'd prefer to leave that line in. What do you think? Thanks.

nardnob:
I see what you mean, I was thinking it was a typo. I agree it's probably best as-is for clarity's sake.

PR #825, opened 2018-06-26
Clarify the RECOMPILE option

PiJoCoder:
there is a "private" plan created for each execution of a query when RECOMPILE is used. That plan is discarded immediately after execution and does not affect any plan already cached for this query if query was run with no hints.


douglaslMS:
@PiJoCoder Joseph, the quotation marks around "private" cause a question or a problem at localization. Can you please remove the quotes? Also, is it possible that _temporary_ is a better word? Thanks.

douglaslMS:
@PiJoCoder ping

craigg-msft:
Thank you @PiJoCoder

PiJoCoder:
Thanks Craig and Douglas

PR #827, opened 2018-06-27
Provide real, working examples of row terminators

stansw:
I noticed that examples here do not work in SQL Server 2017 (14.0.2000.63)

Other people also run into this problem:
https://stackoverflow.com/questions/14365381/column-is-too-long-error-with-bulk-insert
https://blog.sqlauthority.com/2007/08/22/sql-server-t-sql-script-to-insert-carriage-return-and-new-line-feed-in-code/

```
BULK INSERT [dbo].[order_products__train]
FROM 'C:\Sources\MLADS\order_products__train.csv'
WITH
(
    FIRSTROW = 2,
    FIELDTERMINATOR = ',',
    ROWTERMINATOR = '\n',
    TABLOCK
)

Msg 4866, Level 16, State 8, Line 16
The bulk load failed. The column is too long in the data file for row 1, column 4. Verify that the field terminator and row terminator are specified correctly.
Msg 7301, Level 16, State 2, Line 16
Cannot obtain the required interface ("IID_IColumnsInfo") from OLE DB provider "BULK" for linked server "(null)".
```

```
BULK INSERT [dbo].[order_products__train]
FROM 'C:\Sources\MLADS\order_products__train.csv'
WITH
(
    FIRSTROW = 2,
    FIELDTERMINATOR = ',',
    ROWTERMINATOR = '0x0A',
    TABLOCK
)

(1384617 rows affected)
```

### DIFF
--- a/docs/relational-databases/import-export/specify-field-and-row-terminators-sql-server.md
+++ b/docs/relational-databases/import-export/specify-field-and-row-terminators-sql-server.md
@@ -1,4 +1,4 @@
-﻿---
+---
 title: "Specify Field and Row Terminators (SQL Server) | Microsoft Docs"
 ms.custom: ""
 ms.date: "08/10/2016"
--- a/docs/relational-databases/import-export/specify-field-and-row-terminators-sql-server.md
+++ b/docs/relational-databases/import-export/specify-field-and-row-terminators-sql-server.md
@@ -35,15 +35,14 @@ monikerRange: ">= aps-pdw-2016 || = azuresqldb-current || = azure-sqldw-latest |
   
 |Terminating character|Indicated by|  
 |---------------------------|------------------|  
-|Tab|\t<br /><br /> This is the default field terminator.|  
-|Newline character|\n<br /><br /> This is the default row terminator.|  
-|Carriage return/line feed|\r|  
-|Backslash*|\\\|  
-|Null terminator (nonvisible terminator)**|\0|  
+|Tab|`'0x0B'`<br /><br /> This is the default field terminator.|  
+|Newline / Line feed |`'0x0A'`<br /><br /> This is the default row terminator.|  
+|Carriage return|`'0x0D'`|  
+|Backslash|`'\'`|  
+|Null terminator (nonvisible terminator)**|`'0x00'`|  
 |Any printable character (control characters are not printable, except null, tab, newline, and carriage return)|(*, A, t, l, and so on)|  
 |String of up to 10 printable characters, including some or all of the terminators listed earlier|(**\t\*\*, end, !!!!!!!!!!, \t—\n, and so on)|  
-  
- *Only the t, n, r, 0 and '\0' characters work with the backslash escape character to produce a control character.  
+
   
  **Even though the null control character (\0) is not visible when printed, it is a distinct character in the data file. This means that using the null control character as a field or row terminator is different than having no field or row terminator at all.  
   
@@ -182,7 +181,7 @@ BULK INSERT myDepartment FROM 'C:\myDepartment-c-t.txt'
    WITH (  
       DATAFILETYPE = 'char',  
       FIELDTERMINATOR = ',',  
-      ROWTERMINATOR = '\n'  
+      ROWTERMINATOR = '0x0A'  
 );  
 GO  
 ```  


PRMerger14:
@stansw : Thanks for your contribution! The author, @douglaslMS, has been notified to review your proposed change.

craigg-msft:
@douglaslMS please review

douglaslMS:
@stansw @craigg-msft Here's what I've learned from the SQL Server engineering team: 

> The docs are correct as written.  This refers to the format files that BCP uses to interpret input files when importing a flat file into a table. Bcp has the option to generate a format file from an existing table, and when it does, the syntax used is what is in the current docs.

As a result, we prefer not to merge this pull request. Let me know if you have comments.

douglaslMS:
@stansw Thank you for your suggestion. Based on the response from the SQL Server engineering team, which I shared with you a couple days ago, we are going to close this pull request.

douglaslMS:
#please-close

stansw:
@douglaslMS, have you tired to use examples that currently exist in the documentation?

douglaslMS:
@StanislawSwierc I ran the examples on this page, which use BCP, [here](https://docs.microsoft.com/en-us/sql/2014/relational-databases/import-export/specify-field-and-row-terminators-sql-server?view=sql-server-2017#examples) and [here](https://docs.microsoft.com/en-us/sql/2014/relational-databases/import-export/specify-field-and-row-terminators-sql-server?view=sql-server-2017#examples-1). Notice that this article is part of a set of articles about BCP. If you're using C#, then I suppose that yes, the behavior might be different.

douglaslMS:
@StanislawSwierc To answer your specific question, I ran these example BCP commands from a Windows command prompt. I ran the CREATE TABLE and the BULK INSERT examples in SSMS, of course.

stansw:
Thank you. I see where the problem is. When you use BCP, then "\n" is correctly translated  a new line character. In T-SQL (SSMS), however, character '\' is not a special character and does not "escape" so "\n" doesn't get translated. 

Taking that into account the beginning of article, which talks about BCP is correct, but the code example at the end is broken. 

```SQL
USE AdventureWorks;  
GO  
BULK INSERT myDepartment FROM 'C:\myDepartment-c-t.txt'  
   WITH (  
      DATAFILETYPE = 'char',  
      FIELDTERMINATOR = ',',  
      ROWTERMINATOR = '\n'  
);  
GO  
```

It might be hard to explain 2 different (BCP and TSQL) terminators in a single article. Would you consider splitting it up?

douglaslMS:
@StanislawSwierc I'm confident that the BULK INSERT example worked for me as written, but I can test again tomorrow.

stansw:
@douglaslMS I'm confident that BULK INSERT example did not work for me on SQL Server 2017 (14.0.2000.63) and a 100MB csv file. That's when I got "The bulk load failed. The column is too long in the data file for row 1" error pasted in the first comment. 

When you try it tomorrow. Please keep track of:
* SQL Server version
* Input file
* T-SQL code
* Error message ;-)

douglaslMS:
#please-close

stansw:
@douglaslMS, I'm surprised. Could you please share (e.g. OneDrive):

* exact version of your SQL Server
* full T-SQL you executed (including table creation)
* input file

I suspect the root cause might be related to the size of the file. I've been working with pretty large files and maybe that's what triggered the error.

douglaslMS:
I followed the steps of the example exactly as described in the article. Yes, it’s a much smaller file than you’re using.

Microsoft SQL Server 2017 (RTM-GDR) (KB4057122) - 14.0.2000.63 (X64)
       Dec 22 2017 16:01:23
       Copyright (C) 2017 Microsoft Corporation
       Developer Edition (64-bit) on Windows 10 Enterprise 10.0 <X64> (Build 17134: )

stansw:
Found it! The problem is more complicated than what I expected! Is seems that both BCP and `BULK INSERT` are broken or weird in a sense that `\n` is not really `Line Feed` (`LF`), but an "*OS-specific line ending*", which on Windows is `Carriage Return` (`CR`) plus `Line Feed` (`LF`) or short `CRLF`.

Please see this article for the explanation of new lines and the globally accepted meaning of `\n` escape sequence.
https://en.wikipedia.org/wiki/Newline

Here are some experiments I run on my box. **Please notice that when I used `-r \n` options from the documentation I got `ASCII text, with CRLF line terminators`! Yes, `CRLF`!**

```
> bcp AdventureWorks2016CTP3.HumanResources.Department out C:\myDepartment-c-t.txt -c -t, -r \n -T

Starting copy...

16 rows copied.
Network packet size (bytes): 4096
Clock Time (ms.) Total     : 16     Average : (1000.00 rows per sec.)

> "c:\Program Files (x86)\GnuWin32\bin\file.exe" c:\myDepartment-c-t.txt
c:\Sources\myDepartment-c-t.txt; ASCII text, with CRLF line terminators

> bcp AdventureWorks2016CTP3.HumanResources.Department out C:\myDepartment-c-t.txt -c -t, -r \r -T

Starting copy...

16 rows copied.
Network packet size (bytes): 4096
Clock Time (ms.) Total     : 1      Average : (16000.00 rows per sec.)

> "c:\Program Files (x86)\GnuWin32\bin\file.exe" c:\myDepartment-c-t.txt
c:\Sources\myDepartment-c-t.txt; ASCII text, with CR line terminators
```

Coming back to my original problem. I had to load a file which was generated on Linus and it had true `\n` line ending `LF` and *not* the Windows specific `CRLF`. After reading documentation I thought that using `\n` will work, but SQL Server instead of interpreting it as `LF` decided (against my will) that I want `CRLF` and this of course did not work with my data.

The only solution I found to explicitly tell SQL Server that when I say `\n` I mean `LF` was to specify it in a byte encoding `0x0A`.

As you can see current documentation is great for developers who live and breathe in Windows ecosystem, but is painful to use when you work in a mixed environment with Windows *and* Linux. Please make it more inclusive for Linux developers. 

A good beginning would be acknowledging the strange behavior of `\n` character. Next, would be to give for developers who load data from Linux systems and IMHO this PR has a great solution for that.

stansw:
@douglaslMS, what do you think about making it clear that "\n" is not LF but CRLF and providing examples for LF?

douglaslMS:
@stansw I understand the issue now thanks to your careful testing and detailed notes. I am out of the office tomorrow and expect to get to this at the beginning of next week. I'll probably check with you about the additions and changes I make, but will probably do that in a separate PR of my own with the info you've provided here. Thank you!

douglaslMS:
@stansw I am going to email you the text that I propose adding to the article, based on the problem you've identified and the solution.

douglaslMS:
@stansw The text that I sent by email contained a copy/paste error - of course one can't use BULK INSERT to export, only bcp. When I specify `bcp -r '0x0A'`, are the single quotes required? No quotes are used if I specify `bcp -r \n`. Thanks.

stansw:
@douglaslMS sorry for the delay. Last week I was heads down hacking during //oneweek hackathon. Thank you for making the changes.

Text regarding importing from your email is correct. I would suggest changing "only" to "alone". For some reason it sounds better in my head.

>If you want to use a line feed character only (LF) as the row terminator

>If you want to use a line feed character **alone** (LF) as the row terminator

When it comes to export you correctly noticed that bcp.exe is the tool to use. As it turns out bcp.exe documentation already explains how to pass hex-encoded values.

>If you specify the row terminator in hexadecimal notation in a bcp.exe command, the value will be truncated at 0x00. For example, if you specify 0x410041, 0x41 will be used.

Since there are not whitespaces in the terminator string you can drop quotes. For example, the following command did the right thing when run on my box:
```
bcp dventureWorks2016CTP3.HumanResources.Department out C:\myDepartment-c-t.txt -c -t, -r 0x0A -T
```

PR #835, opened 2018-07-03
Added an example of granting the CONTROL permission

way0utwest:
The CONTROL permission is more complex and requires qualifying the object. Added an example to make this clear.


PRMerger19:
@way0utwest : Thanks for your contribution! The author, @CarlRabeler, has been notified to review your proposed change.

CarlRabeler:
#sign-off

PR #840, opened 2018-07-05
Clarified the DROP statement

way0utwest:
Added a note that a column cannot be dropped whether it is a key column or an include. There was a question by a user regarding ambiguity here.


PRMerger13:
@way0utwest : Thanks for your contribution! The author, @CarlRabeler, has been notified to review your proposed change.

CarlRabeler:
#sign-off

PR #848, opened 2018-07-10
Remove useless user and login

MindFlavor:
This PR removes the useless login and user from the script below. Both login and user are useless as the endpoint authentication is done using the shared certificate.
For a security perspective is best to avoid creating useless principals in a production server (particularly is the the same password is replicated many times). 
Also the script is shorter and cleaner this way.

Tested on `Microsoft SQL Server 2017 (RTM-CU8) (KB4338363) - 14.0.3029.16 (X64)   Jun 13 2018 13:35:56   Copyright (C) 2017 Microsoft Corporation  Developer Edition (64-bit) on Linux (Ubuntu 16.04.4 LTS)`


GitHubber17:
@MikeRayMSFT - Would you review the proposed changes? Thanks

MikeRayMSFT:
Thank you @MindFlavor for the submission! #sign-off

PR #853, opened 2018-07-11
add commend on AKV FW

yochananrachamim:
in case of AKV FW is enabled it will block access to the AKV and databases will be dropped, this is common mistake done by customers and this is not clear from the document that this is what is going to happen.


aliceku:
I agree a note about AKV FW is helpful, but here this note is inserted in the section about keys, not key vaults.  This needs to move up to the key vault section.  I can make the change in the right place if that helps.

yochananrachamim:
@aliceku - Thank you! That would be very helpful

aliceku:
I made the change, going through the process now

aliceku:
@yochananrachamim the change is now live you can delete this pull request.

PR #855, opened 2018-07-11
[CU9] SQL Server Agent is disabled if no conf entry

vin-yu:
starting in CU9:

If sqlagent.enabled is not present in the mssql.conf settings file, then SQL Server internally assumes that SQL Server Agent is disabled.


rothja:
#sign-off

PR #869, opened 2018-07-18
Update cast-and-convert-transact-sql.md

sqlsauce:
The first sentence in Example G mentions ISO 8901. That should be ISO 8601.


MashaMSFT:
Hi @sqlsauce, thank you so much for your contribution! You're right, this seems to be a typo :) #sign-off

PR #871, opened 2018-07-19
Updated endpoint to standard definition

SeanGallardy:
Change from "DATA_MIRRORING" to "DATABASE_MIRRORING" to standardize on the same nomenclature that is used in the CREATE ENDPOINT Doc article. While *technically* DATA_MIRRORING works (in on-prem situations) it's not the standard, updating to reflect this.


ktoliver:
@MikeRayMSFT - Would you review the proposed changes? Thanks.

MikeRayMSFT:
@SeanGallardy thank you for the submission. #sign-off

PR #873, opened 2018-07-20
Update select-group-by-transact-sql.md

fzbam:
missing space/linebreak in sample. as all other samples have GROUP BY in seperate line, line break added


rothja:
@fzbam Thanks for this. This is the right way to write the sample. Appreciate the feedback!

PR #876, opened 2018-07-23
Update columns-without-a-name.md

brunom:


PRMerger19:
@brunom : Thanks for your contribution! The author, @douglaslMS, has been notified to review your proposed change.

douglaslMS:
@brunom Thank you for fixing this could block. Could you please (1) remove the tick mark ` at the end of each line, and (2) at the "xml" language tag on the first line, so that instead of " ``` " we have " ```xml". Thanks.

PRMerger20:
@brunom : Thanks for your contribution! The author, @douglaslMS, has been notified to review your proposed change.

brunom:
The rest of the documentation doesn't use ```xml. Do you want me to author a bigger patch that applies the formatting more widely?

douglaslMS:
@brunom As far as adding the missing language tags ("sql" is missing too), why don't we limit the scope of this PR to the section that you revised. We have an internal report that identifies pages with missing language tags, so I think a vendor project is being consider to fix that all at once. Thanks.

brunom:
Great.  I have applied your corrections already.

douglaslMS:
#sign-off

@brunom Thank you!

PR #877, opened 2018-07-23
Update monitoring-performance-by-using-the-query-store.md

tgrohser:
The two examples had a slash before the <= resulting in a syntax error


MikeRayMSFT:
@tgrohser thank you for your submission. #sign-off

PR #887, opened 2018-07-31
Added formatting to headers

philipstratford:
It wasn't clear at first when reading this article that the two lines I've edited were headings because they weren't formatted as such.


PRMerger19:
@philipstratford : Thanks for your contribution! The author, @rothja, has been notified to review your proposed change.

rothja:
@philipstratford Thanks for the contribution. This does read better. I wonder if there needs to be a 2nd H2 header to break up the section of Types of DDL triggers? It seems that section also explains some of the behavior of triggers all-up. If you agree, feel free to submit another edit. Otherwise, this looks good.

#sign-off

PR #889, opened 2018-08-01
Swap titles to match examples

dragmz:


rothja:
@dragmz Thanks for fixing this! #sign-off

PR #890, opened 2018-08-01
very minor grammar fix

jarek91:
'point' needs to be plural since it follows 'several'.


rothja:
@jarek91 Thanks! Grammar fixes are never very minor. Appreciate you improving the docs! #sign-off

PR #893, opened 2018-08-02
Update best-practice-with-the-query-store.md

baleng:
forcing a plan via sp_query_store_force_plan  is not deterministic. in order to always have the forced plan we need a plan guide. 
 regardless of a forced plan in the query store, if something is changed  (for example the number of rows ) , a new plan is buit.
 this happens very often with huge tables , in attachment an example; the flagged circle is the forced plan, and it is not used.
forcedplanko


PRMerger8:
@baleng : Thanks for your contribution! The author, @MikeRayMSFT, has been notified to review your proposed change.

PRMerger17:
@baleng : Thanks for your contribution! The author, @MikeRayMSFT, has been notified to review your proposed change.

MikeRayMSFT:
@baleng - thank you for the submission - reviewing

PR #896, opened 2018-08-06
Added statement when MAX returns NULL

Nabid:
Added statement: when there is no row to select, MAX returns NULL


PRMerger7:
@Nabid : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
Hi @Nabid, awesome, thank you for the contribution! You are correct, and it helps to specify it :) #sign-off

PR #898, opened 2018-08-06
Fixed FCI content

allanhirt:
It was never fail over clustered instance, and virtual server name is no longer a thing; hasn't been for over 10 years. If you have questions, talk to Mike Ray.


MashaMSFT:
Hi @allanhirt, thanks for the edit! #sign-off

PR #923, opened 2018-08-16
CASE (Transact-SQL) Wording consistency

LarnuUK:
In the Remarks section, CASE is referred to as a "CASE statement" (apart from once), everywhere else it is referred to as an Expression. Consistency is important.


douglaslMS:
#sign-off

PR #925, opened 2018-08-18
Update xml-format-files-sql-server.md

DoranDouglas:
In the fixed width example, the closing "/>" was missing for the line.


douglaslMS:
@DoranDouglas The FIELD element and its attributes are wrapped across two lines, as in the immediately preceding FIELD elements. We often do this in the docs to avoid horizontal scrolling.

DoranDouglas:
Ah, on the docs.microsoft.com site it is missing that next line.

https://docs.microsoft.com/en-us/sql/relational-databases/import-export/xml-format-files-sql-server?view=sql-server-2017#SampleXmlFFs

douglaslMS:
@DoranDouglas OK, after looking at all the examples closely, I see that there **is** in fact a similar line that's missing the closing "/>" just as you said ... only I think you may have fixed the wrong one in your PR. The error is in the very last example, F. Importing fixed-length or fixed-width fields. Could you please fix this in your PR? It looks like you added the closing sequence to A. or B. instead of F.

douglaslMS:
@DoranDouglas Do you agree with my suggested change in your pull request?

DoranDouglas:
I do, I was delayed in acting last night.

PR #926, opened 2018-08-18
Correct copy and pasted query

rsanderson2350:
This section of the documentation discussed disabling Managed Backup as a whole, at the Instance Level.  Specifically, the text in this sections paragraph states the need to set the database_name parameter to NULL; however, the SQL Syntax was copy and pasted from above in this document and was not updated to reflect this needed parameter value.  This update is to show this parameter being executed as NULL, as to apply to the whole SQL Instance.


PRMerger9:
@rsanderson2350 : Thanks for your contribution! The author, @MikeRayMSFT, has been notified to review your proposed change.

MikeRayMSFT:
@rsanderson2350 thank you for your submission. #sign-off

PR #927, opened 2018-08-20
Update columnstore-indexes-overview.md

baleng:
row 144: |Convert a columnstore table to a rowstore.|
set OR in lowercase and set the link to ../../t-sql/statements/create-columnstore-index-transact-sql.md  where there is an actual  example of the t-sql


craigg-msft:
Thank you @baleng.  merging. fyi @MikeRayMSFT

PR #934, opened 2018-08-22
IGNORE_DUP_KEY clarification

claudiutomescu:
The original text is confusing implying an error response for this option, even if the next paragraphs for ON/OFF options are a bit more clear. I think the proposed change better reflects the intent of this option.


PRMerger12:
@claudiutomescu : Thanks for your contribution! The author, @CarlRabeler, has been notified to review your proposed change.

CarlRabeler:
#sign-off

PR #938, opened 2018-08-22
Fix minor grammar issues

KompotentKev:
Line 30 -  "your query was sent" rather than "you query sent"
Line 37 - Off topic sub headings, "Your program must distinguish between 'transient errors' versus 'persistent errors.'  The paragraph is discussing   these two topics but the start of the next sentence refers back to a previous topic not the sub topic, - Transient faults. Optionally to avoid double wording to change to "Transient errors are fault conditions..."
Line 269 - Wrong word used - which rather than with


PRMerger7:
@KompotentKev : Thanks for your contribution! The author, @MightyPen, has been notified to review your proposed change.

MightyPen:
@PRMerger7 I like these changes, and I recommend #sign-off.
CC @KompotentKev Thanks Kevin.

PR #947, opened 2018-08-28
Fix typo in minor_id column description

kappa7194:
The description for the minor_id column of the sys.database_permissions system catalog view refers to major_id instead of minor_id.


VanMSFT:
#sign-off

PR #948, opened 2018-08-28
Script to create role and permissions

LarnuUK:
The instructions, although comprehensive, are SSMS specific. This isn't much use to those using other tools, such as SOS, sqlcmd and mssql-cli, etc. The above also requires a lot of button clicking, which isn't quick. A script is a far quicker way of doing this, and much easier of doing this (and non environment specific) it should be provided.

### DIFF
--- a/docs/reporting-services/security/create-the-rsexecrole.md
+++ b/docs/reporting-services/security/create-the-rsexecrole.md
@@ -230,6 +230,38 @@ manager: "kfile"
   
 15. Click the link to open Report Manager. You should see the report server items from the report server database.  
 
+## Creating the RSExecRole Role and permissions using T-SQL
+The role can also be created, and applicable permissions granted, on the system databases using the following T-SQL script:
+```sql
+USE master;
+GO
+IF NOT EXISTS (SELECT 1 FROM sys.database_principals WHERE [type] = 'R' AND [name] = 'RSExecRole') BEGIN
+    CREATE ROLE [RSExecRole];
+END
+GRANT EXECUTE ON dbo.xp_sqlagent_enum_jobs TO [RSExecRole];
+GRANT EXECUTE ON dbo.xp_sqlagent_is_starting TO [RSExecRole];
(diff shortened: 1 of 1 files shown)


PRMerger10:
@LarnuUK : Thanks for your contribution! The author, @markingmyname, has been notified to review your proposed change.

markingmyname:
@LarnuUK Thank you for your feedback! We will investigate and review if we need to change/add the information suggested.

LarnuUK:
Hi @markingmyname , I'm not sure what you're asking here. This PR _does_ contain a script to add to use instead of the need for all the button clicks. If O were to submit another PR it would be identical to what I have already done; which seems somewhat pointless.

Thanks,

markingmyname:
@LarnuUK I apologize, I must have overlooked it.  I see it now.  Let me review it and get back to you.  Thank you for the submission.

LarnuUK:
Thanks @markingmyname.

markingmyname:
@LarnuUK I agree.  I looked over it and thank you for this info.  However, I would like to change the format a bit.  So I am going to create another PR. about this.  Can you close this one, please?

LarnuUK:
Hi @markingmyname  I could do this, but part of the idea here is that others, including yourself, change change the PR, or you can ask me to do so. If I close this PR, and you open one then i lose my contribution. This might only seem small, but that is a slight on the user's who are contributing to the articles if you are going to remove that from them.

What is it you want changing? I'd be happy to adjust the code if it's simply formatting changes. Otherwise feel free to do this yourself if you wish; I've had other contributions have their wording changed to fit with MS' template; why would this not the case here?

markingmyname:
@LarnuUK I did not look at it from that perspective.  I was not trying to diminish your contribution or exclude you from anything.  I do apologize if that is how it came across. After being unsuccessful in editing your PR, it was just easier for me to create a new one. That's all. If you can allow me to edit your PR, then I do not mind doing it within this PR.

LarnuUK:
@markingmyname you should already be able to do so; the PR is assigned to you.

markingmyname:
#sign-off

PR #950, opened 2018-08-29
Update create-an-entity-master-data-services.md

Ashutosh-A:
Added a Tip, which was not specified in documentation earlier. In many cases, the Staging table names are not same as Entity name.
Text Added : 
If using the default naming for Staging tables, MDS will automatically append identifiers (e.g. _1, _2) to the staging table names if an entity with same name exists in another Model.


PRMerger10:
@Ashutosh-A : Thanks for your contribution! The author, @leolimsft, has been notified to review your proposed change.

craigg-msft:
Thank you @Ashutosh-A

PR #951, opened 2018-08-29
Update create-trigger-transact-sql.md

CalvinAllen:
"Previous example" is incorrect, as the trigger in question is three examples above.


PRMerger13:
@CalvinAllen : Thanks for your contribution! The author, @CarlRabeler, has been notified to review your proposed change.

PR #955, opened 2018-08-31
Trying to clarify the distinction between Public and the other fixed …

davidbaxterbrowne:
…server roles

Trying to clarify the distinction between server roles with fixed permissions, and Public which has fixed membership and non-fixed permissions.


PRMerger6:
@davidbaxterbrowne : Thanks for your contribution! The author, @VanMSFT, has been notified to review your proposed change.

VanMSFT:
Hi @davidbaxterbrowne , thanks for your contribution! You're right, this helps with added clarity. Thanks again!
#sign-off

PR #958, opened 2018-09-03
Update sys-dm-db-index-physical-stats-transact-sql.md

baleng:
to me it seems the value 0 in "columnstore_delete_buffer_state" is NOT VALID in "columnstore_delete_buffer_state_desc" column


stevestein:
@baleng - thanks for the edit. I'm also seeing NOT VALID returned.

#sign-off

PR #967, opened 2018-09-11
Formatting on SORT_IN_TEMPDB option for indexes

jpomfret:
Change so the sentence before the bullets is not indented to the bullet level:

Currently looks like this, change is to removed the highlighted whitespace:

image


MikeRayMSFT:
@jpomfret thank you for your contribution! #sign-off

PR #978, opened 2018-09-16
Fixing number of Gen2 polybase readers

furmangg:
In Gen1 DWs, it's 8 polybase readers per node but in Gen2 DWs it's 20 polybase readers per node. I'm not sure what SQL2017 is but I assume 8 given the current state of the documentation saying 8 as a blanket statement.


CarlRabeler:
#sign-off

PR #991, opened 2018-09-20
Update replication-with-sql-database-managed-instance.md

alvorob:
We need to address documentation discrepancy,  leading to misconfiguration, when customers using Access Key instead of Connection String.


PRMerger13:
@alvorob : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
#sign-off

PR #1001, opened 2018-09-25
Fix folder and .EXE name

claudiutomescu:
Fix folder and EXE name for Windows .zip file version. The doc is still pointing to SQL Ops Studio instead of new Azure Data Studio


claudiutomescu:
I haven't tested on MacOS or Linux, but I'm wondering if those need to be fixed as well as they still point to SQL Ops Studio name

stevestein:
@claudiutomescu - thanks for the edit!

PR #1012, opened 2018-09-29
Corrected query

guzmanda:
Added comma between state and state_desc so that the state_desc column value is returned instead of being interpreted as an alias for the preceding state column.


PRMerger9:
@guzmanda : Thanks for your contribution! The author, @MikeRayMSFT, has been notified to review your proposed change.

MikeRayMSFT:
@guzmanda - thank you!

#sign-off

PR #1014, opened 2018-09-30
Add namespace creation, and add namespace to creation of the secret

dbadavekc:
The deployment for sqlserver.yaml does not create the ag1 namespace, and instead will error if the namespace is not already created. Obviously many ways to do this, but I just added the command in-line

As the referenced examples deploy in to a namespace called "ag1", we need to make sure we're creating this secret in that namespace. As written, it would create it in the default namespace, which would prevent the deployed sql server manifest from finding it.


PRMerger18:
@dbadavekc : Thanks for your contribution! The author, @MikeRayMSFT, has been notified to review your proposed change.

MikeRayMSFT:
@dbadavekc - thank you for this submission. It was timely and helpful. The technical issue was corrected in a different pull request. Please keep your submissions coming.

#please-close

PR #1017, opened 2018-10-01
Create Linked Servers after adding a replica

KankuruSQL:
After adding a replica to distribution database AG, we must also create linked servers to all subscribers to avoid Agent message code 20053. Server 'YYY' is not registered at server 'XXX'.


PRMerger13:
@KankuruSQL : Thanks for your contribution! The author, @MikeRayMSFT, has been notified to review your proposed change.

MikeRayMSFT:
#sign-off

PR #1019, opened 2018-10-01
Removed unnecessary div reference

caseydierking:
I removed the div class reference here because it is not needed. The list is a bullet list.


PRMerger17:
@caseydierking : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

caseydierking:
I just realized what this div reference was for. I was only viewing the documentation on github. I see now that it is necessary for displaying it as a checklist on the Microsoft Docs site.

PR #1028, opened 2018-10-03
Versions in prerequisite and link don't match

KevinAdams05:
The text in the Prerequisite column says 2017, but the link goes to 2016.


PRMerger7:
@KevinAdams05 : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

KevinAdams05:
Hi @MightyPen, I double checked my diff and I'm not sure where you are seeing that it says 2016 in the left column. In both the the before and after, the diff shows that the variable [!INCLUDE[ssCurrent] is used for the left column. It's only the URL and URL description in the right column that I updated.

screenshot_20181006_173644

MashaMSFT:
Hi @KevinAdams05 , thank you so much for your contribution! You're right, it looks like this is a byproduct of versioning. Due to the complexity of the change, I fixed it within another PR, and am closing out this one. I will leave this PR open until the change goes live, and notify you :). Thanks!

MashaMSFT:
Hi @KevinAdams05 , thanks again for your contribution! This change has gone live here: https://docs.microsoft.com/en-us/sql/database-engine/availability-groups/windows/prereqs-restrictions-recommendations-always-on-availability?view=sql-server-2017. Please let me know if this fails to address your concern? Thanks again!

PR #1038, opened 2018-10-05
Correct Lock Timeout behavior

davidbaxterbrowne:
On lock timeout no error is sent; you can't TRY/CATCH or examine @@error.  The proc just returns -1 and you must check the return code.


PRMerger18:
@davidbaxterbrowne : Thanks for your contribution! The author, @stevestein, has been notified to review your proposed change.

davidbaxterbrowne:
Here's a quick repro.  Run this from multiple sessions and you'll see the second timeout without an error

```
begin transaction
declare @rc int
exec @rc = sp_getapplock 'only_one','exclusive','transaction',0
select @rc

--rollback transaction
```

MikeRayMSFT:
#sign-off

PRMerger6:
@MikeRayMSFT: I'm sorry - only the author of this article, @stevestein, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #1042, opened 2018-10-06
Clarify that encryption is bi-directional

sensomatic:
Encryption in public key cryptography is bi-directional. In the certificate case it is the private key encrypting data that is important and forms the basis of digital certificates and non-repudiation of authorship.


PRMerger14:
@sensomatic : Thanks for your contribution! The author, @VanMSFT, has been notified to review your proposed change.

PRMerger14:
@MightyPen: I'm sorry - only the author of this article, @VanMSFT, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #1045, opened 2018-10-06
Minor improvement in README file

Akshat-Jain:
I think that `docs.microsoft.com - Terms of Use` is more suitable than the current `docs.microsoft.com Terms of Use`.


craigg-msft:
@Akshat-Jain thank you

PR #1116, opened 2018-10-11
When connecting to a primary replica, data for secondaly may not be current

taroarai:
AG Dashboard should indicate same restriction, since the dashboard's information is based on sys.dm_hadr_database_replica_states with following restriction.

Important
Depending on the action and higher-level states, database-state information may be unavailable or out of date. Furthermore, the values have only local relevance. For example, on the primary replica, the value of the last_hardened_lsn column reflects the information about a given secondary database that is currently available to the primary replica, not the actual hardened LSN value that the secondary replica might have currently.


taroarai:
+
@MashaMSFT

MashaMSFT:
Hi @taroarai , thanks for your contribution! #sign-off

PR #1169, opened 2018-10-25
Update sql-server-docs-contribute.md

HMuhammad:
Sentence previously read "To find the correct applies-to include for your article, use the following tips:".
Changed it to "To find the correct applies-to include file for your article, use the following tips:" for clarity/consistency with other verbiage in the article


ktoliver:
Merging grammar fix

PR #1174, opened 2018-10-26
This works on Azure SQL Database

davidbaxterbrowne:
Don't know if it's supposed to, or is supported.  But it works.  You can capture the event in a XEvent trace, so it's also useful.


CarlRabeler:
@davidbaxterbrowne These procs are specific to SQL Profiler. They haven’t been tested for Extended events (even if they appear to work). The supported way is to use Extended Event DDL to add events.
See https://docs.microsoft.com/en-us/azure/sql-database/sql-database-xevent-db-diff-from-svr, https://docs.microsoft.com/en-us/azure/sql-database/sql-database-xevent-code-event-file, and https://docs.microsoft.com/en-us/azure/sql-database/sql-database-xevent-code-ring-buffer.

PR #1182, opened 2018-10-29
Made all user steps bold

dbergmans:
Made all steps that the user should do bold.


Minewiskan:
@dbergmans I cannot approve this. This change does not meet our formatting guidelines. In procedures, bold is reserved for references to text in the UI.

PR #1184, opened 2018-10-29
Fix incorrect "syntax for SQL Server" heading

AndrewFauldsMsft:
Heading incorrectly says "Syntax for SQL Server" when it should say "Syntax for Azure SQL Data Warehouse". This article does not apply to SQL Server, which is correctly marked with the checkboxes under the main title.


PRMerger10:
@AndrewFauldsMsft : Thanks for your contribution! The author, @uc-msft, has been notified to review your proposed change.

craigg-msft:
thank you @AndrewFauldsMsft

PR #1191, opened 2018-10-31
add a known issue

MSraye:


stevestein:
@MSraye - looks like this is a duplicate of https://github.com/MicrosoftDocs/sql-docs/pull/1192 

1192 was just merged, so you can go ahead and close this. 

Also, in the future, if you can make updates in the private repo that would be better. Thanks!

PR #1194, opened 2018-10-31
Document mapping of input to output types

jeroen-mostert:
The original proposed that the input type is the same as the output type -- this is only mostly correct. The example of **decimal(2,0)** is especially misleading. The results in this table were experimentally established using SQL Server 2012 and 2017.


PRMerger9:
@jeroen-mostert : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
Hi @jeroen-mostert , thank you so much for your contribution! You're right, this helps to add clarity. Thanks again! #sign-off

PR #1210, opened 2018-11-02
Remove spaces in the Feature Parameter Examples

PiJoCoder:
As the documentation, there should be no spaces in the feature list:
"/FEATURES, which is a multivalued parameter, but its format is /FEATURES=AS,RS,IS without a space, comma-delimited "


MashaMSFT:
Hi @PiJoCoder, awesome, thanks for noticing these :) #sign-off

PR #1215, opened 2018-11-04
Update dbcc-show-statistics-transact-sql.md

baleng:
the upper bound is excluded only if distinct_range_Rows and range_rows are > 0 , if in the step there is/are only row(s) from the upper bound's value (only 1 distinct value ) the DISTINCT_RANGE_ROWS is 1 not 0. So the upper bound is not always excluded, but the value 1 is set for it.


craigg-msft:
@uc-msft can you please review.

craigg-msft:
@baleng apologies for the delay.  Your suggested change was reviewed and a slightly modified version is recommended.  Could you please update your PR with this change?  @uc-msft 

"Average number of rows with duplicate column values within a histogram step, excluding the upper bound. When DISTINCT_RANGE_ROWS is greater than 0, AVG_RANGE_ROWS is calculated by dividing RANGE_ROWS by DISTINCT_RANGE_ROWS.  When DISTINCT_RANGE_ROWS is 0, AVG_RANGE_ROWS returns 1 for the histogram step."

PRMerger16:
@baleng : Thanks for your contribution! The author, @uc-msft, has been notified to review your proposed change.

baleng:
cc @craigg-msft 
hello, PR modified, regards !

craigg-msft:
looks good.  @uc-msft I will merge.

PR #1225, opened 2018-11-07
Update the-memory-optimized-filegroup.md

baleng:
after creation (before using it) the container and the filegroup can be removed, once used is not possibile.


PRMerger6:
@baleng : Thanks for your contribution! The author, @CarlRabeler, has been notified to review your proposed change.

PR #1234, opened 2018-11-09
Added clarifying example of adding PK and compression

greglow:
I keep coming across users who don't realize that you can add a data compression clause when adding a clustered primary key. Instead, they rebuild the clustered index, after they add it. This is very efficient. I've added a clarifying example to the previous PK example to show that you can do both. I've also changed a few references from AdventureWorks2012 to just AdventureWorks. Having the 2012 in place adds nothing, and just dates the content.


PRMerger16:
@greglow : Thanks for your contribution! The author, @CarlRabeler, has been notified to review your proposed change.

PR #1238, opened 2018-11-09
Odd behaviour of non-quoted parameter values

alansingfield:
CREATE PROC demo_1
    @arg nvarchar(10)
AS
BEGIN
    SELECT @arg
END
GO

DECLARE @message nvarchar(10) = 'hello';
EXEC demo_1 message
EXEC demo_1 @message

----------
message

----------
hello

Accidentally omitting the @ from a variable name causes EXEC to treat the text supplied as an nvarchar string.
I would hope that this feature could be deprecated as it causes much confusion!

### DIFF
--- a/docs/t-sql/language-elements/execute-transact-sql.md
+++ b/docs/t-sql/language-elements/execute-transact-sql.md
@@ -230,6 +230,8 @@ Execute a character string
   
  If the value of a parameter is an object name, character string, or qualified by a database name or schema name, the whole name must be enclosed in single quotation marks. If the value of a parameter is a keyword, the keyword must be enclosed in double quotation marks.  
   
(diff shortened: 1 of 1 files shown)


gbritton1:
That's a long standing bug.  IIRC It can't be fixed since it will break code that depends on it, intentionally or not.

alansingfield:
Yes - I can see exactly why you wouldn't fix it, it's just the documentation that should be updated to warn you about the unusual behaviour...

gbritton1:
The other thing is, MS doesn't want to advertise this "feature" or encourage anyone to use it.  It is a bug that crept in some time ago.

douglaslMS:
@sign-off

PR #1250, opened 2018-11-11
Update set-showplan-text-transact-sql.md

gbritton1:
Change reference to osql to refer to sqlcmd instead, since osql is deprecated


PRMerger20:

PR #1258, opened 2018-11-12
Update examples-of-bulk-access-to-data-in-azure-blob-storage.md

FonsecaSergio:
Put a note regarding the case sensitive.

>  [!IMPORTANT]
>  All the paths to the container and to the files on blob are `CASE SENSITIVE`. If not correct, it might return error like "Cannot bulk load. The file "file.csv" does not exist or you don't have file access rights."
"


PRMerger9:
@FonsecaSergio : Thanks for your contribution! The author, @CarlRabeler, has been notified to review your proposed change.

PR #1272, opened 2018-11-14
typo: connectin -> connection

nschonni:
Added code fence language as well


MightyPen:
@nschonni Nick, I very much like that you added the language tag 'php' to the 3-backtick lines that open those PHP code blocks.    I am bulk fixing some things in our repo, but bulk fixing the missing language tags is probably implausible.

#sign-off

nschonni:
Running Markdownlint with everything off but this rule should help the cleanup https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md#md040
Note that I find the globbing of files one the commanline doesn't seem to work properly in that library so you might need to go folder by folder

PR #1279, opened 2018-11-15
Sync datetime literal recommendations across docs

jzabroski:
Identical to advice in the following pages, so I copied and pasted the content from the second page (as it seemed to be the most robust/mature/latest):
https://docs.microsoft.com/en-us/sql/relational-databases/views/create-indexed-views
> Expressions that involve implicit conversion of character strings to datetime or smalldatetime are considered nondeterministic. This is because the results depend on the `LANGUAGE` and `DATEFORMAT` settings of the server session. For example, the results of the expression `CONVERT (datetime, '30 listopad 1996', 113)` depend on the `LANGUAGE` setting because the string `'listopad'` means different months in different languages. Similarly, in the expression `DATEADD(mm,3,'2000-12-01')`, SQL Server interprets the string `'2000-12-01'` based on the `DATEFORMAT` setting. Implicit conversion of non-Unicode character data between collations is also considered nondeterministic.

https://docs.microsoft.com/en-us/sql/relational-databases/indexes/indexes-on-computed-columns
> **Determinism Requirements**  
>  
> > [!IMPORTANT]
> >  When you refer to string literals of the date data type in indexed computed columns in [!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)], we recommend that you explicitly convert the literal to the date type that you want by using a deterministic date format style. For a list of the date format styles that are deterministic, see [CAST and CONVERT](../../t-sql/functions/cast-and-convert-transact-sql.md). 
>
> > [!NOTE]
> > Expressions that involve implicit conversion of character strings to date data types are considered nondeterministic, unless the database compatibility level is set to 80 or earlier. This is because the results depend on the [LANGUAGE](../../t-sql/statements/set-language-transact-sql.md) and [DATEFORMAT](../../t-sql/statements/set-dateformat-transact-sql.md) settings of the server session. 
> >
> > For example, the results of the expression `CONVERT (datetime, '30 listopad 1996', 113)` depend on the LANGUAGE setting because the string '`30 listopad 1996`' means different months in different languages. 
> > Similarly, in the expression `DATEADD(mm,3,'2000-12-01')`, the [!INCLUDE[ssDE](../../includes/ssde-md.md)] interprets the string `'2000-12-01'` based on the DATEFORMAT setting.  
> >   
> > Implicit conversion of non-Unicode character data between collations is also considered nondeterministic, unless the compatibility level is set to 80 or earlier.  
> >   
> > When the database compatibility level setting is 90, you cannot create indexes on computed columns that contain these expressions. However, existing computed columns that contain these expressions from an upgraded database are maintainable. If you use indexed computed columns that contain implicit string to date conversions, to avoid possible index corruption, make sure that the LANGUAGE and DATEFORMAT settings are consistent in your databases and applications.

### DIFF
--- a/docs/t-sql/statements/create-partition-function-transact-sql.md
+++ b/docs/t-sql/statements/create-partition-function-transact-sql.md
@@ -61,7 +61,20 @@ FOR VALUES ( [ boundary_value [ ,...n ] ] )
   
 > [!NOTE]  
 >  If *boundary_value* consists of **datetime** or **smalldatetime** literals, these literals are evaluated assuming that us_english is the session language. This behavior is deprecated. To make sure the partition function definition behaves as expected for all session languages, we recommend that you use constants that are interpreted the same way for all language settings, such as the yyyymmdd format; or explicitly convert literals to a specific style. To determine the language session of your server, run `SELECT @@LANGUAGE`.  
-  
+
+> [!IMPORTANT]  
+>  When you refer to string literals of the date data type in indexed computed columns in [!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)], we recommend that you explicitly convert the literal to the date type that you want by using a deterministic date format style. For a list of the date format styles that are deterministic, see [CAST and CONVERT](../../t-sql/functions/cast-and-convert-transact-sql.md). 
+
+> [!NOTE]
+> Expressions that involve implicit conversion of character strings to date data types are considered nondeterministic, unless the database compatibility level is set to 80 or earlier. This is because the results depend on the [LANGUAGE](../../t-sql/statements/set-language-transact-sql.md) and [DATEFORMAT](../../t-sql/statements/set-dateformat-transact-sql.md) settings of the server session. 
+>
+> For example, the results of the expression `CONVERT (datetime, '30 listopad 1996', 113)` depend on the LANGUAGE setting because the string '`30 listopad 1996`' means different months in different languages. 
+> Similarly, in the expression `DATEADD(mm,3,'2000-12-01')`, the [!INCLUDE[ssDE](../../includes/ssde-md.md)] interprets the string `'2000-12-01'` based on the DATEFORMAT setting.  
+>   
+> Implicit conversion of non-Unicode character data between collations is also considered nondeterministic, unless the compatibility level is set to 80 or earlier.  
+>   
+> When the database compatibility level setting is 90, you cannot create indexes on computed columns that contain these expressions. However, existing computed columns that contain these expressions from an upgraded database are maintainable. If you use indexed computed columns that contain implicit string to date conversions, to avoid possible index corruption, make sure that the LANGUAGE and DATEFORMAT settings are consistent in your databases and applications.  
+
  *...n*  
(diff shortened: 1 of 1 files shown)


jzabroski:
@robconery Meta-comment and feedback for your to consider: It would be ideal if this information was extracted into a custom markdown page, `AvoidNonDeterministicLiterals.md`, and there was a way to "import" the common page, similar to a SASS/LESS mixin. I don't think GitHub supports such mixins currently.

TrishaMc:
Hi @craigg-msft: 
Would you review the proposed changes? 
Thanks,
Trisha

jzabroski:
@TrishaMc @craigg-msft Sorry, I just found https://github.com/MicrosoftDocs/sql-docs/tree/live/docs/includes/paragraph-content

I think ideally we refactor all three pages to use a common include - but to be honest, as an external contributor, the includes thing is totally unwieldy:
1.  The names of includes are hard to understand,
2. I'm not sure how to follow the naming convention for includes/*.md files, since it seems to be all over the map and cryptic.
3. avoid-non-deterministic-datetime-literals.md would be my choice but I am not sure how to actually include it in the documents once I'm done

robconery:
Hi - why am I being pinged? I can't tell from context.

jzabroski:
@robconery Sorry, I didn't realize there was already a way to dynamically include content in Microsoft docs.  My understanding is that your new position at Microsoft is to make developers' lives as pain-free and awesome as possible, and a huge portion of that is how we interact with documentation and samples and give feedback to tighten the feedback cycle to correct errors and improve products.

EisenbergEffect:
@jzabroski I don't actually work at Microsoft anymore. Sorry! That said, I can tell you that there are extensions that enable Markdown includes of fragments. At the time when I worked at Microsoft, there was not something specifically for things like variables. However, that was in the backlog. You'll need to reach out to one of the current PMs to learn more about the status of that. You could talk to one of my former direct reports to get the latest. Try @ShevaDas, @yunyunyunyun or @cillroy

MightyPen:
@jzabroski Hi JZabroski, after considering the nice detail that this public PR 1279 provided to us, we discussed the matter. We decided to take corrective action on the problem you found, but to implement the improvement in a different manner. So we are Closing this PR 1279, and have created private PR 8174 (plus 8175) instead.
In 8174, basically we gathered the info about nondeterministic date conversions into a new article, and we link to that article from the articles that you listed. In the articles you listed, we also remove some of the otherwise redundant info about nondeterministic date conversions. The new article is:
.
docs/t-sql/data-types/nondeterministic-convert-date-literals.md
.
https://docs.microsoft.com/sql/t-sql/data-types/nondeterministic-convert-date-literals
.
Thank you very much for bringing this concern to our attention. Although we are Closing this PR 1279, 1279 did cause us to implement the improvement, just in a different way.
GeneMi
.
(htt ps://github . com/ MicrosoftDocs / sql-docs-pr / pull/8174)
https://github.com/MicrosoftDocs/sql-docs-pr/pull/8174
.
CC: @craigg-msft

jzabroski:
Sigh. I think this is not a good solution. People do not read documentation
carefully as-is. I wish I never brought this to your attention I'd this is
the solution.

Hypermedia is not like writing a normal book. We need to think differently
to build computer library dream machines.

PR #1287, opened 2018-11-16
CHAR / VARCHAR typo

camillagaardsted:
A CHAR(10) takes up 10 bytes but a VARCHAR(10) takes 12 bytes.


PRMerger9:
@camillagaardsted : Thanks for your contribution! The author, @MikeRayMSFT, has been notified to review your proposed change.

MikeRayMSFT:
@pmasl - would you mind verifying this submission is technically correct?

Thank you.

pmasl:
It is not entirelly, change to below instead:

This reduction is because `NVARCHAR(10)` requires up to 22 bytes for storage, whereas `VARCHAR(10)` requires up to 12 bytes for the same Unicode string. 

Or keep the original sentence, and simply be change "x2" to "x0" bytes, which was the original intent:

This reduction is because `NCHAR(10)` requires 20 bytes for storage, whereas `CHAR(10)` requires 10 bytes for the same Unicode string. 


Any of those are good and carry the same message.

MikeRayMSFT:
The update in this pull request has been moved to the private repository for publishing. 

Thank you @camillagaardsted  for the submission.

PR #1310, opened 2018-11-21
Update string-split-transact-sql.md

gbritton1:
I added a note in the remarks section discussing the order of the result set, which is not currently specified.  I give two options for the new sentence, depending on whether the function returns the results in order or not as indicated by the options in braces.  At the time of writing I do not know which version is correct but it should be stated one way or the other.

### DIFF
--- a/docs/t-sql/functions/string-split-transact-sql.md
+++ b/docs/t-sql/functions/string-split-transact-sql.md
@@ -61,6 +61,8 @@ STRING_SPLIT ( string , separator )
 |amet.|  
   
 If the input string is **NULL**, the **STRING_SPLIT** table-valued function returns an empty table.  
+
+If the input string is not **NULL**, the **STRING_SPLIT** table-valued function returns a result set in {the order of | no predetermined order} with respect to the delimited substrings as found in the input string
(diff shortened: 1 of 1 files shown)


PRMerger10:
@gbritton1 : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MightyPen:
@gbritton1 Hello gbritton1, I understand the lack of precise specification that you detected in the existing article. Thank you for bringing this situation to our attention. The way that this PR 1310 contains a hypothetical change makes it impossible for Microsoft to formally Merge this PR, and instead we must Close it and never Merge it. However,.....
Your PR is a nicely specific way to explain your point. As a result, we have today implemented PR 8229 in the internal primary repo 'MicrosoftDocs/sql-docs-pr', which implements your suggestion (and has other improvements, as long as we were editing the .md file anyway). So your PR 1310 is a success.
.
CC: @MashaMSFT 
.
#hold-off

gbritton1:
May I know how you phrased the section where I made the suggestion? Or, should I just wait for it to be published?

PR #1320, opened 2018-11-22
Update error-procedure-transact-sql.md

drewjcooper:
Removed confusing text relating to nested `CATCH` blocks.

The text referred to the error number, which is dependent on which `CATCH` block is executing, but which is not returned by `ERROR_PROCEDURE`. Also, `ERROR_PROCEDURE` should return the same value in an outer or inner `CATCH` block, so this text is not relevant to documentation of this function.


MightyPen:
@drewjcooper Hi Andrew, Your deletion of the paragraph is correct, and is much appreciated.  Microsoft is gladly accepting this PR 1320.  Thank you.
.
(CC: @MashaMSFT  Masha, I will Merge this PR now.)
.
GeneMi , 2018/11/27 11:20am

PR #1324, opened 2018-11-23
Update the description of post config step 1

Aniruddh25:
mssql user account is created so that it can be used to run SQL Server service not the launchpad service. Hence, the correction. Also, this setup step is needed only if the user hasnt run it when they installed SQL Server.


PRMerger16:
@Aniruddh25 : Thanks for your contribution! The author, @HeidiSteen, has been notified to review your proposed change.

HeidiSteen:
#sign-off   Thank you!

PR #1330, opened 2018-11-25
Update freetexttable-transact-sql.md

pez-public:
The description of the results for example B do not match what the query returns.
Example C builds on example B, so the query should remain the same while the example description gets updated.


PRMerger7:
@pez-public : Thanks for your contribution! The author, @douglaslMS, has been notified to review your proposed change.

douglaslMS:
@pez-public Thank you for catching this issue. Please change `match` to `matches` (since `description` is the singular subject), and then I'll sign off on this PR. Thanks!

PRMerger10:
@pez-public : Thanks for your contribution! The author, @douglaslMS, has been notified to review your proposed change.

PR #1333, opened 2018-11-25
translate-transact-sql.md - Spelling, Grammar, Wording

LarnuUK:
The document was quite poorly worded. It contained spelling errors, and read like broken English. It also contained statements that weren't true; if NULL is passed for any parameter, then NULL is returned, however the document states the opposite. The word "should" in the document implies that something isn't always the case. For example "he behavior of the TRANSLATE function should be identical to the REPLACE function. " If TRANSLATE doesn't always operate like REPLACE, when does it not? It's not that TRANSLATE should, it does. I've also removed the repetition of this statement.

### DIFF
--- a/docs/t-sql/functions/translate-transact-sql.md
+++ b/docs/t-sql/functions/translate-transact-sql.md
@@ -21,7 +21,7 @@ monikerRange: ">=sql-server-2017||=sqlallproducts-allversions||>=sql-server-linu
 # TRANSLATE (Transact-SQL)
 [!INCLUDE[tsql-appliesto-ss2017-xxxx-xxxx-xxx-md](../../includes/tsql-appliesto-ss2017-xxxx-xxxx-xxx-md.md)]
 
-Returns the string provided as a first argument after some characters specified in the second argument are translated into a destination set of characters.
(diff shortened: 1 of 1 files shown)


MightyPen:
@MashaMSFT Hi Masha, Yes I do recommend that Microsoft accept these changes to TRANSLATE.  These changes are nice improvements.  I will Merge this public PR 1333 now.
.
CC: @LarnuUK Thank you Thom (LarnuUK), these are high quality improvements.
.
GeneMi , 2018/11/27 11:14am

MightyPen:
Minor FYI, private repo PR 8232 was also spurred by this public repo PR 1333. PR 8232 made the four-REPLACE code example much more readable to the human eye.

LarnuUK:
Was this supposed to be live today @MightyPen ? I can see the new formatting for `REPLACE`, however, the typoes, poor grammar and incorrect information is still there.

MightyPen:
@LarnuUK No, not yet. We cannot promise much about how long it takes our system to make live changes from Pull Requests such as this.  Perhaps a "public PR" such as this PR 1333 are visible on the live Docs website within 48 hours; but there are several reasons why this or that PR, or on this or that day, a public PR might not be visible live for several days.  Thanks.

PR #1334, opened 2018-11-26
SSMS Tips and Tricks Book

greglow:
Hi Folks, I have a free eBook for people that covers SSMS Tips and Tricks. It's nearly 200 pages and is free. Would be good to list it right here if that works for you. It has more than a hundred tips and tricks, and is updated every few months.

### DIFF
--- a/docs/ssms/tutorials/ssms-tricks.md
+++ b/docs/ssms/tutorials/ssms-tricks.md
@@ -32,6 +32,8 @@ This tutorial gives you some additional tricks for using SQL Server Management S
 > * Access your SQL Server error log
 > * Find the name of your SQL Server instance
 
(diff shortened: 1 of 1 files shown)


PRMerger16:
@greglow : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
Hi @greglow , thank you so much for your contribution! Unfortunately, we aren't typically allowed to promote third party content on docs.microsoft.com. Additionally, it looks like there's a certificate issue with the site: 
This server could not prove that it is ssmsbook.sqldownunder.com; its security certificate is not trusted by your computer's operating system. This may be caused by a misconfiguration or an attacker intercepting your connection.

I super appreciate your input though! Maybe a good medium would be for you to add some of the more popular tips and tricks to this docs page? That way you'll be listed as a contributor, the content will be available to customers, and we won't get in trouble for promoting content. Thanks again!

PR #1341, opened 2018-11-26
Fix docker image not available

xs4free:
Changed `mcr.microsoft.com/mssql/server/mssql-server-linux` to `mcr.microsoft.com/mssql/server` because image is deprecated:
[We are moving to mcr.microsoft.com where you can pull SQL Server 2017 on Linux containers as well as SQL Server 2019 preview containers. SQL Server 2019 preview containers will only be available on mcr.microsoft.com. Overtime we will only publish to mcr.microsoft.com/mssql/server and eventually deprecate microsoft/mssql-server-linux.](https://hub.docker.com/r/microsoft/mssql-server-linux/)


PRMerger16:
@xs4free : Thanks for your contribution! The author, @MikeRayMSFT, has been notified to review your proposed change.

PR #1367, opened 2018-11-30
TRANSLATE vs REPLACE

LarnuUK:
I realised, after making a now accepted change, that actually `TRANSLATE` and multiple `REPLACE` statements are not equivalent. If a character is replaced to one that is also appears later in the translatations character string, the character will not be replaced. For `REPLACE`, however, this is not true. `REPLACE` will replace all instances of that character, including those that were replaced in one of the nested functions. I have therefore added this information in the Remarks section.


PRMerger9:
@LarnuUK : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
Hi @LarnuUK , thank  you so much for your contribution! Thanks for noticing the inaccuracy and providing this nice example, it helps with clarity. Thanks again! #sign-ff

PRMerger9:
@GitHubber17 : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

PR #1369, opened 2018-11-30
Remove example B

gbritton1:
Remove example B since it might not always work.  As per the remarks: "The order is _not_ guaranteed to match the order of the substrings in the input string."


PRMerger7:
@gbritton1 : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
Hi @gbritton1 , thank you so much for your contribution! We appreciate the thoughtfulness you're showing with removing this example; however, since we've already mentioned that the order is not guaranteed, we're going to leave this as an example as it does showcase the t-sql nicely. I'll add in a separate note to the example in a separate PR reminding users that the output order may vary. Thanks again!

PR #1395, opened 2018-12-10
Update sql-server-linux-whats-new.md

YounggunKim:
AFAIK, TLS is transport layer security.


rjagiewich:
Verified editorial correction.

PR #1399, opened 2018-12-10
Clarify definition syntax.

gbritton1:
Clarify differences between standard and memory optimized table types.  The current page has two syntax sections.  The second one is actually for memory optimized tables. This commit clarifies that, fixes some minor spacing issues and adds table indexes to the standard syntax for non-memory optimized table types, along with the comment concerning version support


PRMerger8:

PR #1401, opened 2018-12-11
Fix SQL example with incorrect syntax

alban-auzeill:
Without brackets the character '-' cause the following error: [S0001][102] Incorrect syntax near '-'.


PRMerger6:

PR #1402, opened 2018-12-11
Extra "```sql" inside a "```sql" block

alban-auzeill:
In the final document, the second "```sql" is printed.


PRMerger9:

PR #1404, opened 2018-12-11
Fix syntax error in example and fix queries

alban-auzeill:
A column alias can not be `SYSDATETIME()` but `[SYSDATETIME()]`, otherwise it's an error: [S0001][102] Incorrect syntax near ')'.
Also update queries to match results


PRMerger8:
@alban-auzeill : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
Hi @alban-auzeill , thank you for contribution and thanks for making this code sample better! #sign-off

PR #1419, opened 2018-12-13
Update set-transaction-isolation-level-transact-sql.md

DanielSSilva:
The READ_COMMITTED_SNAPSHOT is set to OFF by default on SQL Server, on Azure Databases it is ON by default


PRMerger17:
@DanielSSilva : Thanks for your contribution! The author, @CarlRabeler, has been notified to review your proposed change.

PR #1420, opened 2018-12-13
Add https flag to Initialize-HgsAttestation cmdlet

charles-dawes:
Without this flag being present when running the command it creates the endpoints on HTTP rather than HTTPS.


PRMerger15:
@charles-dawes : Thanks for your contribution! The author, @jaszymas, has been notified to review your proposed change.

PR #1427, opened 2018-12-15
Dead link

gotqn:
A link was not working because "This page has been disabled."


PRMerger8:
@gotqn : Thanks for your contribution! The author, @MightyPen, has been notified to review your proposed change.

MightyPen:
@gotqn Hello George, Good catch of the broken link.  And a creative idea, of a dynamic search, for fixing the link!
However, for other reasons, rather than fixing this link, in this particular case we feel this link should be removed entirely.
Therefore I am going to Close this public PR 1427, and then I will remove the link entirely.
Thank you very much for bringing this broken link to our attention.

PR #1432, opened 2018-12-15
columns type length change/ change history deleted

gotqn:
We do not need a change history as we have git. Also, in the other extended events objects such are missing.
Column data types length change to match SQL Server 2017 ones.


PRMerger8:
@gotqn : Thanks for your contribution! The author, @stevestein, has been notified to review your proposed change.

stevestein:
@gotqn - thanks for the update.

#sign-off

PR #1434, opened 2018-12-15
Syntax changed

gotqn:
Syntax changed to match other extended events "Relationship Cardinalities" tables


stevestein:
@gotqn - thanks for the update.

#sign-off

PR #1436, opened 2018-12-15
change history deleted

gotqn:
We do not need a change history as we have git. Also, in the other extended events objects such are missing.


PRMerger15:
@gotqn : Thanks for your contribution! The author, @stevestein, has been notified to review your proposed change.

stevestein:
@gotqn - yes, these articles haven't been touched in a while and the change history sections are no longer needed.

Thanks.

#sign-off

PR #1441, opened 2018-12-16
PARSENAME return type correction

LarnuUK:
The documentation advises that the return type is an `nchar` (with no length declared). This isn't true, it returns the `sysname` datatype; which is a synonym for `nvarchar(128)` (which is a different datatype to `nchar` as well).


pmasl:
#sign-off

PRMerger18:
@pmasl: I'm sorry - only the author of this article, @MashaMSFT, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

pmasl:
@MashaMSFT sign-off

PR #1456, opened 2018-12-20
Update sys-dm-db-page-info-transact-sql.md

AaronBertrand:
There is no view called sys.dm_exec_processes. Seems a conflation of sys.sysprocesses and sys.dm_exec_requests.


PRMerger8:
@AaronBertrand : Thanks for your contribution! The author, @, has been notified to review your proposed change.

PR #1489, opened 2019-01-02
Fixed a precendence order explanation

bbetov:
The order of operations was swapped in the explanation sentence.

### DIFF
--- a/docs/t-sql/language-elements/operator-precedence-transact-sql.md
+++ b/docs/t-sql/language-elements/operator-precedence-transact-sql.md
@@ -24,7 +24,7 @@ manager: craigg
 
   When a complex expression has multiple operators, operator precedence determines the sequence in which the operations are performed. The order of execution can significantly affect the resulting value.  
   
- Operators have the precedence levels shown in the following table. An operator on lower levels is evaluated before an operator on a higher level.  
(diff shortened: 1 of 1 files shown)


PRMerger8:
@bbetov : Thanks for your contribution! The author, @douglaslMS, has been notified to review your proposed change.

douglaslMS:
@bbetov Branimir, I think that our use of "higher" and "lower" is somewhat ambiguous, but I believe that the page is correct as written. The higher-level operator _at the same level_ is evaluated first. For example, in the example `SET @MyNumber = 2 * 4 + 5;`, the higher-level multiplication operator is evaluated first. On the other hand, if I enclose an operation in parentheses, then that nested operation is evaluated first - maybe this is what you had in mind?

bbetov:
@douglaslMS Indeed. I was helping a friend that got confused by the wording and I was trying to make it a bit clearer, but re-reading it now, it is still ambiguous. How about clarifying that 1 is highest and 8 is lowest? 

Perhaps:
*An operator on higher levels is evaluated before an operator on a lower level (1 is highest and 8 lowest level).*

bbetov:
@douglaslMS Sure - will do. Thanks.

PRMerger20:
@bbetov : Thanks for your contribution! The author, @douglaslMS, has been notified to review your proposed change.

Jak-MS:
@douglaslMS - if you approve, please sign off and we'll merge this. thanks.

douglaslMS:
#sign-off

PR #1490, opened 2019-01-02
COMPRESS - Correct Archive SQL Syntax

LarnuUK:
The syntax for the `DELETE` statement was vwrong. The `OUTPUT` clause was after the `WHERE`, and the prefix `deleted` was not specified for the columns; thus they wouldn't be recognised columns.  Updated to be correct syntax, as per [OUTPUT Clause (Transact-SQL) - Using OUTPUT with a DELETE statement](https://docs.microsoft.com/en-us/sql/t-sql/queries/output-clause-transact-sql?view=sql-server-2017#b-using-output-with-a-delete-statement).


PRMerger8:
@LarnuUK : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
Hi @LarnuUK , thank you so much for your contribution! This correction to the syntax looks great. #sign-off

PR #1492, opened 2019-01-02
Update unpack/install commands

asztal:
I had 3 issues when installing:
* The unzip command was incorrect as `-d` is needed to specify the extraction directory
* The `sqlpackage` executable file is not marked as executable and so it won't run without a chmod.
* The previous command to update .bashrc wrote a literal ~ into the file. This works for running the executable in the shell, but some tools (e.g. `which(1)`) do not understand it so running `which sqlpackage` didn't work.


PRMerger19:
@asztal : Thanks for your contribution! The author, @pensivebrian, has been notified to review your proposed change.

PR #1502, opened 2019-01-07
change the place of partition_number description

gotqn:
The `partition_number` column comes before the `hobt_id` one.


stevestein:
@gotqn - thanks for updating this.

#sign-off

PR #1505, opened 2019-01-08
Updated sp-helptracertokens-transact-sql.md

lokeshvij:
Corrected the syntax which mistakenly marked parameters @publisher and @publisher_db as optional.


PRMerger17:
@lokeshvij : Thanks for your contribution! The author, @stevestein, has been notified to review your proposed change.

lokeshvij:
Just to update, the parameters (@publisher and @publisher_db) are mandatory when executing the procedure from distributor. If the same procedure is executed from publisher, they are not required. 
On this basis, I am not sure if the update that I have made is correct!

stevestein:
@lokeshvij -  The existing 'note' provides clarification so I think we should leave the syntax block as is. 

Thanks.

PR #1509, opened 2019-01-09
Added explanations on indexes

rajucomp:
The docs don't seem to shine light on what type of indexes are created when we define a primary key constraint while creating a table. Are they clustered or nonclustered ? If yes, how. What is the expected behavior when indexes already exist on a table ? What kind of index is created when we add a primary key constraint on an existing table ? These are the points I have tried to cover. I added this information from what I found on web and my own understanding.
https://stackoverflow.com/questions/1251636/what-do-clustered-and-non-clustered-index-actually-mean#comment26927463_1251652
https://www.itprotoday.com/changing-definition-clustered-index

I would be very much grateful if more information could be provided on these topics and corrected if I am wrong. Have a nice day !


PRMerger7:
@rajucomp : Thanks for your contribution! The author, @MikeRayMSFT, has been notified to review your proposed change.

MikeRayMSFT:
Closing - taking suggestions to include in https://github.com/MicrosoftDocs/sql-docs-pr/pull/9036.

MikeRayMSFT:
@rajucomp & @baleng - thank you for the comment. I've staged these internally for a quick review before we publish.

PR #1511, opened 2019-01-09
Need a way to verify licensing

User-4574:
Customers need a way to verify if a license is installed or not. This is generating support requests. Thank you.


PRMerger15:
@User-4574 : Thanks for your contribution! The author, @rothja, has been notified to review your proposed change.

rothja:
@User-4574 Thanks for the change. Per an internal thread this morning, I already made some changes like this to clarify licensing in this article. It has not published yet, but it will after it has been reviewed. I'll close this request. Sorry for the duplicate work.

PR #1514, opened 2019-01-10
Update alter-external-data-source-transact-sql.md

pituach:
ALTER EXTERNAL DATA SOURCE is done using SET and not using WITH hint. There was a mistake in the document in the template of the query. The examples are OK.


PRMerger6:
@pituach : Thanks for your contribution! The author, @CarlRabeler, has been notified to review your proposed change.

PR #1516, opened 2019-01-10
Setting the real SSMS option value

gotqn:
In SSMS 17.9, the displayed options are `Read Write` and `Read Only`. As we are enabling for the first time, `Read Only` does not work for us, so it should be `Read Write`. I've updated the T-SQL statements, too.


PRMerger19:
@gotqn : Thanks for your contribution! The author, @julieMSFT, has been notified to review your proposed change.

julieMSFT:
@gotqn: Thank you for your contribution! I will make the changes you have proposed and close the pull request.

julieMSFT:
#sign-off

PR #1517, opened 2019-01-10
Fix one typo and add a clarification regarding temporary tables

gbritton1:
When you create or access temporary tables, you can specify a schema name, but it is ignored.  This change adds a note to that effect.


PRMerger12:

PR #1531, opened 2019-01-14
Fix probable typo in FILESTREAM RAID documentation

yaakov-h:
As excellent as stripping may be, I don't think it's what the original author had in mind.


douglaslMS:
@yaakov-h This may be the first PR that made me laugh out loud. Good catch! Probably true that it's the "Most expensive option."

PR #1542, opened 2019-01-16
Update always-encrypted-api-reference-for-the-jdbc-driver.md

mvdobrinin:
There are a couple of broken tables in the **Always Encrypted API Reference for the JDBC Driver** page.

image


PRMerger15:
@mvdobrinin : Thanks for your contribution! The author, @MightyPen, has been notified to review your proposed change.

ktoliver:
Verified formatting fix.

@mvdobrinin Thanks!

MightyPen:
@mvdobrinin As an fyi, we recently implemented tooling that aims to tighten the rules about what is considered to be cleaner markdown source.  We are currently working to find and fix all problems similar to this that you have fixed for us here with your PR 1542: we need to add more blank lines in certain places, such as before or after markdown tables, and around ## headers.  Thanks a lot.  GeneMi
CC: @ktoliver

mvdobrinin:
@MightyPen Using tools to find or fix these issues would be great!

PR #1556, opened 2019-01-18
yum install doesn't do local installs

User-4574:
You can't use 'yum install' to install a local package, that is 'yum localinstall', fixed that line. This is kinda important for folks that don't know Linux.


PRMerger9:
@User-4574 : Thanks for your contribution! The author, @stevestein, has been notified to review your proposed change.

User-4574:
Actually disregard that, I guess it's changed. Nevermind.

PR #1573, opened 2019-01-26
Fix error in sample SQL code

microsmith:
Existing code:

...
  col9 char (50) NOT NULL  

  WITH (memory_optimized = on)  
);

Should be:
...
  col9 char (50) NOT NULL  
   
)  WITH (memory_optimized = on) ;


PRMerger20:
@microsmith : Thanks for your contribution! The author, @CarlRabeler, has been notified to review your proposed change.

PR #1574, opened 2019-01-27
Update create-symmetric-key-transact-sql.md

scrappyCoco:
Incorrect syntax description for "Create symmetric key". Examples contradict the syntax.

### DIFF
--- a/docs/t-sql/statements/create-symmetric-key-transact-sql.md
+++ b/docs/t-sql/statements/create-symmetric-key-transact-sql.md
@@ -39,11 +39,8 @@ manager: craigg
 CREATE SYMMETRIC KEY key_name   
     [ AUTHORIZATION owner_name ]  
     [ FROM PROVIDER provider_name ]  
-    WITH 
-      [
-          <key_options> [ , ... n ]  
-        | ENCRYPTION BY <encrypting_mechanism> [ , ... n ] 
-      ]
+    WITH <key_options> [ , ... n ]
+    ENCRYPTION BY <encrypting_mechanism> [ , ... n ] 
   
 <key_options> ::=  
       KEY_SOURCE = 'pass_phrase'  


PRMerger18:
@scrappyCoco : Thanks for your contribution! The author, @VanMSFT, has been notified to review your proposed change.

VanMSFT:
Hi @scrappyCoco - Thanks for opening this Pull Request! I appreciate the contribution. We're deciding to leave the current syntax as is. The reason is that the syntax ENCRYPTION BY is not mandatory when creating the key with EKM, and hence why there is a "|" to represent "or". 

Thanks!
Van

scrappyCoco:
@VanMSFT - Thanks for reply! This page have the following example:
```sql
CREATE SYMMETRIC KEY JanainaKey09   
WITH ALGORITHM = AES_256  
ENCRYPTION BY CERTIFICATE Shipping04;  
```
Now in syntax description is written:
```
CREATE SYMMETRIC KEY key_name   
    [ AUTHORIZATION owner_name ]  
    [ FROM PROVIDER provider_name ]  
    WITH 
      [
          <key_options> [ , ... n ]  
        | ENCRYPTION BY <encrypting_mechanism> [ , ... n ] 
      ]
```
Transact-SQL Syntax Conventions say that: "| (vertical bar) - separates syntax items enclosed in brackets or braces. You can use **only one** of the items.". In actual documentation as I understand there can be only <key_options> or only <encrypting_mechanism>, but not both as in the example. Where I'm wrong?

VanMSFT:
@scrappyCoco - Thanks for pointing that out! The T-SQL Syntax Conventions in this case is wrong. I'll follow-up to fix that. An easy example that contradicts this is the [CREATE LOGIN](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-login-transact-sql?view=sql-server-2017) statement, where you can use both these options:

```
<windows_options> ::=        
    DEFAULT_DATABASE = database  
    | DEFAULT_LANGUAGE = language
```

scrappyCoco:
@VanMSFT - Thanks! It will be great!

VanMSFT:
Hi @scrappyCoco, I spoke with my team on this. When enclosed in brackets, the items are optional, and the vertical bar stands for "or". You are only allowed to use the item once, and hence why there is language in there to represent that. For instance, you can only use ENCRYPTION BY once in the statement. It does not mean that you can't use both <key_options> and ENCRYPTION BY in a statement. I hope that makes sense.

Thanks!
Van

scrappyCoco:
@VanMSFT, ok. Thanks!

PR #1584, opened 2019-01-28
Suggest to be syntax error

yoichinak:
It seems to be an example of syntax error.


PRMerger8:
@yoichinak : Thanks for your contribution! The author, @mightypen, has been notified to review your proposed change.

MightyPen:
@yoichinak Hello Yoichi, This is a good catch and an import code fix.  I fixed this exact sane problem, in this exact same article, very recently in another PR, in our internal repo 'MicrosoftDocs/sql-docs-pr'.  This fix should appear soon in the sister-Public repo 'sql-docs'.
.
So I must Close this public PR 1584, without Merging it.
In normal cases we would happily Merge a PR that has such a good fix, but this time the timing is by coincidence a couple day too late.   :-)

PR #1610, opened 2019-02-04
Update pdw-certificate-provisioning.md

thomasbragaloni:
the certificate renew process is not clear in the original doc. I've added a warning note to ensure the process is clear


mzaman1:
@thomasbragaloni  thanks for making the change. You can sign off.

PR #1616, opened 2019-02-07
Fix reference to KeyStoreSecret in connection parameters.

nagytech:
Examples provided used `KeyStoreAuthentication` to pass the secret/password, whereas it should be `KeyStoreSecret`


PRMerger9:
@nagytech : Thanks for your contribution! The author, @v-kaywon, has been notified to review your proposed change.

TrishaMc:
@craigg-msft - - Would you review the proposed changes? Thanks.

PR #1620, opened 2019-02-07
Fix code blocks in determining-effective-database-engine-permissions.md

mandrie:
There are two code blocks that are messed up in this article because the ``` was not inside the > note


PRMerger18:
@mandrie : Thanks for your contribution! The author, @VanMSFT, has been notified to review your proposed change.

TrishaMc:
Code fence fix confirmed.

PR #1639, opened 2019-02-13
Update lesson-2-3-modifying-the-flat-file-connection-manager.md

XibrenX:
Make sure that the Sample Flat File Source Data is a Package.Connection otherwise the var in step 7 cannot be found.

This is because newer versions create it as a Project Connection by default instead of Package Connection for older versions.
See also https://stackoverflow.com/questions/15555125/expression-builder-of-connection-manager-not-showing-variables.


PRMerger9:
@XibrenX : Thanks for your contribution! The author, @douglaslMS, has been notified to review your proposed change.

GitHubber17:
@CarlRabeler - Would you review the proposed changes? Thanks

CarlRabeler:
please assign to @craigg-msft

craigg-msft:
Thank you @XibrenX

PR #1643, opened 2019-02-14
Update set-concat-null-yields-null-transact-sql.md

MysticHeroes:
Previously included no reference to the limitations when using filtered indexes or spatial indexes. I have added those but the set-statements-transact-sql.md should also be modified to include these considerations.

Additionally there's no mention that XML data type methods e.g. XML.value() will fail unless this option is set to ON; or that the SELECT statements in Query Notifications require this option set to ON.


PRMerger6:
@MysticHeroes : Thanks for your contribution! The author, @CarlRabeler, has been notified to review your proposed change.

PR #1650, opened 2019-02-15
non-current row versions update

billyswitzer:
Recommend that users query the history table directly when needing non-current data as CONTAINED IN does query both the current and history tables.


PRMerger17:
@billyswitzer : Thanks for your contribution! The author, @CarlRabeler, has been notified to review your proposed change.

PR #1651, opened 2019-02-15
Updating RAW option for SQL 2019 as per VSTS 11802739.

tejasaks:


PRMerger14:
@tejasaks : Thanks for your contribution! The author, @CarlRabeler, has been notified to review your proposed change.

CarlRabeler:
@MikeRayMSFT please review

PRMerger20:
@tejasaks : Thanks for your contribution! The author, @CarlRabeler, has been notified to review your proposed change.

MightyPen:
@CarlRabeler @MikeRayMSFT @tejasaks @craigg-msft 
This public repo PR 1651 should not have been approved. The INCLUDES syntax in its only changed line clearly does not follow the syntax in any other Docs article that uses INCLUDES.
It is not trivial to fix flaws after they leak from this public repo into our private repo where they cause warnings or errors (I will fix this counterpart, private PR 9101).
This PR 1651 is only one among many over the past months that would have benefitted if this system was configured to run a Build to catch errors before they get migrated into the more important private repo. We have all made or failed to detect flaws that later leaked, me included. I am beginning a group-wide effort to raise awareness about the need for true care and scrutiny before creating or approving any public PR.  Private PRs do not have this problem.
Ideally Msft employees should be allowed to make all their PRs in private repos, where test Builds automatically happen.
Thanks a lot.  GeneMi . 2019/02/15  16:18pm  Friday

PR #1654, opened 2019-02-18
Changed "can" to "must"

Noctis-:
as next line says then can't be the same


PRMerger13:
@Noctis- : Thanks for your contribution! The author, @MikeRayMSFT, has been notified to review your proposed change.

alekseybochkov:
@Noctis- This is not correct change that completely changes the meaning. 
`can` is the right word.

pmasl:
Technically the word "must" is correct, because to make a non-unique index key be unique, we add a 4-byte hidden column called the uniqueifier. But if you don’t reference all this, then the common reference is that clustered indexes must have unique keys and they're usually also mapped to a unique primary key, and non-clustered don’t have to be unique because you can create a primary key constraint associated with a non-clustered index.
In this instance, either the you submit a more thorough description, include a link to that explanation already in https://docs.microsoft.com/en-us/sql/relational-databases/sql-server-index-design-guide#Clustered, or the current needs to stay as-is.

MikeRayMSFT:
Closing - because current wording is sufficient.

PR #1663, opened 2019-02-20
Fixed incorrect markdown and typo

leqid:
Incorrect markdown was displaying two asterisks in between words instead of the desired italic effect.

Typo was "You must create separate role definition for item and system tasks." (Another possible way to resolve typo was, "You must create a separate role definition for item and system tasks." but went with "definitions" to harmonize with plural "tasks".)


PRMerger7:
@leqid : Thanks for your contribution! The author, @markingmyname, has been notified to review your proposed change.

ktoliver:
Verified the updates.

@leqid Thanks!

PR #1666, opened 2019-02-20
Update system-requirements-installation-and-driver-files.md

airtank20:
Removed reference to SQL Server 2012 R2 as there is no R2 version of SQL Server 2012


PRMerger16:
@airtank20 : Thanks for your contribution! The author, @MightyPen, has been notified to review your proposed change.

MightyPen:
#sign-off 
I agree with @airtank20 . Confirmation can be seen at the following https address (only 2008 had 'R2').

- [https://docs.microsoft.com/previous-versions/sql/](https://docs.microsoft.com/previous-versions/sql/)

PR #1684, opened 2019-02-25
Fix UNION syntax

lnyng:
- Fix mismatch angle bracket
- Fix missing curly braces for the second mutually exclusive items group
- Remove redundant item group
- Better indentation
The fix is small so not split into multiple commits


PRMerger9:
@lnyng : Thanks for your contribution! The author, @douglaslMS, has been notified to review your proposed change.

PR #1691, opened 2019-02-27
Clarify that select into is not atomic

jadarnel27:
It seems to be a common misconception that `SELECT...INTO` either succeeds or fails as a whole.  Users are confused when the insert part fails (and is rolled back) but the new table still exists.  For example,

- Recent post on Database Administrators Stack Exchange: [Disk space full during insert, what happens?](https://dba.stackexchange.com/questions/230491/disk-space-full-during-insert-what-happens)
- Older post on SQL Server Central: [schema changed after the target table was created. rerun the select into query](https://www.sqlservercentral.com/Forums/Topic1312406-392-2.aspx)

Paul White's answer in that second link says, in part:

> The cause of your occasional problems is that SELECT...INTO is a bit of an unusual statement: it executes in two parts.

I couldn't find this officially documented anywhere, so I thought it would be helpful to add it to the docs page.


PRMerger6:
@jadarnel27 : Thanks for your contribution! The author, @douglaslMS, has been notified to review your proposed change.

GitHubber17:
@CarlRabeler - Would you review the proposed changed, or recommend an alternate reviewer? Thanks

CarlRabeler:
@GitHubber17 please assign to @craigg-msft for assigning a reviewer

craigg-msft:
looks good

PR #1695, opened 2019-03-01
Remove statement about aggregation in XE DMV

jadarnel27:
There are no aggregations or statistics about runtime duration or execution count in this DMV.  Although it would be cool if there were 😄 

It almost sounds like this is describing the `execution_count` and `execution_duration_ms` columns from [`sys.dm_xe_session_targets`](https://docs.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-xe-session-targets-transact-sql?view=sql-server-2017), but I can't be sure.


PRMerger19:
@jadarnel27 : Thanks for your contribution! The author, @stevestein, has been notified to review your proposed change.

stevestein:
@jadarnel27 - Looks like it's had this descriptive line for quite some time. Thanks for updating this! 

#sign-off

jadarnel27:
@stevestein glad to do it, thanks for following up!

PR #1702, opened 2019-03-01
Documenting an edge case in implicit conversions

jzabroski:
...that apparently has never been documented.  Found this "trick" on StackOverflow.

### DIFF
--- a/docs/t-sql/data-types/data-type-conversion-database-engine.md
+++ b/docs/t-sql/data-types/data-type-conversion-database-engine.md
@@ -38,6 +38,24 @@ Data types can be converted either implicitly or explicitly.
 Implicit conversions are not visible to the user. [!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)] automatically converts the data from one data type to another. For example, when a **smallint** is compared to an **int**, the **smallint** is implicitly converted to **int** before the comparison proceeds.
   
 **GETDATE()** implicitly converts to date style 0. **SYSDATETIME()** implicitly converts to date style 21.
+
+When using the **+** operator for string concatenation, **NULL** implicitly converts the data type length to one character length longer than the string it is added to.  The following code demonstrates:
+
+```sql
+CREATE VIEW dbo.Test
+AS
+	SELECT CAST(NULL AS VARCHAR(1)) AS VarCharOfLengthOne,
+	'' + NULL AS VarCharOfLengthTwo,
+	'' + '' + NULL AS VarCharOfLengthThree
+;
+GO
+
+SELECT isc.COLUMN_NAME, isc.CHARACTER_MAXIMUM_LENGTH, isc.CHARACTER_OCTET_LENGTH
+FROM INFORMATION_SCHEMA.COLUMNS isc WHERE isc.TABLE_SCHEMA = 'dbo' AND isc.TABLE_NAME = 'Test'
+GO
+
+DROP VIEW dbo.Test;
+```
   
 Explicit conversions use the CAST or CONVERT functions.
   


PRMerger9:
@jzabroski : Thanks for your contribution! The author, @MikeRayMSFT, has been notified to review your proposed change.

srutzky:
Hi @jzabroski . I do not believe that this is an "edge case". It appears to be expected behavior. The `NULL` is not adding one character to the current size, it is being converted into `VARCHAR` first. And with no value, the minimum size for any variable type is 1. So the operation is adding a `VARCHAR(1)` to a `VARCHAR(1)`. For example:

```sql
IF (OBJECT_ID(N'tempdb..#tmp') IS NOT NULL)
BEGIN
    DROP TABLE #tmp;
END;

SELECT '' + '' AS [a],
       '' + NULL AS [b],
       CONCAT('', '') AS [c],
       CONCAT('', NULL) AS [d] INTO #tmp;

SELECT *
FROM   tempdb.sys.columns sc
WHERE  sc.[object_id] = OBJECT_ID(N'tempdb..#tmp');

-- max_length = 2 for columns "a", "b", and "c"; 1 for column "d"
```

Apparently, the `CONCAT` function handles `NULL`s differently and completely ignores them. I suspect that the regular string concatenation via `+` doesn't have that option because the T-SQL parser probably has to assign a datatype (if the value can be converted into the string type in the first place).

I even tested again, but this time setting `SET CONCAT_NULL_YIELDS_NULL OFF;` first, and the result was the same, at least in terms of the temp table created. The only difference was that column "b" contained a space instead of `NULL`.

jzabroski:
@srutzky 

I think you're either proving my point, or I am misunderstanding what point you are trying to make.  An edge case does not mean it is a bug.  An edge case can be expected behavior, but it's an edge case because the documentation literally makes no mention of this behavior.

I'm saying: The Implicit Type Conversion Engine page does not make any mention of how concatenating nulls to string data works.

Also, `CONCAT` is an intrinsic function and is not an implicit conversion.  It's completely outside the scope of this PR.  While you contributing some interesting trivia to my SQL knowledge, I don't see what it has to do with the issue I raised where it's completely undocumented how the + operator handles NULL + String and String + NULL.  If this is the wrong place to document this, then please point to me where the right place is.

srutzky:
Hi @jzabroski . No, I was not saying that "edge case" meant bug. I was thinking of it in terms of it being infrequent / out of the ordinary. And perhaps concatenating a literal `NULL` is exactly that 😉 .

Categorization aside, I think the main issue I saw was the presentation of the information. Saying that concatenating a `NULL` increased the size by 1, even if accurate in terms of net-effect, seems like it's magic, or that the effect is somewhat of a special case, separate from how concatenation usually works. This is why I was using empty strings (just like in your example code) as they become `VARCHAR(1)`, even though they are 0 characters in length, just like the `NULL`s.

> Also, `CONCAT` is an intrinsic function and is not an implicit conversion. It's completely outside the scope of this PR.

Well, the reason that I included the `CONCAT` function was due to it also being string concatenation, _and_ the `NULL` being passed in is still a literal `NULL` and the parameters can be of any type and will get converted to string within the function, hence implicit conversion, right?


> I'm saying: The Implicit Type Conversion Engine page does not make any mention of how concatenating nulls to string data works.

Ok. So I was curious and tried a quick test. Since a `NULL` by itself usually is interpreted as being of type `INT`, I swapped out the `NULL` for an integer value:

```sql
SELECT '' + 1,
       '2' + 1,
       SQL_VARIANT_PROPERTY('2' + 1, 'basetype');
-- 1    3    int
```

The first column doesn't really show the issue, but the second and third columns do: if `NULL` is of type `INT`, and `VARCHAR` + `NULL` returns `VARCHAR`, then `VARCHAR` + `INT` value should return a `VARCHAR`, right? But, it doesn't. 

So, I then turned on "Include Actual Execution Plan" and ran the following:

```sql
SELECT TOP (1) so.[name] + NULL
FROM  sys.objects so;
```

From the execution plan I went to "Show Execution Plan XML..." and found only the following:

```xml
<ScalarOperator>
    <Const ConstValue="NULL" />
</ScalarOperator>
```

There is no "convert implicit" node, hence no implicit conversion is happening. The `NULL` is simply being parsed as a `VARCHAR` so no conversion is needed.

> I don't see what it has to do with the issue I raised where it's completely undocumented how the + operator handles NULL + String and String + NULL. If this is the wrong place to document this, then please point to me where the right place is.

Since there is no implicit conversion happening, this page should not be updated. The place to mention this behavior would be the "[+ (String Concatenation)](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/string-concatenation-transact-sql)" page (here mention that both empty string and `NULL`  become `VARCHAR(1)`). And, possibly also mention just the `NULL` part on the [SET CONCAT_NULL_YIELDS_NULL](https://docs.microsoft.com/en-us/sql/t-sql/statements/set-concat-null-yields-null-transact-sql) page.

Also, as far as an example goes, the following shows the behavior in the simplest form that I can think of:

```sql
EXEC sys.sp_describe_first_result_set N'
SELECT '''' + '''' AS [a],
       '''' + NULL AS [b];
';
```

Take care, Solomon..

jzabroski:
> There is no "convert implicit" node, hence no implicit conversion is happening. The `NULL` is simply being parsed as a `VARCHAR` so no conversion is needed.

You're making a distinction between _type inference_ (which can be inductively reasoned about at compile time) and run-time _implicit conversion_ (required to guarantee safety and deterministic behavior of equality and inequality operators).  That distinction is valid.  I guess the question then becomes: Is my PR on the implicit conversion page really the right place for this trivia?  On one hand, if the litmus test is "must show up as CONVERT_IMPLICIT in execution plan", then the answer is trivially no.  On the other, there is clearly logic that determines what data type NULL is based on its (transitively) adjacent operands.

On the other hand, taking your example and furthering it, since the default data type of NULL is INT, then there are type inference precedence rules for establishing what type NULL should take on:

```sql
EXEC sys.sp_describe_first_result_set N'
SELECT '''' + '''' AS [a], -- system_type_name == varchar(2)
       '''' + NULL AS [b], -- system_type_name == varchar(2)
	   NULL as [c];        -- system_type_name == int
';
```

I guess the proper thing to do is to look at the ANSI standards and see how they describe this situation?  Oh, the GREAT hobbies I sign myself up for.

srutzky:
> I guess the question then becomes: Is my PR on the implicit conversion page really the right place for this trivia? On one hand, if the litmus test is "must show up as CONVERT_IMPLICIT in execution plan", then the answer is trivially no.

Hi John. Given that "implicit conversions" are a thing that people often hear about and look out for, I feel very strongly that we need to tread _very_ carefully when it comes to how we apply certain terminology. We need to always keep in mind that readers of this info not only have highly varied technical skill levels, but they also have highly varied reading skill levels (especially if English is not their native language). A lot of the documentation on this site is already worded in ways that mislead and misinform, because the documentation is taken by many as gospel and 100% verified. Hence, if something is not specifically an implicit conversion, then it shouldn't be considered "close enough".

I think the "[+ (String Concatenation)](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/string-concatenation-transact-sql)" page should be updated to reflect this very useful and helpful info. And, I think that the following example should be added to that page to make it explicitly clear to readers _exactly_ what is going on and what they should expect in these various scenarios:

```sql
EXEC sys.sp_describe_first_result_set N'
SELECT NULL AS [a],        -- system_type_name == int
       '''' AS [b],        -- system_type_name == varchar(1)
       '''' + '''' AS [c], -- system_type_name == varchar(2)
       '''' + NULL AS [d], -- system_type_name == varchar(2)
       '''' + 1 AS [e];    -- system_type_name == int
';
```
Having those first two fields (i.e. the `NULL` and `''` by themselves) should help people visualize that:

1. `VARCHAR(1)` is the minimum `VARCHAR` size, given that an empty string has 0 characters, and
1. `NULL`  only has meaning in context, and that on its own, it merely defaults to being `INT`, but it can't inherently _be_ an `INT` due to field "e" which shows that if it were an `INT`, then the resulting type of field "d" would likewise be `INT`.

I do not believe that the "[+= (String Concatenation Assignment)](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/string-concatenation-equal-transact-sql)" page needs to be updated with this info because that operator requires a variable which already has a defined size.

jzabroski:
Agreed.  Thanks for the constructive feedback!  I'll close this PR and rework it for the + (string concatenation) page.

PR #1705, opened 2019-03-04
Change Derived Hierarchies order from Code to ID

v-redu:
A customer notice that the Derived Hierarchies is order by ID without Code. So I check the currently code and old SQL version code. It seems this is by design. So I changed this document.


PRMerger14:
@v-redu : Thanks for your contribution! The author, @leolimsft, has been notified to review your proposed change.

craigg-msft:
@LeoLiMSFT its a small change and I will merge.

PR #1707, opened 2019-03-04
Wording "No data loss happens" is confusing

ShahAnup:
Depending on what user interprets "No data loss happens" out come is different. When SYSTEM_VERSIONING = OFF, history table stops capturing any more updates hence it is better to clearly state that.


PRMerger6:
@ShahAnup : Thanks for your contribution! The author, @CarlRabeler, has been notified to review your proposed change.

PR #1712, opened 2019-03-05
Removed "pull subscription" wording from the description

jesinj22:
From my testing it looks like when the publication is not immediate-updating, we need to run sp_refreshsubscriptions after sp_addarticle irrespective of whether subscription is pull or push


jesinj22:
I have this documented in this forum question
https://social.msdn.microsoft.com/Forums/en-US/e39951ec-0421-4dd2-b1b7-5be0f217db76/sprefreshsubscriptions-document-question?forum=ssdsgetstarted

stevestein:
@jesinj22 - thanks for the update. 

#sign-off

PR #1713, opened 2019-03-05
removed "with override" from reconfigure statement

sivey42:
There is no reason to have the "with override" here, and having it may encourage people to use it everywhere, rather than only when it's necessary. Removed it so we aren't sending the wrong message on how to use the RECONFIGURE WITH OVERRIDE statement.


PRMerger13:
@sivey42 : Thanks for your contribution! The author, @MikeRayMSFT, has been notified to review your proposed change.

MikeRayMSFT:
@sivey42  thank you for this submission.

@bluefooted could you provide a quick review? 

Thank you.

bluefooted:
Agreed, this is correct and definitely best practice to avoid using this unless it's absolutely necessary.  Thanks!

bluefooted:
@sign-off

MikeRayMSFT:
#sign-off

PR #1730, opened 2019-03-07
fix: docs/t-sql/xml/guidelines-for-using-xml-data-type-methods.md

nschonni:
- Add code fence languages
- Trim trailing/leading space
- Enclose elements with backticks

### DIFF
--- a/docs/t-sql/xml/guidelines-for-using-xml-data-type-methods.md
+++ b/docs/t-sql/xml/guidelines-for-using-xml-data-type-methods.md
@@ -6,9 +6,9 @@ ms.prod: sql
 ms.reviewer: ""
 ms.technology: t-sql
 ms.topic: "language-reference"
-dev_langs: 
+dev_langs:
   - "TSQL"
-helpviewer_keywords: 
+helpviewer_keywords:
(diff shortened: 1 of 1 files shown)


PRMerger9:
@nschonni : Thanks for your contribution! The author, @MightyPen, has been notified to review your proposed change.

MightyPen:
@nschonni Hi Nick, I see only one mistake in this change. We do Not want a blank like between the # H1 header and the `---` triple dash line that precedes the # H1.  That is an exception to the general rule - the general rule in markdown favors blank lines to delimit markdown syntax.  LINT tools do not like the blank line between # H1 and the `---` that ends our metadata set.

## H2 lines Do like one blank line before, and another one after themselves.  Only # H1 is odd.

I will approve this public PR 1730, despite this flaw. But going forward, let's not add the blank line, and feel free to remove the blank line if you see it.

nschonni:
OK, I'll skip adding the blank between the H1 and frontmatter block. Is it your internal linter that picks this up, I haven't seen tools like Markdown lint flag this. They just usually care if there is content between H1 and the frontmatter

PR #1745, opened 2019-03-08
Fix showplanxml.xsd drive path for SQL Server 2017

jzabroski:


PRMerger15:
@jzabroski : Thanks for your contribution! The author, @MightyPen, has been notified to review your proposed change.

MightyPen:
#hold-off 

This public PR 1745 does identify a flaw in this Docs article. Thank you to @jzabroski for bringing this flaw to our attention. However, the understandable solution offered by PR 1745 is not quite right.

Instead of changing hardcoded '130' to '140', we need text that describes the relationship between `SELECT @@VERSION`  and the value needed in the path.

_I (MightyPen) will implement this fix in a private PR. And I will @ notify jzabroski when the PR is merged._

I am now permanently **Closing** this public PR 1745.

jzabroski:
Makes perfect sense to me.  I just wasn't going to be a pain in your butt
about that!  Thanks for moving the cheese.

MightyPen:
In private repo, PR 9366 has now given the more elaborate fix I wanted.

PR #1755, opened 2019-03-10
fixing links which hardcoded pointed to the GitHug page instead of ri…

pituach:
…ghtplacein the document

The links in this page which should point to anchor are pointing to the GitHub Page. They should be relative path and not as bellow example::
https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/functions/columns-updated-transact-sql.md#a-using-columns_updated-to-test-the-first-eight-columns-of-a-table

This link take the readers from the Doc to this project instead of redirect him to the right place in the same doc page. The right link should be relative link like bellow:
#a-using-columns_updated-to-test-the-first-eight-columns-of-a-table


PRMerger13:
@pituach : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
Hi @pituach , thank  you for your contribution! You are correct, these links are broken, and we appreciate you helping fix them :) #sign-off

PR #1758, opened 2019-03-11
Added statement terminators to samples

greglow:
Noticed that only one SQL statement in all the samples had a statement terminator. Given we're supposed to use them, I've added them to all the sample statements here.


PRMerger14:
@greglow : Thanks for your contribution! The author, @jaszymas, has been notified to review your proposed change.

jaszymas:
We have added terminators in a separate PR.

jaszymas:
#please-close

PR #1762, opened 2019-03-11
Added extra final step

greglow:
The tutorial gets you to create two windows, One has AE enabled, one doesn't, But you only ever try the code in the one with it enabled. You need to also try it in the other window to see what happens, or it misses a key point of the tutorial.


PRMerger6:
@greglow : Thanks for your contribution! The author, @jaszymas, has been notified to review your proposed change.

jaszymas:
#sign-off

PR #1770, opened 2019-03-12
Fix syntax guide

pkral:
Parameter name is environment_scope, not target_environment. Also specified the default value of the validate_type parameter.


PRMerger18:
@pkral : Thanks for your contribution! The author, @douglaslMS, has been notified to review your proposed change.

craigg-msft:
thank you @pkral

PR #1771, opened 2019-03-12
Update QDS metrics that now apply to on-prem

jadarnel27:
The "*_log_bytes_used" columns in `sys.query_store_runtime_stats` appear to work correctly in SQL Server 2017 on-prem, so I've removed the "applies only to Azure" bit.

You can find a repro of those columns being populated here: https://dba.stackexchange.com/questions/231682/what-is-log-memory-in-query-store-2017


stevestein:
@jadarnel27 - thanks for updating this.

#sign-off

PR #1783, opened 2019-03-14
Change program name

lianetoohey:
Should be "Report Server Configuration Manager" not "Reporting Services Configuration Manager". While the latter is legitimate it will not find the PBI Report Server


PRMerger19:
@lianetoohey : Thanks for your contribution! The author, @markingmyname, has been notified to review your proposed change.

PR #1785, opened 2019-03-14
Need some fix for versions

taroarai:
It is missin SQL Server 2006 SP2, and for SQL Server 2016 SP2 CU1 - CU2, for CU 4 and CU5 reference the same file, so I add CU5.

Moreover, when I download SRO_3.2.2.20000_1033.cab, the file name is SRO_3.2.2.16000_1033.cab. 
Also, the link ID 866039 is not referenced by the setups instead setups are referencing 836819.

You also need to change it on following Docs;
https://docs.microsoft.com/en-us/sql/advanced-analytics/install/sql-ml-component-install-without-internet-access?view=sql-server-2017


PRMerger12:
@taroarai : Thanks for your contribution! The author, @HeidiSteen, has been notified to review your proposed change.

PR #1788, opened 2019-03-15
Replicas must be async for upgrade.

cafealternativo:
Given that an upgrade requires a restart of the sql service, a synchronous replica being upgraded can potentially block the primary for a large amount of time.
The original change in the document states: "Remove automatic failover on all synchronous-commit replicas", which is not enough; my understanding is that a replica can be have automatic failover disabled but still being synchronous.

I have not tested the steps myself, but a synchronous replica restart will certainly block an active primary replica.

I have only changed the general steps, but I believe this document requires an additional review and possibly more changes.

I may be wrong, I suppose. If someone has tested this steps as originally written; please let me know. My email is fesibaja@microsoft.com.

### DIFF
--- a/docs/database-engine/availability-groups/windows/upgrading-always-on-availability-group-replica-instances.md
+++ b/docs/database-engine/availability-groups/windows/upgrading-always-on-availability-group-replica-instances.md
@@ -72,20 +72,28 @@ Observe the following guidelines when performing server upgrades or updates in o
   
  ![AG Upgrade in HADR Scenario](../../../database-engine/availability-groups/windows/media/alwaysonupgrade-ag-hadr.gif "AG Upgrade in HADR Scenario")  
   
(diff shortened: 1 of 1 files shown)


PRMerger12:
@cafealternativo : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
Hi @cafealternativo , thanks for bringing this to our attention! Let me do some research and get back to you. Thanks!

MashaMSFT:
Hi @cafealternativo , thanks for your contribution! Actually, setting the replica to async is not necessary as the secondary becoming unavailable does not pause queries on the primary. Instead of accepting this change, I made another PR to add a note clarifying this point. Please let me know if further clarity is needed here. Thanks!

PR #1825, opened 2019-03-26
Fix "applies to" for sp_wait_for_database_copy_sync

ricec:
`sp_wait_for_database_copy_sync` is only available on Azure SQL DB, but the documentation currently shows support in SQL Server 2016+. This change includes the appropriate file to accurately reflect support for the stored procedure.


PRMerger19:
@ricec : Thanks for your contribution! The author, @stevestein, has been notified to review your proposed change.

stevestein:
@ricec - thanks for making this update.

#sign-off

Jak-MS:
@ricec @stevestein - since we have no staging on the public repo, can you make sure to check that the article calls the include correctly (after the changes are live)?

PR #1829, opened 2019-03-28
Description copied from other events is erroneous

Rob-S:
Although the description matches that of the StartTime for other events, in this case the StartTime is _not_ the start time of the Audit Logout event, as is the usual case for most other events. It it a reference to the StartTime in the Audit Login event, allowing it to be used to match this Audit Logout event with the corresponding Audit Login event.


PRMerger16:
@Rob-S : Thanks for your contribution! The author, @stevestein, has been notified to review your proposed change.

stevestein:
@Rob-S - thanks for updating this.

#sign-off

PR #1832, opened 2019-03-29
Add link to installing-updates-from-the-command-prompt.md

uecasm:
There is otherwise no obvious link between these entirely related topics, and that doesn't show up in the TOC tree of this page, making it exceptionally hard to discover.

Arguably the other command-line topic should also include this link as well.


PRMerger14:
@uecasm : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
hi @uecasm , thank you so much for your contribution! You're right, this definitely helps with discoverability :) #sign-off

PR #1838, opened 2019-03-29
fixed a bad alias that gave wrong [Parameter-Name]

RansomingSql:
The extended event action join was referencing the event package and not the action package, which is different then what you get when you script out the extended event.

### DIFF
--- a/docs/relational-databases/extended-events/selects-and-joins-from-system-views-for-extended-events-in-sql-server.md
+++ b/docs/relational-databases/extended-events/selects-and-joins-from-system-views-for-extended-events-in-sql-server.md
@@ -246,7 +246,7 @@ SELECT
 		s.name              AS [Session-Name],
 		'3_EVENT_ACTION'    AS [Clause-Type],
 
-		e.package + '.' + a.name
+		a.package + '.' + a.name
 		                    AS [Parameter-Name],
 
 		'(Not_Applicable)'  AS [Parameter-Value]


RansomingSql:
When you use SQL Server Management Studio (14.0.17289.0 at least) to script out the default 'system_health' extended event it generates the code that looks like this
`CREATE EVENT SESSION [system_health] ON SERVER 
ADD EVENT sqlclr.clr_allocation_failure(
    ACTION(package0.callstack,sqlserver.session_id)),`

but when you query the dynamic management views using the query provided in the documentation for system_health (like so)
`
    SELECT
		s.name              AS [Session-Name],
		'3_EVENT_ACTION'    AS [Clause-Type],

		e.package + '.' + a.name
		                    AS [Parameter-Name],

		'(Not_Applicable)'  AS [Parameter-Value]
	FROM
		      sys.server_event_sessions         AS s
		JOIN  sys.server_event_session_events   AS e

			ON  e.event_session_id = s.event_session_id

		JOIN  sys.server_event_session_actions  As a

			ON  a.event_session_id = s.event_session_id
			AND a.event_id         = e.event_id
	WHERE
		s.name = 'system_health'
		AND e.name = 'clr_allocation_failure'`

You get the following Parameter-Name(s)
sqlclr.callstack
sqlclr.session_id

which is NOT what is in the definition of 'system_health'

MightyPen:
#sign-off

@RansomingSql Thank you very much for this Fix!

PRMerger10:
@RansomingSql : Thanks for your contribution! The author, @MightyPen, has been notified to review your proposed change.

PR #1840, opened 2019-03-30
Correcting language

davidbaxterbrowne:
The existing language suggests 1) that ERROR_PROCEDURE can only be called in a stored procedure, and 2) that it always returns the name of the current procedure.

### DIFF
--- a/docs/t-sql/functions/error-procedure-transact-sql.md
+++ b/docs/t-sql/functions/error-procedure-transact-sql.md
@@ -43,7 +43,7 @@ ERROR_PROCEDURE ( )
 **nvarchar(128)**  
   
 ## Return Value  
-When called in a stored procedure CATCH block where an error occurs, `ERROR_PROCEDURE` returns the name of that stored procedure.  
+When called in a CATCH block, `ERROR_PROCEDURE` returns the name of the stored procedure in which the error originated.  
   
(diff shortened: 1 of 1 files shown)


PRMerger14:
@davidbaxterbrowne : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

PRMerger17:
@MashaMSFT : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
Hi @davidbaxterbrowne , thank you so much for your contribution! There's actually a subtle issue witht he suggest wording. You can have a CATCH block in an ad-hoc batch (i.e. not in a stored procedure or trigger),  but ERROR_PROCEDURE() only returns a value if the CATCH block is within a stored procedure or trigger. That is why the original sentence was worded as such, and the sentence you've changed isn't clear about this distinction. I've made a slight change to the wording of your new sentence  to clarify this. Other than that though,  your addition is great, thank you! #sign-off

MashaMSFT:
Hi @davidbaxterbrowne , that is actually true, I just tried it :) Thanks for bringing this to our attention, I'm going to change the sentence to: 
When called in a CATCH block, `ERROR_PROCEDURE` returns the name of the stored procedure or trigger in which the error originated.
Thanks!

PR #1865, opened 2019-04-04
Changed "marks" to "euros" in one place

rcabr:
The example uses euros, but there was one sentence that referred to marks.  Changed "marks" to "euros" in that one place.


PRMerger20:
@rcabr : Thanks for your contribution! The author, @minewiskan, has been notified to review your proposed change.

tfosmark:
@rcabr Good catch! Thank you for the fix.

PR #1872, opened 2019-04-05
Update scale-out-support-for-high-availability.md

ZeroRows:
I thought this would be helpful to prevent having to recreate certificates. I wrongly assumed this was the availability group endpoint when I initially configured the master server. It was only when I got to step 7 that I realised that the master service endpoint is a separate computer object with a unique IP address. 

Incidentally, it appears you can just use the availability group endpoint for the master service to get around this. Presumably that reduces some level of HA though, as the endpoint will be abstracted from the clustered service itself.

### DIFF
--- a/docs/integration-services/scale-out/scale-out-support-for-high-availability.md
+++ b/docs/integration-services/scale-out/scale-out-support-for-high-availability.md
@@ -37,7 +37,7 @@ This account must be able to access SSISDB on the secondary node in the Windows 
 
 ### 2.2 Include the DNS host name for the Scale Out Master service in the CNs of the Scale Out Master certificate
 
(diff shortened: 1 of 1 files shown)


PRMerger17:
@ZeroRows : Thanks for your contribution! The author, @haoqian, has been notified to review your proposed change.

HaoQian-MS:
#sign-off

PRMerger14:
@HaoQian-MS: I'm sorry - only the author of this article, @haoqian, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #1894, opened 2019-04-09
Update define-colors-on-a-chart-using-a-palette-report-builder-and-ss…

matepolsps:
…rs.md

Dear MS, this article says "Custom palettes are report-specific." It looks as if you cannot reuse a palette within a report; it looks like it is chart specific rather than report-specific. Please either correct it or give guidance on how to reuse it for other charts within the report. Thank you.


PRMerger6:
@matepolsps : Thanks for your contribution! The author, @markingmyname, has been notified to review your proposed change.

markingmyname:
#sign-off

PR #1931, opened 2019-04-15
Update tutorial-use-azure-blob-storage-service-with-sql-server-2016.md

sql-williamd:
The "Create stored access policy and shared access storage" script only half-references the current Az module commands (v1.7.0), resulting in this demo block failing with incorrect function names. I have refactored to use the new names and they now "work on my machine"


PRMerger9:
@sql-williamd : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
Hi @sql-williamd , thanks for your contribution! We went through the effort to update all of our commands, and looks like we missed this article :) Thanks for catching it for us! #sign-off

PR #1943, opened 2019-04-17
use square brackets for column name

chucklu:
make sure the column name won't be shown with the keyword color

Currently, the column name was shown with keyword color, it's confused.
image

### DIFF
--- a/docs/t-sql/functions/cast-and-convert-transact-sql.md
+++ b/docs/t-sql/functions/cast-and-convert-transact-sql.md
@@ -46,15 +46,16 @@ These functions convert an expression of one data type to another.
 
 **Cast**
 ```sql  
-SELECT 9.5 AS Original, CAST(9.5 AS int) AS int, 
-    CAST(9.5 AS decimal(6,4)) AS decimal;
+SELECT 9.5 AS Original,
+       CAST(9.5 AS INT) AS [int],
+       CAST(9.5 AS DECIMAL(6, 4)) AS [decimal];
 
 ```  
 **Convert**
 ```sql  
-
-SELECT 9.5 AS Original, CONVERT(int, 9.5) AS int, 
-    CONVERT(decimal(6,4), 9.5) AS decimal;
+SELECT 9.5 AS Original,
+       CONVERT(INT, 9.5) AS [int],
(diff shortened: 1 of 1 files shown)


PRMerger17:
@chucklu : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
Hi @chucklu , thank you for your contribution! You're right, this syntax correction provides clarity to the code block. Thanks again! #sign-off

chucklu:
@GitHubber17  @MashaMSFT @Jak-MS 
Anyone knows what kind of sql syntax parser used on https://docs.microsoft.com?
The documentation still show the column name with the keyword color. 
image
\
\
And I find it behaves like the sql parser in github's markdown as following:
```sql
SELECT 9.5 AS Original,
       CAST(9.5 AS INT) AS [int],
       CAST(9.5 AS DECIMAL(6, 4)) AS [decimal];
SELECT 9.5 AS Original,
       CONVERT(INT, 9.5) AS [int],
       CONVERT(DECIMAL(6, 4), 9.5) AS [decimal];
```

What I got in ssms 17 is as following:
image

MashaMSFT:
Hi @chucklu , thanks for bringing this to our attention! Let me do some research and get back to you. Thanks!

DuncanmaMSFT:
I've been looking into this, the code (highlight.js) that we use has a specific pattern for finding all possible keywords. It then checks against a big list of words, but the initial pattern happily includes words in brackets: https://github.com/highlightjs/highlight.js/blob/master/src/languages/sql.js#L21 

If I can change this pattern to this `\w+(?![^[]*])`, then I think that would work better. Tester setup here: https://regexr.com/4cjqr

MikeRayMSFT:
Thanks @DuncanmaMSFT -
This seems to be better for us (T-SQL) . Would it affect other languages?

DuncanmaMSFT:
It would be for all SQL syntaxes, but nothing else. I'd like to test with a bunch more sample code blocks, but otherwise it should be an easy change to make.

chucklu:
Hi @DuncanmaMSFT ,
   Will you create a new pull request in https://github.com/highlightjs/highlight.js to fix the wrong column color?

PR #1950, opened 2019-04-17
Add a note that Microsoft uses the in place upgrade approach on their…

MarkIannucci:
… own Azure IAAS production databases

I took the time to suggest this edit because I'm encountering a lot of resistance to using in place upgrades for our production databases.  The people who are resistant point to this documentation which seems pretty out of date to me in light of the blog post 5 years ago indicating that Microsoft have switched to using the in place migration approach.


PRMerger16:
@MarkIannucci : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
Hi @MarkIannucci , thanks for bringing this to our attention! Let me look into this. Thanks!

PRMerger19:
@MashaMSFT : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
Hi @MarkIannucci , thank you so much for your contribution! I've modified the language a little bit as the wording was outside the scope of core documentation. Additionally, there may be specific requirements where this scenario may not be the best approach, and as such, we want to be careful about inadvertently misleading customers. Thanks again! #sign-off

MarkIannucci:
@MashaMSFT , not quite as aggressive as I hoped, but it is progress never the less.  Thank you for following up on my PR and taking the feedback seriously :)

PR #1977, opened 2019-04-23
Update sql-server-linux-manage-powershell.md

SQLvariant:
Added  H2 section with instructions and an example of "Using the SQL Server PowerShell Provider" to connect to a SQL-on-Linux instance.


PRMerger12:
@SQLvariant : Thanks for your contribution! The author, @rothja, has been notified to review your proposed change.

PR #1978, opened 2019-04-23
Update sql-server-linux-manage-powershell.md

SQLvariant:
Added H2 section with instructions and an example of "Using the SQL Server PowerShell Provider" to connect to a SQL-on-Linux instance.


PRMerger6:
@SQLvariant : Thanks for your contribution! The author, @rothja, has been notified to review your proposed change.

Jak-MS:
Hi @SQLvariant - the size of this submission is more appropriate for the private repo. I checked and you should have access to that. Can you resubmit the commits in this PR to the private repo so we can use the staging and validation automation there? 

go ahead and tag @rothja when you create that new PR.

thanks.
#please-close

PR #1990, opened 2019-04-25
Update datediff-transact-sql.md

brunom:


PRMerger12:
@brunom : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
Hi @brunom, thanks for submitting your content suggestion! Would you mind clarifying the reasoning behind this change? Thanks! @pmasl  FYI

brunom:
Hi Masha, The old version was just wrong. https://docs.microsoft.com/en-us/sql/t-sql/functions/date-and-time-data-types-and-functions-transact-sql documents that the accuracy is 100 nanoseconds.

pmasl:
The statement in the docs is correct as-is. The statements shown under datepart boundaries  all have 1 ys diff. Perhaps @brunom is misreading the doc. It just says the following examples differ in time by one microsecond (.0000001 second). It has no bearing on the data type precision.

brunom:
@pmasl as it stands, the docs are in contradiction. "one microsecond" is not ".0000001 second". Previously in the same sentence it says that dates are "adjacent" which to me means you cannot represent a date in between, so the difference must be the accuracy of the data type.

pmasl:
@brunom you are correct there is one extraneous zero as I counted. In this case might as well change the example, @MashaMSFT please merge the PR.

MashaMSFT:
#sign-off thanks @pmasl

PR #2030, opened 2019-04-29
Clarify effect of transactions on RESEED

ispedals:
Add a clarification note that if an insertion was attempted but rolled back, it is treated as if the table had a row inserted and deleted, and the identify correction method used will be new_reseed_value + current increment.
. . . . . . . . . . . . . .

(MightyPen, 2019/June/03)
Specifically, the proposed new sentence says:
"*new_reseed_value* will not be used as the identity if the insertion of a row into the table was attempted in a rolled back transaction, even if the table contains no rows after the roll back."

'pmasl' suggestion (of Apr 29) might be an iterative improvement over the particular phrasing of the above proposed new sentence.


PRMerger18:
@ispedals : Thanks for your contribution! The author, @pmasl, has been notified to review your proposed change.

pmasl:
The note may seem confusing. Suggest changing to:

|DBCC CHECKIDENT ( *table_name*, RESEED, *new_reseed_value* )|Current identity value is set to the *new_reseed_value*. If no rows have been inserted into the table since the table was created, or if all rows have been removed by using the TRUNCATE TABLE statement, the first row inserted after you run DBCC CHECKIDENT uses *new_reseed_value* as the identity.If rows are present in the table, or if all rows have been removed by using the DELETE statement, the next row inserted uses *new_reseed_value* + the [current increment](../../t-sql/functions/ident-incr-transact-sql.md) value.<br /><br />If a transaction inserts a row and is later rolled back, the next row inserted uses *new_reseed_value* + the [current increment](../../t-sql/functions/ident-incr-transact-sql.md) value as if the row had been deleted.<br /><br /><br /><br /> If the table is not empty, setting the identity value to a number less than the maximum value in the identity column can result in one of the following conditions:<br /><br /> -If a PRIMARY KEY or UNIQUE constraint exists on the identity column, error message 2627 will be generated on later insert operations into the table because the generated identity value will conflict with existing values.<br /><br /> -If a PRIMARY KEY or UNIQUE constraint does not exist, later insert operations will result in duplicate identity values.|

MightyPen:
This public PR 2030 successfully brought to our attention a need for improvement in the DBCC CHECKIDENT documentation.
It was necessary to use another PR to adjust the phrasing of the new sentence (https://github.com/Potapy4/sql-docs/pull/160).
Therefore, this PR 2030 does not need to be Merged now, and indeed would Conflict with the re-phrased update.

I am now Closing this successful PR 2030.
@ispedals Thank you very much for bringing this documentation shortcoming to our attention.

PR #2051, opened 2019-05-03
Handling ErrorColumn = -1

saaabbir:
This bit is necessary when you use Table Lock for a memory optimised (In Memory) destination table


PRMerger16:
@saaabbir : Thanks for your contribution! The author, @janinezhang, has been notified to review your proposed change.

MightyPen:
@saaabbir Hello, Thank you for this documentation fix!  Merging it now.  Should be on live Docs within a day or two.

PR #2084, opened 2019-05-08
numColumn should be set before malloc

misbaharefin:
Current code tries to allocate memory for columnLabels without setting the value of numColumn.
The fix is to move line
   retCode = SQLNumResultCols(hstmt, &numColumn);  
before
   SQLPOINTER* columnLabels = (SQLPOINTER *)malloc( numColumn * sizeof(SQLPOINTER*) );  
   struct DataBinding* columnData = (struct DataBinding*)malloc( numColumn * sizeof(struct DataBinding) );

### DIFF
--- a/docs/odbc/reference/syntax/sqlcolattribute-function.md
+++ b/docs/odbc/reference/syntax/sqlcolattribute-function.md
@@ -206,11 +206,12 @@ struct DataBinding {
 void printStatementResult(SQLHSTMT hstmt) {  
    int bufferSize = 1024, i;  
    SQLRETURN retCode;  
-   SQLSMALLINT numColumn = 0, bufferLenUsed;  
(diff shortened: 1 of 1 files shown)


PRMerger10:
@misbaharefin : Thanks for your contribution! The author, @MightyPen, has been notified to review your proposed change.

MightyPen:
Makes sense.  Cannot malloc with 0 as a multiplication factor.
#sign-off

MightyPen:
Merging now.
# sign off did not apply any kind of ready-to-merge label, so I will have to merge this myself.

MightyPen:
@PRMerger10 Should this PR 2084 have been Assigned to me (in the upper-right corner of this webpage)?  I have self-Assigned this just now.  Without an assignee, perhaps PRs like this could accidentally become forgotten and unnoticed.  Thanks, GeneMi.

PR #2103, opened 2019-05-09
Update sql-linux-quickstart-connect-query.md

iJoris:
localhost does not work, 127.0.0.1 does.


PRMerger6:
@iJoris : Thanks for your contribution! The author, @, has been notified to review your proposed change.

rothja:
@iJoris Thanks for the suggestion! localhost worked in my tests, but maybe it doesn't work in all configurations. Let me look into this briefly. Thank you for your patience!

iJoris:
I'm using Ubuntu 16.04, maybe that helps ;)

rothja:
@iJoris So sorry for the long delay! After looking into this, we think this will normally work to use localhost (I have 16.04 too, so I'm not sure what the difference is). So what we did was to add this to the connection troubleshooting article that we point to. That way people will find this suggestion if they are having issues with localhost. For that, thank you very much for the improvement!! But I'll be closing this specific PR without making the change here. I hope that makes sense. Let me know if not. Thanks!

PR #2110, opened 2019-05-10
syntax highlighting of syntax and code examples

misbaharefin:


PRMerger17:
@misbaharefin : Thanks for your contribution! The author, @MightyPen, has been notified to review your proposed change.

jborsecnik:
@MightyPen, please review. If you want to proceed with this PR, we will move it to the private repo for the sake of validation. Thanks.

MightyPen:
@jborsecnik Hi Jeff, Normally I would agree that a big public PR that touches 100 files would be better if moved to the private repo. However, in this case of PR 2110, there is no need for that extra step.  All the changes herein are nicely limited to the one narrow goal of tagging code blocks with 'cpp'; which is a safe change.

Herein I also see that at least one syntax code block was tagged with 'console'. We do not favor that change (for reasons too involved to explain here), but our concern there is minor, and so we are comfortable with this change as part of this entire desirable PR 2110.

Herein I also see some unwanted space characters deleted in front of a few Alerts.  Those stray from the main goal, but again they are safe improvements.

So, I will # sign off on this PR 2110, as a public PR.
Thanks.

MightyPen:
#sign-off

CC: @jborsecnik @misbaharefin

PR #2116, opened 2019-05-11
Removed the "vertical" and "horizontal" partition description

dbrownems:
Vertical and Horizontal partitioning have an established meaning in database literature that does not match the way the terms were used here.  Whether measure group partitions are based on a multiple separate tables, or multiple queries against a single table does not change the fact that both are "horizontal partitioning".  "Vertical Partitioning" would be where some measures map to one table, and other measures map to a different table, with both tables having matching rows.


PRMerger17:
@dbrownems : Thanks for your contribution! The author, @minewiskan, has been notified to review your proposed change.

Minewiskan:
@dbrownems - Agreed. This article has probably been that way since 2005. Never too late to make things more clear. Thanks for contributing.
#sign-off

PR #2132, opened 2019-05-14
Update Example 2

sqlpup:
was set to return value [1] for both lines, should be [0] to return the first option for "Paris"


PRMerger18:
@sqlpup : Thanks for your contribution! The author, @jovanpop-msft, has been notified to review your proposed change.

rothja:
@sqlpup Thank you for this correction! #sign-off

PR #2169, opened 2019-05-20
Fixed SQL result showing correct timestamp

sagmulders:
Fixed sql ouput based on query above


PRMerger19:
@sagmulders : Thanks for your contribution! The author, @MashaMSFT, has been notified to review your proposed change.

MashaMSFT:
Hi @sagmulders , thanks for contacting Microsoft and for your contribution! You're right, the sample result set is incorrect. Thanks! #sign-off

PR #2179, opened 2019-05-21
Correct note about deployment models

philosophicles:
The edited note was confusing, as it reused the phrase "package deployment model" which already refers to the "Legacy Package Deployment Model". The old wording suggested (implicitly) that the Legacy Package Deployment Model was re-introduced in SQL Server 2016. 

I'm just learning about this but I believe that to be false. Based on text further down (line 57), the truth seems to be that the new Project Deployment Model was extended in 2016, to reintroduce a feature to deploy per-package, but that this wasn't really the same as the Legacy Package Deployment Model, which was and is "gone for good". 

Assuming my interpretation is correct, I've edited the note to match the content of line 57.


PRMerger17:
@philosophicles : Thanks for your contribution! The author, @janinezhang, has been notified to review your proposed change.

MightyPen:
@philosophicles Hello Stuart, Thank you very much for clarifying and fixing the terminology used in this SSIS documentation article.
I am Merging your fix now. The fix should be visible on live Docs within a day or two.

PR #2232, opened 2019-05-30
STRING_AGG Add sentence to Example A

JChrisCompton:
Added a sentence to Example A to clarify that grouping isn't needed when STRING_AGG is the only item in the select. 
I was confused initially when I didn't see a GROUP BY in the first three examples (A, B, and C).  I believe a mention at the end of Example A would have helped me.
Note: The new sentence does not indicate that GROUP BY is needed when there is more than one item in the select list, if all the items in the select list are aggregate functions it is not required. Proof: you do not need a GROUP BY if you change Example A to: SELECT Min(FirstName), Max(FirstName), STRING_AGG (FirstName, CHAR(13)) AS csv FROM Person.Person;


PRMerger12:
@JChrisCompton : Thanks for your contribution! The author, @MikeRayMSFT, has been notified to review your proposed change.

msebolt:
@JChrisCompton Is this not specified in the article's **Arguments** section, `<order_clause>` clause?

JChrisCompton:
@msebolt I see the WITHIN GROUP in the `<order_clause>` which specifies the order of the items in the list.
That does not tell me that I need a GROUP BY clause in my select statement. 
(If I'm wrong let me know; sql is a second language to me.)

This is the experience, which led me to the pull request:
1. I wanted several items in comma delimited lists. 
2. I looked up the function.
3. STRING_AGG is the function name, 1st param is field name, 2nd param separator character.
4. I wrote the SQL I needed. 
The SQL was very complex with my real tables... here is simplified version:
create table #jcc(box char(2),boxType char(1) );
insert into #jcc (box, boxType)VALUES('P6','A'), ('P6','B'), ('P6','C'), ('P5','D'), ('P7','E'), ('P7','F'), ('P7','A'), ('P7','A');select * from #jcc;select box, string_agg(boxType, ',')from #jcc;
drop table #jcc
5. It didn't work... so I looked at examples A and B... they looked like my SQL (no group by clause)
6. I figured out I needed `from #jcc group by (box);`
It took me a little while to figure out that my SQL needed "group by (box)" because the first example (_and_ the next two... which was as far as I looked) didn't have a Group By.  
I concede: in hindsight it is obvious that I need a group by (as with all of the other aggregate functions.)
That's maybe obvious in my contrived example... it was not obvious using my real SQL.

I thought for a while about what would have helped me realize (faster) what my error was, and submitted the pull request.  I think it will help someone. 
If you don't think it would help anyone, or that somehow it would do more harm than good, I respect that decision.

PRMerger10:
@msebolt : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #2240, opened 2019-06-02
Change SQL Server Agent from install to enable

ricoisme:
SQL Server agent is disabled by default since SQL Server 2017 CU4.
I think that  we have to use enable instead of install.

Thanks


PRMerger20:
@ricoisme : Thanks for your contribution! The author, @rothja, has been notified to review your proposed change.

rothja:
@ricoisme Excellent catch and change. Thank you very much! #sign-off

PR #2247, opened 2019-06-03
Correct error in the doc.

mikelawell:
With more data in memory, it is more likely that the columnstore index will find the data it needs in memory with incurring additional reads from disk. 

It should read without incurring additional reads...


PRMerger17:
@mikelawell : Thanks for your contribution! The author, @MikeRayMSFT, has been notified to review your proposed change.

rothja:
@mikelawell Thank you very much for contributing to the docs. This one small typo makes a big change in the meaning, and we appreciate you catching it. I'll sign off on this pull request and it should publish within a few days. Sorry for the delayed response! #sign-off

mikelawell:
Glad I could help.

Mike

PR #2275, opened 2019-06-06
Update catalog-check-schema-version.md

jonburchel:
The article was incorrect.  The parameter is NOT a bool, but requires an INT be used.


PRMerger6:
@jonburchel : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MightyPen:
#sign-off 
I think this change to INT is correct.
@jonburchel Thank you for creating this PR 2275. Your correction/improvement should become visible on live Docs within 24 hours. We hope to hear from you again!

FYI CC: @janinezhang

PR #2286, opened 2019-06-08
Linked to online JDBC connection string builder

databoffin:


PRMerger20:
@databoffin : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MightyPen:
I am looking at this PR 2286.

MightyPen:
@databoffin Hello Phil, I have decided to not accept this PR 2286.

We would need a Description of the reasoning behind this proposed additional link, but the description is empty. If the reason for this added link is that - the 3rd party product the link points to is relevant to the article's particular topic, that is usually not a sufficient reason. Microsoft must remain conservative when it comes to any kind of commercial or advertising use of links in our documentation articles.
Perhaps I might have arrived at a different conclusion had the Description given a reason that I am unaware of for the added link. If I have missed the full reasoning, you may use the @ notification system in this PR 2286 to communicate the complete reasoning to me.

I am Closing this PR 2286 now.
The @ notification system continues to work even in PRs/Issues that are Closed.
Thank you for your interest in Microsoft documentation.

databoffin:
Hi @MightyPen, we created the tool to help people understand the vast number of settings and how they differ for each version (it works 4.1 to 7.0). We've also documented the synonyms in the tooltips as they can be confusing. 

We don't have any advertisement on the page and the download button is for Aireforge Community edition, which is another free tool. So, it's relevant but it's also a good way for people to learn what the various options mean and to quickly get a working connection string that fits their needs.

PR #2298, opened 2019-06-11
Clarified clusterless

allanhirt:
They are read scale (availability) groups, and don't need a cluster, but terminology needs to be consistent


PRMerger20:

rothja:
@allanhirt Thank you. We'll get this change published. #sign-off

PR #2328, opened 2019-06-18
Added  a space between an and on

ManojNair:
Added  a space between an and on


chbain:
Verified minor typo fix.

PR #2357, opened 2019-06-21
Minor PowerShell edits

energizedtech:
CLS Alias replaced with Clear-Host (Suggested practice in scripts to not use aliases), Invoke-Item removed as it does not work (.BCP file extension is not associated with any editor by default.)  Notepad.exe will launch and edit the file if the name is provided as a parameter.


PRMerger13:
@energizedtech : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MashaMSFT:
Hi @energizedtech , these changes look good, thanks for providing them! #sign-off

PR #2379, opened 2019-06-25
Update unicode-transact-sql.md

BrianCMeredith:
the original author meant length, not datalength. As written, the code will return several NULL rows.


PRMerger15:
@BrianCMeredith : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@BrianCMeredith Thanks for this correction. I just tested the original sample and your change, and your correction was a needed fix. We'll get this published!

PR #2401, opened 2019-06-28
Fixed grammar definition, added example

jzabroski:
Hi @stevestein  and @MightyPen , 

I saw this issue today while someone was trying to create an extension for FluentMigrator (project I'm co-maintainer of).  I think it is mostly just a formatting issue, probably as people were porting from the old Microsoft documentation over to GitHub markdown.  I also added an example, since I actually never knew you could add a primary key or index on a user-defined data type.


PRMerger20:
@jzabroski : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #2409, opened 2019-06-30
fix "OVER 句 (Transact-SQL)"→"SELECT - OVER 句 (Transact-SQL)"

hyoshioka0128:
https://docs.microsoft.com/en-us/sql/t-sql/functions/analytic-functions-transact-sql?view=sql-server-2017


hyoshioka0128:
I was misunderstood.
Close #2409

PR #2446, opened 2019-07-07
Correct typo in ENCRYPTION RESUME syntax

rabryst:


PRMerger19:
@rabryst : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@rabryst Thanks for the update. I assume that you replaced a non-standard character with the intended space? It looks identical in the review. We'll get this published.

PR #2448, opened 2019-07-08
fix: MD038/no-space-in-code

nschonni:
Spaces inside code span elements


PRMerger16:
@nschonni : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@nschonni Thank you for submitting these updates, however, when I look at the live files, the current formatting is not incorrect. It is just another way to show code blocks with the single tick. The registry key is debatable, but I think they are escaping the key and the value separately. I'm going to close this pull request without making the change, but please let me know if you have any questions. Thanks!

PR #2506, opened 2019-07-12
Fix sample code external table

jrowlandjones:
Fix external table over HDFS so that it is consistent with previous tutorials. This prevents an error when someone is walking through the tutorials step by step.


PRMerger8:
@jrowlandjones : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #2517, opened 2019-07-15
odd formatting on description of @params argument

mikejwhitty:
the two `*` without a space between them, each to make `parameter_name` and `data_type` emphasised were just displaying as two asterisks on docs.ms


PRMerger8:
@mikejwhitty : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #2518, opened 2019-07-15
remove self-referencing link

mikejwhitty:
the 

> For additional examples, see sp_executesql (Transact-SQL).

link at the bottom of the page goes to: https://msdn.microsoft.com/library/ms188001.aspx

which then permanently redirects to https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-executesql-transact-sql (the page in the documentation we're currently on)

I tried looking on the internet archive wayback machine to see if there was any content to be salvaged, but it doesn't seem to have that page indexed.

### DIFF
--- a/docs/relational-databases/system-stored-procedures/sp-executesql-transact-sql.md
+++ b/docs/relational-databases/system-stored-procedures/sp-executesql-transact-sql.md
@@ -244,8 +244,6 @@ EXECUTE sp_executesql
           @level = 109;  
 ```  
   
- For additional examples, see [sp_executesql (Transact-SQL)](https://msdn.microsoft.com/library/ms188001.aspx).  
-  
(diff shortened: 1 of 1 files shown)


PRMerger9:
@mikejwhitty : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jborsecnik:
@stevestein, just delete the link, or is any replacement available/appropriate? Thanks.

stevestein:
@mikejwhitty - thanks for deleting the link. My guess is that in the past the data warehouse example was in a different article and it pointied back to this one. When the content was added to this article the link was pasted in and not removed.

stevestein:
@jborsecnik - we're ok to just delete the link.

#sign-off

PR #2538, opened 2019-07-17
Clarified earlier version support

greglow:
The new mods were good but it makes it a bit weird when people with earlier versions are reading it. I just think that needs to be clarified.


PRMerger17:

pmasl:
This does not add any useful information. Especially because the paragraph already starts with "Starting with SQL Server 2019". @MikeRayMSFT please close the PR and don't merge.

Issue #2552, opened 2019-07-18
extended_properties storage seems to be valid up to SQL 2017 only


Guillaume-Fourrat:
extended_properties is used as storage of sensitivity metadata, as per current doc, up to SQL 2017.
SQL 2019 (as of CTP3) seems to use sys.sensitivity_classifications instead (not a mere wrapper since extended_properties is empty when sys.sensitivity_classifications is populated. Appears extended_properties exposes raw data from sysxprops when sensitivity_classifications uses sysobjvalues).

(I just had a look to closed entries and I believe this is the cause behind #2209)

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 34863695-cd07-1260-4fea-66738c5fa80a
* Version Independent ID: 4c01015d-9c1d-db8e-84f5-379ffc5d5133
* Content: [SQL Data Discovery & Classification - SQL Server](https://docs.microsoft.com/en-us/sql/relational-databases/security/sql-data-discovery-and-classification?view=sqlallproducts-allversions)
* Content Source: [docs/relational-databases/security/sql-data-discovery-and-classification.md](https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/relational-databases/security/sql-data-discovery-and-classification.md)
* Service: **sql-database**
* GitHub Login: @barmichal
* Microsoft Alias: **mibar**

AndreasWolter:
Yes, this is correct. The extended properties were merely used as a workaround for SQL 2017, because Data Classification was published as a client feature long after SQL 2017 shipped. This way it could still be integrated into the box product, whereas in Azure it was built-in from start

VanMSFT:
Thanks @Guillaume-Fourrat and @AndreasWolter! I'll add a note for SQL Server 2017 for the existing example once the public PR merges.

PR #2555, opened 2019-07-18
Include metadata query for SQL2019

Guillaume-Fourrat:
SQL 2019 CTP3 seems to store classification data differently, adding the TSQL query that allows to retrieve this (extracted the source table info from running the SSMS report under profiler). Might be the reason for #2209 . And is linked to my #2552 .
I suspect this should be better implemented in a version-dynamic fashion that would only display the query relevant to product version but I'm not sure how to do that =)

### DIFF
--- a/docs/relational-databases/security/sql-data-discovery-and-classification.md
+++ b/docs/relational-databases/security/sql-data-discovery-and-classification.md
@@ -135,6 +135,21 @@ FROM
     ON  EP.major_id = C.object_id AND EP.minor_id = C.column_id
 ```
 
+Or on SQL Server 2019 :
+```sql
+SELECT 
+    schema_name(O.schema_id) AS schema_name,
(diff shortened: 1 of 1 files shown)


PRMerger9:

rothja:
Also including @VanMSFT on this pull request. Thanks.

VanMSFT:
Thanks @Guillaume-Fourrat! We appreciate the testing on this. I'll approve this edit, and add a note for the existing example that it applies to SQL Server 2017. Thanks again!

#sign-off

PRMerger6:
@VanMSFT: I'm sorry - only the author of this article, @barmichal, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #2626, opened 2019-07-30
Fixed indents in code

garyericson:
In the code in this article, indenting is used for visual clarity. The problem is that if the user copies the code as-is, Python will give an error about the unexpected indents (R may as well). So I've removed the indents from the embedded scripts.


PRMerger14:
@garyericson : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #2677, opened 2019-08-07
ForEach file loop order specification

BdR76:
The order of files is important when you are processing changefiles with cummulative updates. Even if an explicit order is not supported this is still valuable information to know for developers, so that they can take steps to mitigate this, that's why I added this text here.


PRMerger7:
@BdR76 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MightyPen:
#sign-off 

@PRMerger7 Regardless of Janine Zhang being listed as the Reviewer etc, please Merge this PR 2677.
Thanks.  (GeneMi)

PRMerger12:
@MightyPen: I'm sorry - only the author of this article, @janinezhang, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #2699, opened 2019-08-09
FORMAT - removal of reference to ssCurrent

LarnuUK:
The statement that currently shows in the documentation *"Returns a value formatted with the specified format and optional culture in SQL Server 2017"* is very misleading; it makes it seem like `FORMAT` was added with SQL Server 2017, however, the first version it was made available in is 2012. Stating that the function works in a specific way in 2017, when it hasn't functionally changed since it was added in 2012 doesn't add anything to the article, and in fact, likely only provides confusion.


PRMerger17:

MikeRayMSFT:
@LarnuUK - thank you. 
#sign-off

PR #2701, opened 2019-08-09
Fix syntax

stummsft:
ALGORITHM is mandatory and must be the first option.
ENCRYPTION BY is not exclusive of any key options.


PRMerger12:
@stummsft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stummsft:
Will reopen after I come up with a better break-down of required options.

PR #2718, opened 2019-08-12
Fixed terms

allanhirt:
Always On has a space, and you cannot say things like SQL Always On. It's wrong. It's been fixed in most of the UI as well, and SSCM is fixed in 2019 hence not touching that.


PRMerger9:

rothja:
@allanhirt Thanks for these corrections. I have reviewed them and they do follow standards. I would probably not abbreviate AG for Availability Group, but we can modify that at a later time if needed. Thanks, again. #sign-off

PRMerger12:
@rothja: I'm sorry - only the author of this article, @lrtoyou1223, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #2722, opened 2019-08-13
missing value for azdata bdc create --accept-eula

Guillaume-Fourrat:
as counterintuitive as it seems, it's "azdata bdc create --accept-eula=yes", not just "azdata bdc create --accept-eula"
update both sample and text.


PRMerger14:

PR #2729, opened 2019-08-14
Replace forward slash to back slash

blakhani-msft:
Bash needs back slash as command continuation in next line. Replaced.


PRMerger16:
@blakhani-msft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MikeRayMSFT:
Thank you! @blakhani-msft

PR #2752, opened 2019-08-16
Modify batch file to not delete *.bak

PiJoCoder:
Not sure how to provide feedback on the referenced bootstrap-sample-db.cmd in 
https://raw.githubusercontent.com/Microsoft/sql-server-samples/master/samples/features/sql-big-data-cluster/
There is a line in that file that does this, which removes ALL .bak files in the data folder. This could delete other bak files. So this script should be modified to only delete the one backup file tpcxbb_1gb.bak

echo Removing database backup files...
%DEBUG% kubectl exec %MASTER_POD_NAME% -n %CLUSTER_NAMESPACE% -c mssql-server -i -t -- bash -c "rm -rvf /var/opt/mssql/data/*.bak"

### DIFF
--- a/docs/big-data-cluster/tutorial-load-sample-data.md
+++ b/docs/big-data-cluster/tutorial-load-sample-data.md
@@ -28,7 +28,7 @@ This tutorial explains how to use a script to load sample data into a SQL Server
    - **kubectl**
    - **sqlcmd**
    - **curl**
-
+ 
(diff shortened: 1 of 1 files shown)


PRMerger20:
@PiJoCoder : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #2795, opened 2019-08-23
Added URL to reserved keywords

EnjoyinMoss:
Proposed change


PRMerger8:
@EnjoyinMoss : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@EnjoyinMoss Thanks for this suggestion (I think it might be related to a similar issue filed?). However, when writing T-SQL in SSMS queries, the URL term is not restricted from use like the other reserved keywords. I think by the definition in the article, it is not reserved in the same way as the others. Please let me know if you have any questions, and I apologize for the delay.

PR #2811, opened 2019-08-26
Update docker image name

michellemcone:
`store/microsoft/mssql-server-linux:2017-latest` is no longer used and was left a few places in this file.


rothja:
@michellemcone Thanks for this change. @vin-yu I believe this is correct and will merge. If there are any concerns, you can reach out to me.

PR #2818, opened 2019-08-27
Adding querying sys.tables as an alternative to OBJECTPROPERTY

SvenAelterman:
The value of **large_value_types_out_of_row** table option cannot be queried using OBJECTPROPERTY.

*Also found a space before a comma.*


stevestein:
@SvenAelterman - thanks for the update.

#sign-off

PR #2833, opened 2019-08-29
Update dac-support-for-sql-server-objects-and-versions.md

cairneym:
The description for the validity of the article seems back to front. It suggests that it was valid for 2012 but not for 2014 or later but then provides the links to the earlier versions. I have proposed a change in the wording to make it clear that this doc is for 2014 or later and 2012 and earlier is in the links.

### DIFF
--- a/docs/relational-databases/data-tier-applications/dac-support-for-sql-server-objects-and-versions.md
+++ b/docs/relational-databases/data-tier-applications/dac-support-for-sql-server-objects-and-versions.md
@@ -21,7 +21,7 @@ ms.author: "sstein"
 
 
 > [!IMPORTANT]
-> This article is valid for SQL Server 2012, but not for SQL Server 2014 or later.
+> This article is valid for SQL Server 2014 or later, but not for SQL Server 2012 or earlier.
 > For DAC articles about SQL 2012 and earlier, see the following links:
 > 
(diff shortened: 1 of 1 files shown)


PRMerger20:
@cairneym : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MightyPen:
@cairneym Hello, I fully understand why you might shake your head in dubious amusement when I now reply that - the IMPORTANT note that has you confused is in fact phrased exactly how we intend it to be phrased.  Yes it is an odd case.  Our world is not perfect in every nook and cranny.

We labored over this odd situation on 2018-09-13, almost exactly a year ago. For a combination of reasons, including always limited resources, this article is what it is.

- https://github.com/MicrosoftDocs/sql-docs-pr/pull/7234

So, I am politely rejecting your sensible PR #2833 by permanently closing 2883, without merge.

Thank you for submitting this PR. We hope to hear from you again.

cc: @stevestein Steve, I have attached historical info to PR 7234, if you are curious.

cairneym:
So you mean that the document is NOT valid for SQL Server 2014 and later?

If so then why does it show up when I filter the docs to e.g. SQL Server 2017?   Surely then it should be removed for that version?

I still don’t understand your reply and the target of this article – which versions IS it valid for?   If you can explain it to me then I’m happy for the PR to be closed.

cc: @stevestein<https://github.com/stevestein>

PR #2886, opened 2019-09-04
Replace "Database Login" with "Database User"

pituach:
Replace "Database Login" with "Database User" - this type of mistakes confuses readers to think that a LOGIN is in the database level. I see multiple questions in the MSDN forum which confuse LOGIN and USER which came from reading a blog or document.


PRMerger13:

MashaMSFT:
Hi @hyoshioka0128 , thank you so much for your contribution, and for making docs better! You've actually identified an issue in our docs - these parameters should be in code blocks, and if they are, there is no escape character necessary. I am updating all of these within a bulk update which will supersede this PR. As such, I will proceed with closing out this PR, but please feel free to reply should you need anything further.

Thanks!
Masha from the SQL Docs team

PR #2897, opened 2019-09-05
Fix long dash that should be normal dash

twright-msft:


PRMerger12:
@twright-msft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MikeRayMSFT:
#sign-off

PR #2905, opened 2019-09-05
provide link to sql agent fixed database role docs

MarkIannucci:
so if people wind up on this page seeking the role definitions (like I did), they can quickly go to the correct reference material.


PRMerger6:
@MarkIannucci : Thanks for your contribution! The author(s) have been notified to review your proposed change.

markingmyname:
@MarkIannucci

That information seems to be linked within the same section.  Please review and let me know if this is what you are trying to achieve. 

Members of these database roles can view and execute jobs that they own, and create job steps that run as an existing proxy account. For more information about the specific permissions that are associated with each of these roles, see [SQL Server Agent Fixed Database Roles](https://docs.microsoft.com/en-us/sql/ssms/agent/sql-server-agent-fixed-database-roles?view=sql-server-2017).

MarkIannucci:
Sigh... I don't know why / how I missed the text that you're referencing; it is two paragraphs below my addition!  

Thank you for your patience in reviewing this PR.  Let's abandon this contribution since the info I was after is already there.

PR #2909, opened 2019-09-05
adding missing (bdc) subcommand to azdata and correcting ha file name in patch Command 

rl-msft:


PRMerger18:

PR #2910, opened 2019-09-05
Broken link due to an extra Character "]" in the URL

rl-msft:


PRMerger6:

Jak-MS:
@rl-msft - thanks for the catch!

PR #2917, opened 2019-09-06
Guidance for choice of pwd on RC1

Guillaume-Fourrat:
ADS currently (RC1) works better with "same knox/controller/sa pwds" situation.


PRMerger6:

Guillaume-Fourrat:
As per discussion with UC, we don't want BDC to handle ADS issues, and the change to SA_PWD post setup is confirmed as recommended. 
Better approach will be to create add'l SQL logins to help ADS, will file new PR when full kit is clarified.

PR #2919, opened 2019-09-06
Use object_id instead of hobt_id for matching resource_associated_entity_id

Mario-Hofstaetter:
Following the example I tried to `SELECT * FROM sys.partitions` the object where corrently a lock is on.

Having an active transaction on table `ValuesT0Single`, running this code:

```sql
SELECT resource_associated_entity_id FROM sys.dm_tran_locks  
WHERE resource_database_id = 6 AND request_session_id = 52 and resource_type = 'object'

SELECT object_name(object_id) as objname, object_id, hobt_id
FROM sys.partitions  where hobt_id = 66099276

SELECT object_name(object_id) as objname, object_id, hobt_id
FROM sys.partitions  where object_id = 66099276
```

returns in my test case

```text
resource_associated_entity_id
-----------------------------
66099276

(1 row affected)

objname                 object_id   hobt_id
----------------------- ----------- --------------------

(0 rows affected)

objname                 object_id   hobt_id
----------------------- ----------- --------------------
ValuesT0UInt32          66099276    72057594045267968

(1 row affected)
```

As you can see, `where hobt_id = 66099276` did not return a result, while `where object_id = 66099276` does.
`ValuesT0UInt32` is the correct table name for my case.

So in my humble opinion, it should have been`WHERE object_id=<resource_associated_entity_id>` in the docs.

### DIFF
--- a/docs/relational-databases/system-dynamic-management-views/sys-dm-tran-locks-transact-sql.md
+++ b/docs/relational-databases/system-dynamic-management-views/sys-dm-tran-locks-transact-sql.md
@@ -333,7 +333,7 @@ SELECT resource_type, resource_associated_entity_id,
 ```  
 SELECT object_name(object_id), *  
     FROM sys.partitions  
-    WHERE hobt_id=<resource_associated_entity_id>  
+    WHERE object_id=<resource_associated_entity_id>  
 ```  
   
  The following query will show blocking information.  


PRMerger10:
@Mario-Hofstaetter : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevestein:
@Mario-Hofstaetter - thanks for opening this issue.

When running the examples in this article, I get the correct info with the code as is. When I change to object_id, no info is returned:

image

Mario-Hofstaetter:
The docs example is indeed correct, I did not get a correct result because I was using another setup.

To clearify, my misunterstanding was because:

If the `resource_type` is any of:
- `RID`
- `PAGE`
- `KEY`

then `resource_associated_entity_id` is matching `[sys].[partitions].[hobt_id]`.

If the `resource_type` is of type `OBJECT`
then `resource_associated_entity_id` is matching `[sys].[partitions].[object_id]`.

Thats why I initially didn't get a result.

Complete Example:
```sql
USE [master];
BEGIN TRY DROP DATABASE PR#2919; END TRY BEGIN CATCH END CATCH;
GO

CREATE DATABASE [PR#2919];
GO
USE [PR#2919];

CREATE TABLE MyTestTable  (ID INT PRIMARY KEY);
CREATE TABLE MyTestTable2 (ID INT);
BEGIN TRAN;

INSERT INTO MyTestTable  VALUES (1);
INSERT INTO MyTestTable2 VALUES (1);

SELECT [resource_type], [resource_associated_entity_id], OBJECT_NAME(p.object_id) OBJNAME, p.*
FROM [sys].[dm_tran_locks] tl
LEFT JOIN [sys].[partitions] p ON p.[hobt_id] = tl.[resource_associated_entity_id]
WHERE tl.[resource_associated_entity_id] != 0;

SELECT [resource_type], [resource_associated_entity_id], OBJECT_NAME(p.object_id) OBJNAME ,p.*
FROM [sys].[dm_tran_locks] tl
LEFT JOIN [sys].[partitions] p ON p.[object_id] = tl.[resource_associated_entity_id]
WHERE tl.[resource_associated_entity_id] != 0;

COMMIT TRAN;
```

DfCxwLhMZx

PR #2925, opened 2019-09-06
Elaborate on CHAR function behavior with an eye to UTF-8

### LINKED ISSUE #2922: Suggestion: call input range disparity between CHAR() and NCHAR()
Given that the `NCHAR` function was extended to support non-BMP/supplementary codepoints with adequate storage for UTF-16 surrogate pairs, changing the return type to`NVARCHAR(2)`, one might reasonably expect the `CHAR` function to likewise support UTF-8 collations by returning `VARCHAR(4)`. As this is not the case, please consider updating the "non-European character set" note which references the legacy `Shift_JIS` "double-byte" case to include mention of modern multi-byte UTF-8 and the effective input range 0-127.

Also, please consider adding an example showing how one may safely produce multibyte `VARCHAR(4)` from a given codepoint when the default collation is not a _UTF8 variant. In other words, rather than saying "it won't work", show a workaround.

Assuming effective collation is SQL_Latin1_General_CP1_CI_AS:
```
SELECT CHAR(9834) COLLATE Japanese_CI_AI -- NULL
SELECT CONVERT(CHAR(2), NCHAR(9834)) COLLATE Japanese_CI_AI -- ?
SELECT CONVERT(CHAR(2), NCHAR(9834) COLLATE Japanese_CI_AI) -- ♪

SELECT CHAR(9835) COLLATE Latin1_General_100_CI_AI_UTF8 -- NULL
SELECT CONVERT(VARCHAR(4), NCHAR(9835)) COLLATE Latin1_General_100_CI_AI_UTF8 -- ?
SELECT CONVERT(VARCHAR(4), NCHAR(9835) COLLATE Latin1_General_100_CI_AI_UTF8) -- ♫
```

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b8cf272b-3e28-160e-7e32-cf824766f87a
* Version Independent ID: cb16a55b-c7c3-8e5a-7b1e-9fb8d16cedc2
* Content: [CHAR (Transact-SQL) - SQL Server](https://docs.microsoft.com/en-us/sql/t-sql/functions/char-transact-sql?view=sql-server-2017)
* Content Source: [docs/t-sql/functions/char-transact-sql.md](https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/functions/char-transact-sql.md)
* Product: **sql**
* Technology: **t-sql**
* GitHub Login: @MikeRayMSFT
* Microsoft Alias: **mikeray**

xemed-ahope:
Reframe some of the ASCII and double-byte language to something more
appropriate for the advent of UTF-8 to T-SQL CHAR types.
Add examples to direct users away from this function for those cases
for which it is unsuited.

Corrected pull-request origin account for CLA compliance.

fixes #2922

### DIFF
--- a/docs/t-sql/functions/char-transact-sql.md
+++ b/docs/t-sql/functions/char-transact-sql.md
@@ -200,8 +200,10 @@ NULL                                NULL
   
 ### G. Using CONVERT instead of CHAR to decode multibyte characters
 This example relies on the default codepage of the current database to decode a multibyte character.
-`CONVERT` here operates on a character encoding: SHIFT\_JIS.
-`CHAR` operates on character sets and codepoints: JIS X 208 code 2-86 (row 2 cell 86).
+`CONVERT` here operates on a character encoding, SHIFT\_JIS.
+`CHAR` and `NCHAR` operate on character sets and codepoints,
+which works well with ASCII code 13 above, but not with JIS X 208 code 2-86 here.
+Encoding JIS X 208 ku-ten codes to SHIFT\_JIS is outside the scope of this example.
 
 ```sql
 CREATE DATABASE [multibyte-char-context]
@@ -248,7 +250,7 @@ UTF-8 **char** and UTF-16 **nchar** are different _encoding forms_ using 8-bit a
   FROM enc
 ```
 
-[!INCLUDE[ssResult](../../includes/ssresult-md.md)] Generated under a `\_SC` collation with supplementary character support.
+[!INCLUDE[ssResult](../../includes/ssresult-md.md)] Generated under a `_SC` collation with supplementary character support.
 
 ```
 Music note Music note (UTF-8) Code Point  UTF-16LE bytes UTF-8 bytes
--- a/docs/t-sql/functions/char-transact-sql.md
+++ b/docs/t-sql/functions/char-transact-sql.md
@@ -181,8 +181,8 @@ single_byte_representing_complete_character single_byte_representing_complete_ch
 ### F. Using CHAR to return multibyte characters
 This example uses integer and hex values in the valid range for Extended ASCII.
 However, the `CHAR` function returns `NULL` because the parameter represents only the first byte of a multibyte character.
-A **char(2)** double-byte character cannot be composed from or divided into constituent parts without conversion:
-such parts are not generally valid **char(1)** values.
+A **char(2)** double-byte character cannot be composed from or divided into parts without conversion:
+those parts are not generally valid **char(1)** values.
   
 ```sql
 SELECT CHAR(129) AS first_byte_of_double_byte_character, 
@@ -198,7 +198,7 @@ first_byte_of_double_byte_character first_byte_of_double_byte_character
 NULL                                NULL                                         
 ```
   
-### G. Using CONVERT instead of CHAR to return legacy multibyte characters
+### G. Using CONVERT instead of CHAR to decode multibyte characters
 This example relies on the default codepage of the current database to decode a multibyte character.
 `CONVERT` here operates on a character encoding: SHIFT\_JIS.
 `CHAR` operates on character sets and codepoints: JIS X 208 code 2-86 (row 2 cell 86).
@@ -222,7 +222,7 @@ eighth-note context-dependent-convert context-dependent-cast
 ♪           ♪                         ♪
 ```
 
-### H. Using NCHAR instead of CHAR to return UTF-8 characters
+### H. Using NCHAR instead of CHAR to look up UTF-8 characters
 This example highlights the distinction the Unicode standard makes between a character's _code point_ and the _code unit sequence_ under a given _encoding form_.
 The binary code assigned to a character in a classic character set is its only numeric identifier.
 In contrast, the UTF-8 byte sequence associated with a character is an algorithmic encoding of its assigned numeric identifier: the code point.
@@ -233,7 +233,7 @@ UTF-8 **char** and UTF-16 **nchar** are different _encoding forms_ using 8-bit a
     -- BMP character
     SELECT NCHAR(9835)
     UNION ALL
-    -- non-BMP supplementary character or, under legacy collation, NULL
+    -- non-BMP supplementary character or, under downlevel collation, NULL
     SELECT NCHAR(127925)
   ),
   enc(u16c, u8c) AS (
--- a/docs/t-sql/functions/char-transact-sql.md
+++ b/docs/t-sql/functions/char-transact-sql.md
@@ -177,7 +177,7 @@ single_byte_representing_complete_character single_byte_representing_complete_ch
 ```
 
 ### F. CHAR inability to return multibyte characters
-This example uses the an integer and hex values in the valid range for ASCII. However, the CHAR function returns NULL because the parameter matches the first byte of a number of multibyte characters in the collation, but no single complete character.
+This example uses integer and hex values in the valid range for ASCII. However, the CHAR function returns NULL because the parameter matches the first byte of a number of multibyte characters in the collation, but no single complete character.
   
 ```sql
 SELECT CHAR(129) AS first_byte_of_double_byte_character, 
--- a/docs/t-sql/functions/char-transact-sql.md
+++ b/docs/t-sql/functions/char-transact-sql.md
@@ -224,29 +224,30 @@ UTF-8 `CHAR` and UTF-16 `NCHAR` are different _encoding forms_ using 8-bit and 1
 ```sql
 ; WITH uni(c) AS (
     -- BMP character
-    SELECT NCHAR(0x266B)
+    SELECT NCHAR(9835)
     UNION ALL
-    -- supplementary character, collation-independent construction
-    SELECT CONVERT(NVARCHAR(2), 0x3CD8b5DF)
+    -- non-BMP supplementary character or, under legacy collation, NULL
+    SELECT NCHAR(127925)
   ),
   enc(u16c, u8c) AS (
     SELECT c, CONVERT(VARCHAR(4), c COLLATE Latin1_General_100_CI_AI_SC_UTF8)
     FROM uni
   )
-  SELECT u16c AS [music note]
-    , u8c AS [music note (UTF-8)]
+  SELECT u16c AS [Music note]
+    , u8c AS [Music note (UTF-8)]
+    , UNICODE(u16c) AS [Code Point]
     , CONVERT(VARBINARY(4), u16c) AS [UTF-16LE bytes]
     , CONVERT(VARBINARY(4), u8c)  AS [UTF-8 bytes]
   FROM enc
 ```
 
-[!INCLUDE[ssResult](../../includes/ssresult-md.md)]
+[!INCLUDE[ssResult](../../includes/ssresult-md.md)] Generated under a `_SC` collation with supplementary character support.
 
 ```
-music note music note (UTF-8) UTF-16LE bytes UTF-8 bytes
----------- ------------------ -------------- -----------
-♫          ♫                  0x6B26         0xE299AB
-🎵         🎵                 0x3CD8B5DF     0xF09F8EB5
+Music note Music note (UTF-8) Code Point  UTF-16LE bytes UTF-8 bytes
+---------- ------------------ ----------- -------------- -----------
+♫          ♫                  9835        0x6B26         0xE299AB
+🎵         🎵                 127925      0x3CD8B5DF     0xF09F8EB5
 ```
 
 ## See also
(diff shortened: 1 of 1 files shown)


PRMerger8:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger13:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger19:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger18:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger19:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger6:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MikeRayMSFT:
@xemed-ahope - thanks for the detailed pull request. We're reviewing this.

PRMerger12:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger12:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

xemed-ahope:
> @xemed-ahope - thanks for the detailed pull request. We're reviewing this.

I hope pushing another change wasn't too disruptive. When I took another look at that last example, I realized that it just didn't make sense. The whole point was to show how to look up a character by its code point before encoding it, and there I was decoding UTF-16 bytes instead.

xemed-ahope:
For what it's worth, I don't like the weasel-words around collation ASCII compatibility. I'm almost positive that no collation, of SQL or Windows flavor, fails to include ASCII intact. Even EBCDIC and SHIFT_JIS collations are implemented with ASCII-compatible encodings.

pmasl:
Not sure I get what "weasel-words" mean but I'll be reviewing this ASAP.

xemed-ahope:
@pmasl Sorry, by "weasel-words" I mean noncommittal, deliberately vague.

> Many common character sets share ASCII as a sub-set and will return the same character for integer values in the range 0 through 127

This wording implies that users need to guard against deviation in the ASCII range. I _think_ that it's safe to assume strict ASCII compatibility, but it's hard to prove a negative. From what I can tell, the only relevant character sets/encodings that deviate from ASCII are EBCDIC and the official SHIFT_JIS standard. Support for the former is limited to comparison/ordering; support for the latter is limited to an ASCII-compatible variant, Windows code page 932.

I did end up making that change and others (see 05689cc1b3b7e4d613059010a358bcf86862c336). I assume I shouldn't keep tweaking the pull request directly if someone's actually looking at it. Right?

docs/t-sql/functions/char-transact-sql.md
```diff
@@ -176,8 +176,8 @@ single_byte_representing_complete_character single_byte_representing_complete_ch
 ｼ                                           ｼ                                         
 ```
 
-### F. Using CHAR to return multibyte characters  
-This example uses the an integer and hex values in the valid range for ASCII. However, the CHAR function returns NULL because the parameter represents only the first byte of a multibyte character.
+### F. CHAR inability to return multibyte characters
```
pmasl: Do not change the title of the section.

docs/t-sql/functions/char-transact-sql.md
```diff
@@ -43,10 +43,10 @@ CHAR ( integer_expression )
   
 ## Arguments  
 *integer_expression*  
-An integer from 0 through 255. `CHAR` returns a `NULL` value for integer expressions outside this range, or when then integer expresses only the first byte of a double-byte character.
+An integer from 0 through 255. `CHAR` returns a `NULL` value for integer expressions outside this range, or when no one character matches exactly the integer by both codepoint value and encoded value. Many common character sets share ASCII as a sub-set and will return the same character for integer values in the range 0 through 127.
```
pmasl: While correct, it's a more convoluted explanation that IMO is less clear to the average user. Please keep the "when then integer expresses only the first byte of a double-byte character." but also keep your edit "Many common character sets share ASCII as a sub-set and will return the same character for integer values in the range 0 through 127."

docs/t-sql/functions/char-transact-sql.md
```diff
@@ -43,10 +43,10 @@ CHAR ( integer_expression )
...
 
 > [!NOTE]
-> Some non-European character sets, such as [Shift Japanese Industrial Standards](https://www.wikipedia.org/wiki/Shift_JIS), include characters than can be represented in a single-byte coding scheme, but require multibyte encoding. For more information on character sets, refer to [Single-Byte and Multibyte Character Sets](/cpp/c-runtime-library/single-byte-and-multibyte-character-sets). 
+> Some character sets, such as [Unicode](https://en.wikipedia.org/wiki/Unicode#Mapping_and_encodings) and [Shift Japanese Industrial Standards](https://www.wikipedia.org/wiki/Shift_JIS), have multibyte encodings than can represent some characters in a single-byte, but require as many as four for others. For more information on character sets, refer to [Single-Byte and Multibyte Character Sets](/cpp/c-runtime-library/single-byte-and-multibyte-character-sets).
```
pmasl: Keep "include characters than can be represented in a single-byte coding scheme, but require multibyte encoding."

docs/t-sql/functions/char-transact-sql.md
```diff
@@ -176,8 +176,8 @@ single_byte_representing_complete_character single_byte_representing_complete_ch
 ｼ                                           ｼ                                         
 ```
 
-### F. Using CHAR to return multibyte characters  
-This example uses the an integer and hex values in the valid range for ASCII. However, the CHAR function returns NULL because the parameter represents only the first byte of a multibyte character.
+### F. CHAR inability to return multibyte characters
+This example uses integer and hex values in the valid range for ASCII. However, the CHAR function returns NULL because the parameter matches the first byte of a number of multibyte characters in the collation, but no single complete character.
```
pmasl: Keep original form.

Jak-MS:
@xemed-ahope @pmasl @MikeRayMSFT - after these changes are approved, they'll need to be moved to the private repo for validation and staging. If need be, we can help with that when ready. Thanks.

docs/t-sql/functions/char-transact-sql.md
xemed-ahope: I agree with your objection, but I still think that "first byte of a double-byte" character language is both misleading and incorrect. You'd get NULL when asking for the first, second, third, or fourth bytes of a multibyte UTF-8 character. How about "or not associated with a complete character compatible with the return type"?

PRMerger6:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

docs/t-sql/functions/char-transact-sql.md
xemed-ahope: I reverted to the original title and sentences, but I've added a line to address character composition. I assume that that's the user intent that drives this example.

docs/t-sql/functions/char-transact-sql.md
xemed-ahope: Done.

PRMerger18:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger20:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

docs/t-sql/functions/char-transact-sql.md
xemed-ahope: This still bothers me. Wouldn't it make sense to draw a distinction here between the abstraction level here? If `CHAR()` takes the ordinal/codepoint, then passing in a component from the encoded form reflects a misunderstanding, not a technical limitation. Is the misunderstanding mine? Is `CHAR()` supposed to accept an _encoded byte_ value? If so, then it's not just a disparity relative to `NCHAR()` but a semantic deviation.

rothja:
@xemed-ahope Thanks for making the changes. @pmasl could you please take a look at the File Changes tab and see the latest changes for approval. It looks like we can't merge this in as-is without going through the private repo. But once you approve the changes, we can start that. Thanks!

docs/t-sql/functions/char-transact-sql.md
```diff
@@ -176,8 +177,10 @@ single_byte_representing_complete_character single_byte_representing_complete_ch
 ｼ                                           ｼ                                         
 ```
 
-### F. Using CHAR to return multibyte characters  
-This example uses the an integer and hex values in the valid range for ASCII. However, the CHAR function returns NULL because the parameter represents only the first byte of a multibyte character.
+### F. Using CHAR to return multibyte characters
+This example uses integer and hex values in the valid range for Extended ASCII. However, the CHAR function returns NULL because the parameter represents only the first byte of a multibyte character.
+A `CHAR(2)` double-byte character cannot be composed from constituent first and second `CHAR(1)` bytes in the way that
```
pmasl: This is too convoluted and needs to be explained in a way that's less cryptic for the average user. Please optimize for that.

docs/t-sql/functions/char-transact-sql.md
```diff
@@ -193,6 +196,63 @@ first_byte_of_double_byte_character first_byte_of_double_byte_character
 NULL                                NULL                                         
 ```
   
+### G. Using CONVERT instead of CHAR to return legacy multibyte characters
+This example relies on the default codepage of the current database to map a two-byte legacy codepoint to a single character.
```
pmasl: what's a "legacy codepoint"?

PRMerger7:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

docs/t-sql/functions/char-transact-sql.md
xemed-ahope: I've updated that with a slightly broader, less wordy assertion.

docs/t-sql/functions/char-transact-sql.md
xemed-ahope: Thanks - that was unnecessarily pejorative. My assumption is that CONVERT isn't suitable for UTF-8 lookups, since NCHAR() is an option. It's the only option for SHIFT_JIS and company. I've changed the language to "decode" and "look up" to express a more subtle bias.

PRMerger12:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

docs/t-sql/functions/char-transact-sql.md
```diff
@@ -176,8 +178,11 @@ single_byte_representing_complete_character single_byte_representing_complete_ch
 ｼ                                           ｼ                                         
 ```
 
-### F. Using CHAR to return multibyte characters  
-This example uses the an integer and hex values in the valid range for ASCII. However, the CHAR function returns NULL because the parameter represents only the first byte of a multibyte character.
+### F. Using CHAR to return multibyte characters
+This example uses integer and hex values in the valid range for Extended ASCII.
+However, the `CHAR` function returns `NULL` because the parameter represents only the first byte of a multibyte character.
+A **char(2)** double-byte character cannot be composed from or divided into parts without conversion:
+those parts are not generally valid **char(1)** values.
```
pmasl: One final suggestion for clarity. Would you change:

"A **char(2)** double-byte character cannot be composed from or divided into parts without conversion: those parts are not generally valid **char(1)** values. "

To:

"A **CHAR(2)** double-byte character cannot be partially represented nor divided without some conversion operation. The individual bytes of a double-byte character don't generally represent valid **CHAR(1)** values."

Hopefully I've captured your intent properly.

PRMerger18:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger14:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

docs/t-sql/functions/char-transact-sql.md
xemed-ahope: I hope its okay if I snuck in a "neither" to go with your "nor". I've also kept the lowercase on the datatype since that seems to be the dominant convention. Should I change that?

ktoliver:
@pmasl Could you review the latest changes? Thanks.

xemed-ahope:
@pmasl, if you're going to take another look, let me take one more stab at clarifying Example G

PRMerger8:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger10:
@xemed-ahope : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MikeRayMSFT:
Taking request over to internal repository for PM review. https://github.com/MicrosoftDocs/sql-docs-pr/pull/12924

PR #2947, opened 2019-09-10
bdc subcommnad was missing in few places

rl-msft:
bdc subcommnad was missing in few places also correcting URL http://jmespath.org/ , it had extra ']' character


PRMerger20:

MikeRayMSFT:
#sign-off

jborsecnik:
Note: The link at the last change doesn't work for me. Should it? Thanks.

rl-msft:
@jborsecnik  the link should be 

http://jmespath.org/

which is working for me

jborsecnik:
Thanks.

PR #2997, opened 2019-09-17
Fix indentation issue

vladmihalcea:
From the second item onwards, a `&nbsp;` is added when rendering the page. This is because of the extra spaces between the list marker and the text.


PRMerger14:
@vladmihalcea : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Jak-MS:
@vladmihalcea - thanks for the fix.

PR #3038, opened 2019-09-24
Update configure-distributed-availability-groups.md

jagriffin5:
Added necessary steps to add ProbePort and OverrideAddressMatch parameters to client access point in order for solution to work in Azure environment.  Without these steps the solution does not work with Load balancers.  Additions are lines 149 - 165

### DIFF
--- a/docs/database-engine/availability-groups/windows/configure-distributed-availability-groups.md
+++ b/docs/database-engine/availability-groups/windows/configure-distributed-availability-groups.md
@@ -145,7 +145,25 @@ ALTER AVAILABILITY GROUP [ag2]
     ADD LISTENER 'ag2-listener' ( WITH IP ( ('2001:db88:f0:f00f::cf3c'),('2001:4898:e0:f213::4ce2') ) , PORT = 60173);    
 GO  
 ```  
-  
+
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


PRMerger20:
@jagriffin5 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MashaMSFT:
Hi @jagriffin5 , thank you so much for your contribution! Since this article is intended for on-premises, this may not be the best place for this information. However, the information you've added is certainly useful in the SQL VM documentation and so we've created an internal work item ([1606839](https://mseng.visualstudio.com/TechnicalContent/_workitems/edit/1606839)) to add this information  to that doc set. 

Thanks again for making our documentation better! I will close out this PR at this time, but please let us know if we can be of further assistance in the meantime. 

Thanks!
Masha from the SQL Docs team

jagriffin5:
Thank you!  I appreciate you adding it where you think it will provide the most benefit.  I followed this documentation because I didn’t see anything specific for creating a Distributed AG within Azure it the document has several notes and mentions of Azure Load Balancers so I assumed it was for both on-prem and Azure.  It would be nice though if we could add something in this document for a troubleshooting tip that if you get the message below that you know what to do to resolve.

A connection timeout has occurred while attempting to establish a connection to availability replica '%%%%' with id [%%%%]. Either a networking or firewall issue exists, or the endpoint address provided for the replica is not the database mirroring endpoint of the host server instance.

Again, thank you for reviewing and finding the right location.

Thanks,

James “Bert” Griffin

Premier Field Engineer, Data & AI Domain - Americas
Microsoft Services
Griffin.Bert@microsoft.com<mailto:Griffin.Bert@microsoft.com>
1 (215) 527-5269

PR #3091, opened 2019-10-01
Correction to rows returned without ESCAPE

Rob-S:
"If ESCAPE and the escape character aren't specified" then the "!" would _not_ be treated as the escape character and instead would be part of the string being searched for. Therefore, it would return any rows with the string "30!", not "any rows with the string 30."


PRMerger20:

julieMSFT:
@Rob-S Thank you for catching this. I will approve the change.

#sign-off

PR #3109, opened 2019-10-03
Update overview-of-always-on-availability-groups-sql-server.md

shaneis:
I'm assuming that "avt1" is actually "primary replica"?

This is since I haven't seen another mention of "avt1" in the documentation so far.


PRMerger14:
@shaneis : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MashaMSFT:
Hi @shaneis, thank  you so much for your contribution! Yes, you are right, it should be primary replica :) #sign-off

PR #3123, opened 2019-10-05
Update STRING_ESCAPE to consider the SQL compatibility version

mkxml:
The current docs does not state what SQL Server version is the minimum compatible version for use with the STRING_ESCAPE function.

I found this odd because [STRING_SPLIT](https://docs.microsoft.com/en-us/sql/t-sql/functions/string-split-transact-sql?view=sqlallproducts-allversions) function have a statement showing the minimum compatible version.

### DIFF
--- a/docs/t-sql/functions/string-escape-transact-sql.md
+++ b/docs/t-sql/functions/string-escape-transact-sql.md
@@ -23,7 +23,13 @@ monikerRange: "= azuresqldb-current||>= sql-server-2016||>= sql-server-linux-201
 
 [!INCLUDE[tsql-appliesto-ss2016-asdb-xxxx-xxx-md](../../includes/tsql-appliesto-ss2016-asdb-xxxx-xxx-md.md)]
 
-  Escapes special characters in texts and returns text with escaped characters. **STRING_ESCAPE** is a deterministic function.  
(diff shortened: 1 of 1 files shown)


PRMerger16:
@mkxml : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
@MikeRayMSFT - Could you review the proposed changes and indicate whether the PR can be merged? Thanks.

MikeRayMSFT:
@mkxml - thank you for submitting this. We're evaluating.

bluefooted:
Hi @mkxml ! Version is different than compatibility level. Some features in SQL Server are protected by the database compatibility level, generally when they can influence query optimization and therefore query performance. STRING_SPLIT is unusual in that it is gated by db compat, which is why compat level 130 is called out in the article. STRING_ESCAPE is not gated by db compat, you can use it no matter which compatibility level you are at in SQL Server 2016 and later. Here's an article about db compat for more information on this practice: https://aka.ms/dbcompat

PRMerger7:

PRMerger7:

MikeRayMSFT:
#sign-off

PR #3154, opened 2019-10-08
Update determine-which-queries-are-holding-locks.md

mufukuha:
This sample does not work in Azure SQL Database because Azure SQL Database does not support these ones used in this sample:
- sys.server_event_sessions
- CREATE EVENT SESSION FindBlockers ON SERVER 
- ALTER EVENT SESSION FindBlockers ON SERVER  
- sqlserver.lock_acquired and sqlserver.lock_released event
- sys.dm_xe_session_targets

### DIFF
--- a/docs/relational-databases/extended-events/determine-which-queries-are-holding-locks.md
+++ b/docs/relational-databases/extended-events/determine-which-queries-are-holding-locks.md
@@ -20,7 +20,7 @@ monikerRange: "=azuresqldb-current||>=sql-server-2016||=sqlallproducts-allversio
 ---
 # Determine Which Queries Are Holding Locks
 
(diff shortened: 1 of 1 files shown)


PRMerger7:
@mufukuha : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MightyPen:
Fixed instead by private PR #12531.

Thus now permanently Closing public PRs #3154 and #3155, without merging them into master.
Even though these two public PRs are Closed unmerged, they were successful in that their attempt alerted us o the problem, and helped us to implement an alternative fix for the problem.

@mufukuha Thank you for submitting these PRs. We hope to hear from you again.

PR #3181, opened 2019-10-10
Typo "*@password*"→"*\@password*"

hyoshioka0128:
Italic with escape characters

https://docs.microsoft.com/en-us/sql/database-engine/availability-groups/windows/configure-replication-for-always-on-availability-groups-sql-server?view=sql-server-2017


jborsecnik:
Correction confirmed. Thanks.

PR #3209, opened 2019-10-10
Typo "*@project_name*"→"*\@project_name*"

hyoshioka0128:
Italic with escape characters

https://docs.microsoft.com/en-us/sql/integration-services/system-stored-procedures/catalog-deploy-project-ssisdb-database?view=sql-server-2017


jborsecnik:
Correction confirmed. Thanks.

PR #3211, opened 2019-10-10
Typo "*@trigger_op*"→"*\@trigger_op*"

hyoshioka0128:
Italic with escape characters

https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-check-for-sync-trigger-transact-sql?view=sql-server-2017


jborsecnik:
Correction confirmed. Thanks.

PR #3236, opened 2019-10-11
Typo "**@keepfulltextindexfile**"→"**\@keepfulltextindexfile**"

hyoshioka0128:
bold with escape characters

https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-detach-db-transact-sql?view=sql-server-2017


rothja:
@hyoshioka0128 Thank you!

PR #3312, opened 2019-10-15
Fixes Grammar and Clarity

NicoleWilbur:
Fixes grammar and clarity in these three files:
overview-template-annotated-example.md
overview-template.md
tutorial-template-annotated-example.md

Outstanding issue (overview-template-annotated-example.md file): I was unable to determine the meaning of "refers to keeping to multiple secondary"

in 

"In Azure storage, the term *replication* refers to copying files from one location to another. SQL's *database replication* refers to keeping to multiple secondary databases synchronized with a primary database.

### DIFF
--- a/contributor-guide/templates/overview-template-annotated-example.md
+++ b/contributor-guide/templates/overview-template-annotated-example.md
@@ -1,6 +1,6 @@
 ---
 
-title:="Azure SQL Database backups - automatic, geo-redundant | Microsoft Docs" 
+title:="Azure SQL Database backups - automatic, geo-redundant | Microsoft Docs"
 description: "SQL Database automatically creates database backups and uses Azure read-access geo-redundant storage for geo-redundancy."
 services: "sql-database"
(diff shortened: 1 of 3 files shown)


rothja:
@NicoleWilbur Thanks for these suggested improvements. One note is that there seems to be an added grammar-fix file that is empty in this pull request? At least that is the way it looks like on the files changed tab.

We will need to review these to make sure that no technical meaning has changed. @CarlRabeler is out of the office right now, so I am going to look into this. As a side note, I need to investigate how these templates are being used. It is possible that we are no longer using these and might need to remove them. If so, I apologize for your work, but I need to check first. Thank you for your patience.

rothja:
@NicoleWilbur Thank you for the offer. I'm the manager of the SQL content team that manages this repo. I'll mention this to the team as well. For these templates, I'm doing a little investigation first before deciding if these are still being used. Thanks!

PR #3350, opened 2019-10-18
Update sql-server-backup-to-url-best-practices-and-troubleshooting.md

PhilB-MSFT:
The BackupToURL exe will only output the log file (using trace flag 3051) and the Windows Event Log if you are using Page Blobs (i.e. the backup T-SQL has WITH CREDENTIAL). If you are using block blobs then the TF is ignored and no output is written to log file or the Windows Event Log.

I'm propsing this update as the doc currently gives you the impression that using the TF will result in the verbose output in all use cases, which is not true. 

If there is a better way to convey this situation please update the doc to reflect that.


PRMerger9:
@PhilB-MSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger7:

MikeRayMSFT:
#sign-off

PRMerger8:

PR #3354, opened 2019-10-20
Need to remove the SSAS and SSRS reference in SQL 2017 upgrade content

JoshGer:
The section I edited doesn’t make sense anymore with the original verbiage. Since RS 2017 and AS 2017 are separate installs than SQL Server 2017 now, you can’t really upgrade them at the same time through SQL Server setup like you could before. And RS and AS version numbers are completely different than SQL Server. So having them the exact same 14.x version is impossible.

I feel like the content here was copied text from the 2016 doc here - https://docs.microsoft.com/en-us/sql/database-engine/install-windows/supported-version-and-edition-upgrades

... and the SQL 2016 content does make sense to me because in 2016 you couldn’t upgrade or patch an AS or RS instance only if there was also a SQL Server instance on the same server. For example, if both SQL (database engine) and SSRS were both installed on a server and you ran the SP update on it, you HAD to upgrade both SQL and RS to the SP version. You couldn’t do only one of them.

Starting with SQL 2017, the Analysis Services and Reporting Services part of this content should be removed as we've had a few customers and CSS engineers confused by it.

### DIFF
--- a/docs/database-engine/install-windows/supported-version-and-edition-upgrades-2017.md
+++ b/docs/database-engine/install-windows/supported-version-and-edition-upgrades-2017.md
@@ -36,7 +36,7 @@ monikerRange: ">=sql-server-2016||=sqlallproducts-allversions"
   
 ## Unsupported Scenarios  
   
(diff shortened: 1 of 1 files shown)


PRMerger17:
@JoshGer : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MashaMSFT:
Thanks Josh :) #sign-off

PR #3355, opened 2019-10-21
Update configure-distributed-availability-groups.md

BIGabor:
Distrubuted AOG status check query changed.

### DIFF
--- a/docs/database-engine/availability-groups/windows/configure-distributed-availability-groups.md
+++ b/docs/database-engine/availability-groups/windows/configure-distributed-availability-groups.md
@@ -261,15 +261,29 @@ The following Transact-SQL examples demonstrate the detailed steps to fail over 
 1. Wait until the status of the distributed availability group has changed to `SYNCHRONIZED`. Run the following query on the global primary, which is the primary replica of the primary availability group. 
     
       ```sql  
-      SELECT ag.name
-             , drs.database_id
-             , drs.group_id
-             , drs.replica_id
-             , drs.synchronization_state_desc
-             , drs.end_of_log_lsn 
-        FROM sys.dm_hadr_database_replica_states drs,
-        sys.availability_groups ag
-          WHERE drs.group_id = ag.group_id;      
+      SELECT	ag.name, 
+		ar.replica_server_name, 
+		DB_NAME(drs.database_id) AS DatabaseName,
+		ar.availability_mode_desc, 
+		drs.synchronization_state_desc, 
+		drs.end_of_log_lsn,
+		ars.synchronization_health_desc, 
+		ars.connected_state_desc, 
+		drs.secondary_lag_seconds, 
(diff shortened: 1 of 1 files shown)


PRMerger20:
@BIGabor : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MashaMSFT:
Hi @BIGabor , thank you so much for your contribution! Let me confirm that there are no issues with this script - and so far there seem to be none, and then we can approve it :) Thanks!

BIGabor:
> 
> 
> Hi @BIGabor , thank you so much for your contribution! Let me confirm that there are no issues with this script - and so far there seem to be none, and then we can approve it :) Thanks!

Ok. Included more columns compare to original query, those are usable when you are waiting for a sync. Lag seconds, etc.

MashaMSFT:
Hi @BIGabor , thanks again for your submission! I did some research and it looks like the intent of the query during this step is to be super simple - we just want to show that we're ready for the next step. Additionally complexity in the result set may detract from the goal, and/or cause unnecessary confusion. As such, though the query you've provided is really helpful in other scenarios, it would be considered out of scope for this particular step.  However, since you've also submitted this query as a Git Issue also, it is visible to anyone who visits the page, and so they may benefit from it, without adding it to the documentation. 

Having said that, the query you've written would be a useful addition to this doc, under a "explore health of a distributed ag" heading: https://docs.microsoft.com/en-us/sql/database-engine/availability-groups/windows/dynamic-management-views-and-system-catalog-views-always-on-availability-groups?view=sql-server-ver15

Did you want to add it there? Or if not, I can add it there as well :)
Either way, I'm going to proceed with closing out this particular PR, but please feel free to comment if you need anything else!

Thanks again for helping improve our documentation!
Masha from the SQL Docs team

BIGabor:
Hi Masha

ok, I will do some testing on weekend, and yes, that end of log lsn should
be changed to last redone lsn. Plus I have another script for step 4, to
check status on secondary(forwarder) too. Testing on weekend with failover,
and I will share it.
Another problem with this simple script. No dbname, no replicaname, just
ids. But who knows which id belongs to which server. This failover is
timecritic, I don't want to execute  individual lookup scripts. We have
more groups, with more dbs, that's why filtering for distributed is good. I
want to know, which db is in which state, ok, not in step 4. If you have
big delays in step 4, that's a problem. And it would be more simple, if the
same script could be used for both. So I would like to display all the
required info in case of problem. If it's in sync, it's ok, than you need
only two columns. Ag.name, and state. But if you have problems, I need db
name, lsn, lag secs(not in step 4), etc. Plus we have async replicas too in
normal AGs belongs to distributed AG replica. If those are not in sync
state, doesn't matter. The primary and secondary(forwarder) distributed
nodes should be in sync only during this failover.
So feedback on next Monday.

Regards,
Gabor

PR #3366, opened 2019-10-22
Missing paren on PRINT

LowlyDBA:


PRMerger14:
@LowlyDBA : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MightyPen:
@LowlyDBA Hello John, Thank you for bring this concern to our attention. In the future, it is helpful if these public PRs include more text to explain the PR. For example, why do you believe this is a source code bug? The challenge is that - I am told this code sample worked when we tested it a year ago.  On what version of Python did you encounter a problem?  What was the error message?

Please, can you enter a brief follow-up Comment to tell us a those kinds of information.  And in the Comment, @ notify me directly so I will know.
To be clear, I suspect your concern is correct.  I just want a bit more info.

Thank you very much.
MightyPen

LowlyDBA:
@MightyPen Sure thing, no problem.

As of Python 3, [print is a function](https://docs.python.org/3.0/whatsnew/3.0.html#print-is-a-function) so it requires parentheses when used:

image

MightyPen:
I sent email to PyTalk for a point of clarification:
Were parentheses ALLOWED before version 3 pf Python?

MightyPen:
#sign-off

PR #3384, opened 2019-10-24
Update catalog-get-parameter-values-ssisdb-database.md

DavidHiltenbrand:
This statement, "Use NULL to retrieve all project parameters and the parameters from all packages." is incorrect. @package_name cannot be set to NULL when executing this procedure. The procedure starts with explicitly checking and raising an error when any of the required parameters are NULL:

    IF (@folder_name IS NULL OR @project_name IS NULL 
            OR @package_name IS NULL )
    BEGIN
        RAISERROR(27138, 16 , 1) WITH NOWAIT 
        RETURN 1 
    END


PRMerger10:
@DavidHiltenbrand : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #3397, opened 2019-10-27
Update data-ingestion-curl.md

Kuchan08:
**Issue:**  We need to fix this part of documentation or add the sample airline data in Load Sample Data Script, as customer would expect that the mentioned Sample Data file at HDFS Location “hdfs:///airlinedata".

**Issue Description:** 
Documentation refers the Sample Data : "hdfs:///airlinedata" which should be available in HDFS in order to List Data in HDFS using CURL. But it seems that sample data was not being downloaded by “Load Sample Data” script. 

               Reference: https://docs.microsoft.com/en-us/sql/big-data-cluster/data-ingestion-curl?view=sqlallproducts-allversions

curl -i -k -u root:root-password -X GET 'https://<gateway-svc-external IP external address>:30443/gateway/default/webhdfs/v1/airlinedata/?op=liststatus'

**Solution:**
For quick correction, we can update the documentation to point Sample Data as : hdfs:///product_review_data instead of hdfs:///airlinedata. 
Note: Sample Data : hdfs:///product_review_data will be downloaded by the user through prerequisite step Load Sample Data.
               https://docs.microsoft.com/en-us/sql/big-data-cluster/tutorial-load-sample-data?view=sqlallproducts-allversions

### DIFF
--- a/docs/big-data-cluster/data-ingestion-curl.md
+++ b/docs/big-data-cluster/data-ingestion-curl.md
@@ -17,6 +17,10 @@ ms.technology: big-data-cluster
 
 This article explains how to use **curl** to load data into HDFS on [!INCLUDE[big-data-clusters-2019](../includes/ssbigdataclusters-ver15.md)] (preview).
 
+## <a id="prereqs"></a> Prerequisites
+
+- [Load sample data into your big data cluster](tutorial-load-sample-data.md)
+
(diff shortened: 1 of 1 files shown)


PRMerger10:
@Kuchan08 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #3444, opened 2019-10-31
Limitations of external tables need to be pointed out 

simonsabin:
External tables uses a preview feature, and has a wide range of limitations. These need to be called out


PRMerger16:
@simonsabin : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #3464, opened 2019-11-04
adding data exposed episode on HDFS Tiering

jeroenterheerdt:
adding data exposed episode on HDFS Tiering


jeroenterheerdt:
never mind. I followed the wrong process.

PR #3586, opened 2019-11-14
Typo "**SqlAgent@Adventure-Works.com**"→"**SqlAgent\@Adventure-Works.…

hyoshioka0128:
…com**"

Bold with escape characters

https://docs.microsoft.com/en-us/sql/relational-databases/database-mail/configure-database-mail?view=sql-server-ver15

### DIFF
--- a/docs/relational-databases/database-mail/configure-database-mail.md
+++ b/docs/relational-databases/database-mail/configure-database-mail.md
@@ -189,13 +189,13 @@ ms.author: sstein
  View or update the description of the account. The description is optional.  
   
  **E-mail address**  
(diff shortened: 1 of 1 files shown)


PRMerger17:

ktoliver:
@stevestein - Could you review the proposed update?

Escaping the @ character here (which currently appears in the live version), makes the email address "hot," which isn't optimal in this case because it's an example email address.

rothja:
@ktoliver I think this is not optimal but OK as a customer understands it is an example. #sign-off

PRMerger9:

PR #3590, opened 2019-11-14
Typo '**jdoe@xyz.com**'→'**jdoe\@xyz.com**'

hyoshioka0128:
Bold with escape characters

https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-add-operator-transact-sql?view=sql-server-ver15


PRMerger19:

ktoliver:
@stevestein - Could you review the proposed update?

Escaping the @ character here (which currently appears in the live version), makes the email address "hot," which isn't optimal in this case because it's an example email address. What might work best is using inline code: `jdoe` or `jdoe@xyz.com`

rothja:
#sign-off

PRMerger8:

PR #3621, opened 2019-11-18
Correction of column names in dm_os_sys_info

Rob-S:
Also (not changed) - I see a "automatic soft-NUMA disabled" configuration option, but not "SOFTNUMA" or any other "...numa..." options, even with "show advanced options" enabled.

### DIFF
--- a/docs/database-engine/configure-windows/soft-numa-sql-server.md
+++ b/docs/database-engine/configure-windows/soft-numa-sql-server.md
@@ -126,7 +126,7 @@ SET PROCESS AFFINITY CPU=4 TO 7;
   
 -   [sp_configure &#40;Transact-SQL&#41;](../../relational-databases/system-stored-procedures/sp-configure-transact-sql.md): Displays the current value (0 or 1) for SOFTNUMA  
   
--   [sys.dm_os_sys_info &#40;Transact-SQL&#41;](../../relational-databases/system-dynamic-management-views/sys-dm-os-sys-info-transact-sql.md): The *softnuma* and *softnuma_desc* columns displays the current configuration values.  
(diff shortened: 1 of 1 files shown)


PRMerger16:

CarlRabeler:
@Rob-S thanks for your contribution. Regarding your comment "see a "automatic soft-NUMA disabled" configuration option, but not "SOFTNUMA" or any other "...numa..." options, even with "show advanced options" enabled." - could you please elaborate? I inherited responsibility for this topic and will need to find the right product team expert to assist me here, so I need your assistance in clarifying what your question is and how I can improve this article. 
Thanks, Carl

Rob-S:
Directly above the change that was made, for another bullet with "sp_configure" the article mentions "Displays the current value (0 or 1) for SOFTNUMA." SOFTNUMA is the parameter used in the ALTER SERVER CONFIGURATION command, which is either OFF or ON. One might think that the value 0 represents OFF and 1 represents ON, but that would be wrong. Executing "sp_configure" (commands below) with just "numa" shows all of the configuration options containing that string. There is only a configuration option named "automatic soft-NUMA disabled" which, from that name, would use 1 (true) for disabled and 0 (false) for enabled. It's probably worth some kind of clarification, that sp_configure displays a value where 1 means OFF/disabled and 0 means ON/enabled, but I am not 100% certain that the "automatic soft-NUMA disabled" configuration option is exactly the negation of the "SOFTNUMA" SERVER CONFIGURATION option.

```
select SERVERPROPERTY(N'ProductVersion');
execute sp_configure N'show advanced options', 1;
reconfigure;
execute sp_configure N'numa';
```

PR #3662, opened 2019-11-22
Improve the ADAL release note

David-Engel:
Improve the ODBC and OLE DB driver release notes regarding the change to include ADAL in the installer and the fact that it upgrades an existing Microsoft Active Directory Authentication Library package installation.


PRMerger9:
@David-Engel : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #3686, opened 2019-11-25
Replaced code that returns the "instance_name"

Rob-S:
Even on a standalone instance, just like on an FCI instance mentioned in the "FCI_network_name" paragraph above it, the "SELECT @@SERVERNAME" would return the "full replica name", not just the "instance_name" as the paragraph currently states. The type of instance makes no difference in what "SELECT @@SERVERNAME" returns - always the "full replica name" - nor in the "InstanceName" property - always the instance name, or null for an unnamed "default server instance".


PRMerger20:

MikeRayMSFT:
@Rob-S  - thank you for the submission. Reviewing.
@kfarlee - Please review.

jborsecnik:
@MikeRayMSFT, okay to merge?

MikeRayMSFT:
#sign-off

MikeRayMSFT:
@jborsecnik - it's ready to merge. #sign-off

PR #3731, opened 2019-12-03
Update datename-transact-sql.md

ldemidov:
Fix the datename/tz example results.   They are right for datepart but not datename.


PRMerger7:
@ldemidov : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MikeRayMSFT:
Thank you @ldemidov - #sign-off

PR #3740, opened 2019-12-04
loopback_detection update

briancarrig:
Remove mention of only including for backward compatibility with SQL 2000 (!). Remove advice to use peer-to-peer instead of bi-directional (on bi-directional page we say it can outperform P2P - which is true).


PRMerger19:
@briancarrig : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@briancarrig Thanks, Brian! #sign-off

PRMerger20:

PR #3819, opened 2019-12-17
Update SSIS 2019 for Linux doc

yidetu:
Update SSIS 2019 for Linux doc with right repository

### DIFF
--- a/docs/linux/sql-server-linux-setup-ssis.md
+++ b/docs/linux/sql-server-linux-setup-ssis.md
@@ -37,7 +37,7 @@ To install the `mssql-server-is` package on Ubuntu, follow these steps:
    ```
 ::: moniker-end
 <!--SQL Server 2019 on Linux-->
-::: moniker range="= sql-server-linux-2019 || = sql-server-2019"
+::: moniker range="= sql-server-linux-2019 || = sql-server-2019 || = sqlallproducts-allversions"
    ```bash
    sudo add-apt-repository "$(curl https://packages.microsoft.com/config/ubuntu/16.04/mssql-server-2019.list)"
    ```
@@ -86,7 +86,7 @@ To install the `mssql-server-is` package on RHEL, follow these steps:
    ```
 ::: moniker-end
 <!--SQL Server 2019 on Linux-->
-::: moniker range="= sql-server-linux-2019 || = sql-server-2019"
+::: moniker range="= sql-server-linux-2019 || = sql-server-2019 || = sqlallproducts-allversions"
    ```bash
    sudo curl -o /etc/yum.repos.d/mssql-server.repo https://packages.microsoft.com/config/rhel/7/mssql-server-2019.repo
    ```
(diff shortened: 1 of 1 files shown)


yidetu:
#sign-off

PRMerger14:
@yidetu: I'm sorry - only the author of this article, @lrtoyou1223, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

ShannonLeavitt:
Hello @rothja - Would you review the proposed updates and indicate whether the PR can be merged? Thanks!

PRMerger9:
@yidetu : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@yidetu Thank you for catching this and letting us know. However, I think we have a bigger change to make here, because the article needs to work for both 2017 and 2019 content based on the version selector on docs. The easiest way is to just instruct the user to first install SQL Server. Based on their version, they would have configured the appropriate repo. I'm going to add some suggestions for this change to your PR and you can accept them and still remain the contributor on this request. Thank you! @Vanmsft FYI

PRMerger16:
@yidetu : Thanks for your contribution! The author(s) have been notified to review your proposed change.

yidetu:
Jason, thanks for reviewing the change. I didn't aware that this doc is for both versions.
You suggested change is not ideal.  Install SQL engine for Linux is not a prerequisite of install SSIS. So it's not reasonable to put install SQL Server as first step.
I've updated the doc with a moniker switch. Please help to take a look.

rothja:
@yidetu Your change looks good. But I suggest adding one moniker to the 2019 condition in case the user selects "all versions" in the filter list. In that case, I think showing the 2019 example makes sense rather than both lines which would be potentially confusing:

```
::: moniker range="= sql-server-linux-2019 || = sql-server-2019 || = sqlallproducts-allversions"
```

Can you make that change to both 2019 sections? Then I think this is great. Thank you!

PRMerger14:
@yidetu : Thanks for your contribution! The author(s) have been notified to review your proposed change.

yidetu:
Thanks for the suggestion, updated.

rothja:
@yidetu Perfect. Thanks for patiently working with me through this. #sign-off @vanmsft FYI.

PRMerger15:
@rothja: I'm sorry - only the author of this article, @lrtoyou1223, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

MightyPen:
@rothja @yidetu Jason & Yide, there is no such Versioning moniker named 'sql-server-2019', thus PR **13294** in the public repo 'sql-docs-pr' has failed.

You must mean moniker 'sql-server-ver15' (as in SELECT \@\@VERSION returns a value of "15.whatever".

yidetu:
I've updated to "::: moniker range=">= sql-server-linux-ver15 || >= sql-server-ver15 || =sqlallproducts-allversions"", thanks Gene.

rothja:
@MightyPen I'm sorry I didn't catch that when reviewing.

PR #3848, opened 2019-12-20
Update Powershell script to Change ownership for all subscriptions ow…

UmaSM-MS:
…ned by a specific user

Change the parameter within the IF block of the ForEach Block from $previousOwner to $currentOwner as the parameter has been declared with the name $currentOwner. The script does not work when $previousOwner is used


PRMerger20:
@UmaSM-MS : Thanks for your contribution! The author(s) have been notified to review your proposed change.

maggiesMSFT:
#sign-off
Thanks, @UmaSM-MS ! Good catch. I've made the change.

PR #3856, opened 2019-12-22
Update deploy-big-data-tools.md

suresh-kandoth:
SQLCMD has dependency on ODBC package and it is better to provide the right sequence of install for users. Before installing SQLCMD, request user to install ODBC package from https://www.microsoft.com/en-us/download/confirmation.aspx?id=36434&6B49FDFB-8E5B-4B07-BC31-15695C5A2143=1


PRMerger8:

rothja:
@suresh-kandoth Thanks for this suggestion. Apologies for the delay.

PR #3857, opened 2019-12-22
Update deploy-big-data-tools.md

suresh-kandoth:
Please consider moving the section "Which tools are required?" before the table with all the tools. 
Also it is confusing with the flow of this page because the instructions for Python and Azure Data Studio are appearing again after the table.


PRMerger17:

rothja:
@suresh-kandoth Thanks for this feedback. Since the change you're suggesting is not in this PR, I am going to close out the PR. But I'll make sure that we take note of the recommendation for possible updates later. @MikeRayMSFT could you please see Suresh's suggestion for possible improvements in the future? Thanks!

PR #3919, opened 2020-01-07
Update polybase-linux-setup.md

pradm:
RECONFIGURE [WITH OVERRIDE]; this errors out with syntax error on both SSMS and ADS it should be RECONFIGURE WITH OVERRIDE;


PRMerger12:
@pradm : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@pradm Thanks for the fix and apologies for the delay.

PR #3921, opened 2020-01-07
Update sql-server-linux-active-directory-authentication.md

liweiSecurity:
Kerberos authentication only works for remote session in SQL Windows platform. SQL Linux makes it available for local connection.

I added following statement in section 'Connect to SQL Server using AD Authentication'
Different from SQL Windows, Kerberos authentication works for local connection in SQL Linux. You can test the Windows Authentication in the SQL Linux box.

### DIFF
--- a/docs/linux/sql-server-linux-active-directory-authentication.md
+++ b/docs/linux/sql-server-linux-active-directory-authentication.md
@@ -190,6 +190,7 @@ Make sure you've installed the [mssql-tools](sql-server-linux-setup-tools.md) pa
 ```bash
 sqlcmd -S mssql-host.contoso.com
 ```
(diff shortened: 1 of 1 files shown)


PRMerger8:

Dylan-MSFT:
To prevent confusion, can you please specify that it only works if you specify the fqdn of the host for local connections. Maybe something like the following:

Different from SQL Windows, Kerberos authentication works for local connection in SQL Linux. However, you still need to provide the FQDN of the host, and AD authentication will not work if you attempt to connect to '.', '(local)', '127.0.0.1', etc.

PRMerger17:

Dylan-MSFT:
#signoff

PR #3924, opened 2020-01-08
Updated header for creating login with server auth and T-SQL

wattengard:
Header for creating login using SQL Server Authentication using T-SQL said "... with SSMS" instead.


PRMerger19:
@wattengard : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #3929, opened 2020-01-08

setemsft:
Removing the limitation of supplementary characters in Replication due to an internal table having an NTEXT column. As of SQL Server versions 2014 SP3, 2016 SP2, and 2017 CU5, the NTEXT has been replaced with NVARCHAR(max), thus making supplementary characters possible.

@pmasl - can you have a look?


PRMerger17:
@setemsft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

pmasl:
#sign-off

PR #3931, opened 2020-01-09

setemsft:
Removing misleading note about UTF8 flag being applicable to Version 90 _SC and _VSS collations. The note is misleading due to the following reasons:
- There are no _VSS collations in any version except 140. 
- For collations version 100 and 140, the UTF8 flag can also be applied only to _SC (both 100 and 140) and _VSS (140) collations. The _VSS implies _SC, and _SC is what's really needed for the UTF8 flag to apply.

The "can't be applied" section" also only references version 90 for _SC and _VSS, which is wrong due to the same reason.

The best thing is to simply remove version splitting, and say that all linguistic collations must have either _SC or _VSS to have UTF8 flag applicable to them. I simplified the language and removed the unnecessary version details.

@pmasl @bluefooted - can you please review?


PRMerger15:
@setemsft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

pmasl:
#sign-off

PR #3970, opened 2020-01-16
Update database-files-and-filegroups.md

pituach:
The following statement in the document is inaccurate:

"Read/write data filegroups and log files cannot be placed on an NTFS compressed file system. Only read-only databases and read-only secondary filegroups can be put on an NTFS compressed file system."

It is NOT Supported - True. It is NOT Recommended - True. It is a type of a hack - True as well. But it can be done! I showed how to do it in a post I published in 2013 in my blog:
http://ariely.info/Blog/tabid/83/EntryId/118/shrink-SQL-database-during-restore.aspx

I did a small change in the statement so it will provide the same information but will be accurate (replacing "cannot with not supported" for example:

My new version: "Read/write data filegroups and log files are not supported on an NTFS compressed file system. Only read-only databases and read-only secondary filegroups allowed to be put on an NTFS compressed file system."

### DIFF
--- a/docs/relational-databases/databases/database-files-and-filegroups.md
+++ b/docs/relational-databases/databases/database-files-and-filegroups.md
@@ -61,7 +61,7 @@ ms.author: "sstein"
 > [!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)] data and log files can be put on either FAT or NTFS file systems. On Windows systems, we recommend using the NTFS file system because the security aspects of NTFS. 
 
 > [!WARNING]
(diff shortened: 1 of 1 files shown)


PRMerger13:

pituach:
Hi, I will add the missing "are". sry for the grammar mistake

PRMerger13:

pmasl:
#sign-off

PRMerger8:
@pmasl: I'm sorry - only the author of this article, @stevestein, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #4001, opened 2020-01-22
Added xlsx and docx as common formats

greglow:
These are very common formats now and really should be in this list


PRMerger7:

maggiesMSFT:
#sign-off

PR #4010, opened 2020-01-22
Update to add clarity to how TRANSLATE works

greglow:
The previous description makes it sound like TRANSLATE will not make multiple replacements. Many people think it won't replace say all the square brackets when using TRANSLATE('[[00]] 2343:2342', '[]:', '()-') and that only the first instance of each character would be replaced. This isn't how the function works, so I've added text to clarify.


PRMerger7:

rothja:
@greglow Thanks for the improved explanation.

PR #4011, opened 2020-01-23
About locking with REORGANIZE

rrg92:
Original text says that REORGANIZE not block concurrent readers and writers. But is not true.

Use following code to check this:

```sql
-- create some test table
create table __reorgtest(c int);

-- create sample nonclustered index
create index x on __reorgtest(c);

-- insert some data
insert into __reorgtest values(2020);
go 1000


-- Open another session and run this (we call Session X):
begin tran;;
alter index x on __reorgtest reorganize;

-- Open another session and run this (we call session Y):
begin tran;;
select count(*) from __reorgtest

-- On this session, check tran_locks! 
select * from sys.dm_tran_locks where request_session_id in (  X,Y ) -- replace and y by sessions id opened above.

-- Current behavior on my tests: Session Y gets blocked by X.
-- THe inverse can be checked: Try start first a select (wth holdlock) or update operation on some row, and try start X(reorganize). Reorganize gets blocked!

```

Behavior checked on 
* Microsoft SQL Server 2019 (RTM-GDR) (KB4517790) - 15.0.2070.41 (X64)   Oct 28 2019 19:56:59   Copyright (C) 2019 Microsoft Corporation  Developer Edition (64-bit) on Windows 10 Home Single Language 10.0 <X64> (Build 18363: ) 
* Microsoft SQL Server 2016 (SP2-CU10) (KB4524334) - 13.0.5492.2 (X64)   Oct  4 2019 19:14:08   Copyright (c) Microsoft Corporation  Standard Edition (64-bit) on Windows Server 2012 R2 Standard 6.3 <X64> (Build 9600: )

### DIFF
--- a/docs/t-sql/statements/alter-index-transact-sql.md
+++ b/docs/t-sql/statements/alter-index-transact-sql.md
@@ -252,7 +252,7 @@ PARTITION
  REORGANIZE a **rowstore** index  
  For rowstore indexes, REORGANIZE specifies to reorganize the index leaf level. The REORGANIZE operation is:  
   
--   Always performed online. This means long-term blocking table locks are not held and queries or updates to the underlying table can continue during the ALTER INDEX REORGANIZE transaction.  
+-	It locks the object using schema shared locks (Sch-S) and exclusive locks. If another session reads data using NOLOCK hint, it will bypass the REORGANIZE (REBUILD blocks all, including statements with NOLOCK hint). Statements that modify data are always blocked.
 -   Not allowed for a disabled index  
 -   Not allowed when ALLOW_PAGE_LOCKS is set to OFF  
 -   Not rolled back when it is performed within a transaction and the transaction is rolled back.  
--- a/docs/t-sql/statements/alter-index-transact-sql.md
+++ b/docs/t-sql/statements/alter-index-transact-sql.md
@@ -252,13 +252,13 @@ PARTITION
  REORGANIZE a **rowstore** index  
  For rowstore indexes, REORGANIZE specifies to reorganize the index leaf level. The REORGANIZE operation is:  
   
--   Always performed online. This means long-term blocking table locks are not held and queries or updates to the underlying table can continue during the ALTER INDEX REORGANIZE transaction.   
--   Not allowed for a disabled index  
--   Not allowed when ALLOW_PAGE_LOCKS is set to OFF  
--   Not rolled back when it is performed within a transaction and the transaction is rolled back.  
+-   Always performed online. This means long-term blocking table locks are not held and queries or updates to the underlying table can continue during the ALTER INDEX REORGANIZE transaction.
+-   Not allowed for a disabled index.
+-   Not allowed when ALLOW_PAGE_LOCKS is set to OFF.
+-   Not rolled back when it is performed within a transaction and the transaction is rolled back.
 
 > [!NOTE]
-> When ALTER INDEX REORGANIZE uses explicit transactions (e.g., ALTER INDEX inside a BEGIN TRAN ... COMMIT/ROLLBACK) instead of the default implicit transaction mode, the locking behavior of REORGANIZE becomes more restrictive, potentially causing blocking. For more information on implicit transactions, see [SET IMPLICIT_TRANSACTIONS &#40;Transact-SQL&#41;](../../t-sql/statements/set-implicit-transactions-transact-sql.md)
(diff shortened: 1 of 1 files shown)


PRMerger9:
@rrg92 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@rrg92 Thanks for the feedback. @pmasl Could you review this change and sign off on it if you approve? Thank you!

ktoliver:
@pmasl Hi, could you take a look at this PR? Thank you.

rrg92:
Hey guys, this PR make some sense?

ktoliver:
@pmasl Could you review this PR? Thanks.

pmasl:
Your test has REORG encapsulated in an open tran. That way anything blocks. The use of REORG as an autocommit statement is done online, which means only the singleton record impacted blocks or is blocked. In summary, your test doesn't represent the intended or real-use case of REORG.

rrg92:
Thanks @pmasl !

I opened a Extended events session to check this that you says and I confirmed that. 
This is lock held into this object without begin tran:

image


THis is lock held using begin tran:

image


I search in SQL docs and not find anything about that fact autocommit mode change this behavior... Do you guess that adding this info into doc is pertinent? I guess most of users dont know this.

pmasl:
Locking behavior is different between explicit transactions and auto-commit (aka implicit transactions which is the default) for every (or most) DML and DDL. 

If any I'd suggest adding to that note something to the effect of: 

"If using explicit transactions instead of the default implicit transaction mode, the locking behavior of REORGANIZE becomes more restrictive. For more information on implicit transactions, see [SET IMPLICIT_TRANSACTIONS (Transact-SQL)](../../t-sql/statements/set-implicit-transactions-transact-sql.md)."

PRMerger6:
@rrg92 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rrg92:
I changed, putting original text and adding a Note on about the fact of autocommit.

docs/t-sql/statements/alter-index-transact-sql.md
```diff
@@ -252,11 +252,14 @@ PARTITION
...
  For rowstore indexes, REORGANIZE specifies to reorganize the index leaf level. The REORGANIZE operation is:  
   
--   Always performed online. This means long-term blocking table locks are not held and queries or updates to the underlying table can continue during the ALTER INDEX REORGANIZE transaction.  
+-   Always performed online. This means long-term blocking table locks are not held and queries or updates to the underlying table can continue during the ALTER INDEX REORGANIZE transaction.   
 -   Not allowed for a disabled index  
 -   Not allowed when ALLOW_PAGE_LOCKS is set to OFF  
 -   Not rolled back when it is performed within a transaction and the transaction is rolled back.  
 
+> [!NOTE]
+> If using explicit transactions instead of the default implicit transaction mode (e.g., ALTER INDEX within BEGIN TRAN ... COMMIT/ROLLBACK), the locking behavior of REORGANIZE becomes more restrictive. For more information on implicit transactions, see [SET IMPLICIT_TRANSACTIONS &#40;Transact-SQL&#41;](../../t-sql/statements/set-implicit-transactions-transact-sql.md)
```
pmasl, line 261: The way it reads now almost seems like implicit is inside a BGEIN/END TRAN. Suggest the following:

When ALTER INDEX REORGANIZE uses explicit transactions (e.g., ALTER INDEX inside a BEGIN TRAN ... COMMIT/ROLLBACK) instead of the default implicit transaction mode, the locking behavior of REORGANIZE becomes more restrictive. For more information on implicit transactions, see [SET IMPLICIT_TRANSACTIONS &#40;Transact-SQL&#41;](../../t-sql/statements/set-implicit-transactions-transact-sql.md).

rrg92: Do you guess that we can add following bold part to clarify how more restrictive it can be?

he locking behavior of REORGANIZE becomes more restrictive, **causing long-term blocking table locks**

pmasl: "long" is subjective - it's as long as the reorg takes. It may be short or not. I'd be ok with ", potentially causing blocking."

rrg92: Yes. I agree the word "potentially" is the right one for this. Thanks

PRMerger7:
@rrg92 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rrg92:
All done, text will looks like this:

image

pmasl:
#sign-off

PRMerger15:

PR #4024, opened 2020-01-25
Changed "server" to "instance"

rrg92:
This is avoid confusion of some users that can guess Windows need be rebooted.


PRMerger16:
@rrg92 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@rrg92 That is an excellent clarification. :) Thank you very much.

PR #4040, opened 2020-01-28
Add more detail about SQL Server named instance.

liweiSecurity:
The current version does not cover the named instance, which will cause the Data movement fail to start . I added following statement: If the head node is a named instance, you need to manually add the port of SQL Server to Windows firewall on head node, also start the SQL Browser on head node.


PRMerger7:

PRMerger12:
@rothja : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@liweiSecurity Thanks for the suggestion. I made some minor changes to the wording which you can review. But I believe it retains the technical meaning. I'll merge this change in. Thank you!

PR #4096, opened 2020-02-06
SQL2017 CU17, CU18 contained same R/Python Cabs

Aniruddh25:
Since CU17, CU18 had no updates to the runtime, mention them in the same column.


PRMerger7:
@Aniruddh25 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #4122, opened 2020-02-10
small syntax fix

taodailOryx:


MightyPen:
Good fix.  Thanks.  Merging now.

PR #4123, opened 2020-02-11
Update xp-cmdshell-transact-sql.md

suresh-kandoth:
Example A has an extra single quote at the end which results in parsing error as copied from the code block. Need this fixed.


PRMerger12:

PR #4161, opened 2020-02-18
Update Header Title to Reflect Document Name

LarnuUK:
The Header Title of this document was still *"GO (Transact-SQL)"*. The document itself states *"GO is **not** a Transact-SQL statement"*, and the title of the document reflects this therefore the title header should also be reflective of this. The incorrect information is also displayed in things like Google Searches, which could cause confusion for users performing searches.


PRMerger19:

rothja:
@LarnuUK Thank you for this recommended change, and apologies for the long delay. #sign-off

PR #4190, opened 2020-02-21
Missing change in Replicate function on 90 to 100

taroarai:
It appear the following change was also added to the database compatibility level 100, but it was only noted on following document.
I understand it is only affecting SQL Server 2012, which is extended support phase, or SQL Server 2008/2008 R2 with ESU, but could you add this change to this Docs?

Behavior Changes to Database Engine Features in SQL Server 2008
https://docs.microsoft.com/en-us/previous-versions/sql/sql-server-2008/ms143359(v=sql.100)?redirectedfrom=MSDN#replace
*****
REPLACE Function

In SQL Server 2005, trailing spaces specified in the first input parameter to the REPLACE function are trimmed when the parameter is of type char. For example, in the statement SELECT '<' + REPLACE(CONVERT(char(6), 'ABC '), ' ', 'L') + '>', the value 'ABC ' is incorrectly evaluated as 'ABC'.
In SQL Server 2008, trailing spaces are always preserved. For applications that rely on the previous behavior of the function, use the RTRIM function when specifying the first input parameter for the function. For example, the following syntax will reproduce the SQL Server 2005 behavior SELECT '<' + REPLACE(RTRIM(CONVERT(char(6), 'ABC ')), ' ', 'L') + '>'.
*****


PRMerger9:

Issue #4219, opened 2020-02-26
READONLY: Definition and explanation is misslading


Bartolomeus-649:
READONLY  for a parameter in the function prototype can only be used for table value parameters.

The documentation as it is now, both in the syntax of how to define the prototype of the function and in the explanation of READONLY, gives no info that READONLY can only be used with table value parameters. On the contrary, the text in the explanation of READONLY  indicate that it can be used for other types than table value types.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 71bc28bf-993a-79e3-1a83-b39052bab882
* Version Independent ID: 73350c77-8839-dcd3-af67-9342a49dbf22
* Content: [CREATE FUNCTION (Transact-SQL) - SQL Server](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver15)
* Content Source: [docs/t-sql/statements/create-function-transact-sql.md](https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/statements/create-function-transact-sql.md)
* Product: **sql**
* Technology: **t-sql**
* GitHub Login: @CarlRabeler
* Microsoft Alias: **carlrab**

srutzky:
> the text in the explanation of READONLY indicate that it can be used for other types than table value types.

Not only is that misleading and incorrect, but the current wording for `READONLY` also implies that it's optional for TVPs, and that's also provably false.

I'm submitting a PR for this.

srutzky:
@Bartolomeus-649 and @CarlRabeler Please see PR #4224

PR #4224, opened 2020-02-26
Fix "READONLY" definition for CREATE FUNCTION

srutzky:
This PR fixes Issue #4219 

Previous wording:

> READONLY  
> Indicates that the parameter cannot be updated or modified within the definition of the function. If the parameter type is a user-defined table type, READONLY should be specified.

 implied that the `READONLY` keyword:

1. could be used by parameter types other than user-defined table types (TVPs)
2. was optional (even if recommended) for user-defined table types (TVPs)

The following tests prove both implications are incorrect, and that `READONLY` is;

1. not optional for user-defined table types
1. not valid for other parameter types

```sql
USE [tempdb];

------------------------------------------

GO
CREATE PROCEDURE dbo.TestReadOnlyWithNonTVP
(
	@Param INT = 5 READONLY
)
AS

SELECT @Param;
GO
/*
Msg 346, Level 15, State 1, Procedure TestReadOnlyWithNonTVP, Line XX [Batch Start Line YY]
The parameter "@Param" can not be declared READONLY since it is not a table-valued parameter
*/

------------------------------------------

CREATE TYPE dbo.TestTableType AS TABLE
(
	Col INT
);


GO
CREATE PROCEDURE dbo.TestNoReadOnlyWithTVP
(
	@Param dbo.TestTableType
)
AS

SELECT * FROM @Param;
GO
/*
Msg 352, Level 15, State 1, Procedure TestNoReadOnlyWithTVP, Line XX [Batch Start Line YY]
The table-valued parameter "@Param" must be declared with the READONLY option.
*/

------------------------------------------
```

Take care,
Solomon...
https://SqlQuantumLift.com/
https://SqlQuantumLeap.com/
https://SQLsharp.com/

### DIFF
--- a/docs/t-sql/statements/create-function-transact-sql.md
+++ b/docs/t-sql/statements/create-function-transact-sql.md
@@ -1,7 +1,7 @@
 ---
 title: "CREATE FUNCTION (Transact-SQL) | Microsoft Docs"
 ms.custom: ""
-ms.date: 11/06/2018
+ms.date: 02/26/2020
 ms.prod: sql
 ms.prod_service: "database-engine, sql-database"
 ms.reviewer: ""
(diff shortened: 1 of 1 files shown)


PRMerger20:

PR #4228, opened 2020-02-26
Fix NO_OUTPUT, should be outside cmd string

LowlyDBA:
Currently the code does not run. NO_OUTPUT option needs to be *outside* of the string command being executed.


PRMerger13:
@LowlyDBA : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #4291, opened 2020-03-06
Update resource-governor-classifier-function.md

andrecrms:
I added a note in the steps of the classification process because the way it is today it may imply that the resource governor makes use of logon triggers and this is not true.


PRMerger18:
@andrecrms : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger8:
@julieMSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

julieMSFT:
#sign-off

PR #4323, opened 2020-03-11
Fixed broken links

evertdhont:


PRMerger12:
@evertdhont : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
Hi @evertdhont Thanks for correcting the syntax on the two links. The include file reference is actually correct as-is, and it works correctly on the live site--I've reverted that change. Thanks again.

PR #4373, opened 2020-03-19
Update querying-data-in-a-system-versioned-temporal-table.md

JediSQL:
Line 29: added "When using a table alias with a temporal tables, the **FOR SYSTEM_TIME** clause must included between the temporal table name and the alias (see "Query for a specific time using the AS OF sub-clause" second example, below)."
If someone can make "Query for a specific time using the AS OF sub-clause" a hyperlink, that would be cool.


PRMerger13:
@JediSQL : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #4410, opened 2020-03-25
Update to syntax convention

JohnOsborneOPM:
I don't think the trailing period on 'database_name.schema_name.table_name.' should be there but I may be mistaken


PRMerger10:
@JohnOsborneOPM : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #4452, opened 2020-04-01
Apply to multiple versions

way0utwest:
The ability to update a base table isn't limited to current, so this inclusion is misleading. This should be in all versions of SQL Server, subject to the limitations in the CREATE VIEW item.


PRMerger16:

stevestein:
@way0utwest - thanks for clarifying this one! 

#sign-off

PR #4495, opened 2020-04-07
The event name obtained by xEvent is different

MasayukiOzawa:
The name of the event obtained by xEvent is sql_statement_completed, but at https://www.microsoft.com/en-us/download/details.aspx?id=53595 it is stated as follows
> sql_batch_completed event for application / xevent trace assessments
Syntax error check is not performed unless sql_batch_completed event is fetched.


PRMerger10:
@MasayukiOzawa : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #4524, opened 2020-04-09
Remove reference to pymssql

joshts0:
We updated sqlmlutils to use pyodbc instead of pymssql - no longer need to install pymssql


PRMerger12:

garyericson:
#sign-off

PR #4564, opened 2020-04-15
Add prominent warning note to RAISERROR doc

jkaye2012:
SET_XACT_ABORT has a nice, prominent warning about an important difference between RAISERROR and THROW (and why THROW should be used in all new applications). Unfortunately, the RAISERROR page has only a single sentence that is easy to miss.

I added a very similar note to the top of RAISERROR. Hopefully, this will help point people to THROW instead.

### DIFF
--- a/docs/t-sql/language-elements/raiserror-transact-sql.md
+++ b/docs/t-sql/language-elements/raiserror-transact-sql.md
@@ -34,6 +34,9 @@ monikerRange: ">=aps-pdw-2016||=azuresqldb-current||=azure-sqldw-latest||>=sql-s
 # RAISERROR (Transact-SQL)
 [!INCLUDE[tsql-appliesto-ss2008-all_md](../../includes/tsql-appliesto-ss2008-all-md.md)]
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


PRMerger12:
@jkaye2012 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@jkaye2012 Thanks for this suggestion. Is this the same note from the SET_XACT_ABORT article? I have one request on this change if you could make it. Could you move the note to below the intro paragraph? We try not to start articles with notes when we're adding new ones. Let me know if you can go to the file changes tab and click edit to make the change the recommit it. Thank you!

jkaye2012:
Yeah, same basic note copied from there, added a word or two for clarity, but the wording is consistent. Sure, I'll make the change.

jkaye2012:
Actually, I just realized something - I put this on top because that's how it is here:

https://docs.microsoft.com/en-us/sql/t-sql/statements/set-xact-abort-transact-sql?view=sql-server-ver15

This is also verbatim the message that I included in this PR. It's not a big deal to me though, please just let me know how you'd like to proceed!

rothja:
@jkaye2012 good point! I'll merge and fix both later. Thanks!!

jkaye2012:
Thank you!

PR #4579, opened 2020-04-17
computed_column_expression clarifications

tkhadimullin:
- Trying to clarify a point that expression must return a value. 

An example of expression that appears to be valid but fails the syntax check: https://stackoverflow.com/q/61089445/12339804

- Adding an example of how bitwise operators can be used here too.


PRMerger18:
@tkhadimullin : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #4629, opened 2020-04-23
Fixed syntax in JSON_MODIFY example

ncsmikewoods:
The `path` parameter needs to be in single quotes instead of double quotes.  This was correct in all of the examples except for the last one.


PRMerger10:
@ncsmikewoods : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@ncsmikewoods Thanks for the fix! Sorry for the delayed response.

PR #4635, opened 2020-04-24
Added comma after scan_phase in the example query.

sql3d:
The example query did not have a commas after multiple column names, which caused the next column name to become an alias for the previous column name.  For example, no comma between scan_phase and error_count, which then made "error_count" the column alias for scan_phase.
Commas added after: scan_phase, tran_count, and schema_change_count


PRMerger13:
@sql3d : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@sql3d Thanks for the fix. #sign-off

PRMerger10:

PR #4646, opened 2020-04-27
Update change-the-ip-address-of-a-failover-cluster-instance.md

alhandra:
A reference to SQL Server Configuration Manager was missing from this page. Failure to do this makes SQL Server Configuration Manager show wrong information.


PRMerger20:
@alhandra : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MashaMSFT:
Hi @alhandra , thank you for your contribution! But actually, unfortunately this change is inaccurate. Things like the service, and the IP address, should typically be managed with the Failover Cluster Manager when the SQL Server instance is clustered, rather than using the SQL Server Management Console. As such, it is not necessary to update the IP address using the SQL Server Management Console when you have a SQL Server instance participating in a failover cluster. 

Thank you for your submission! Unfortunately, I will have to close out this PR though. 
Masha from the SQL Docs team

PR #4654, opened 2020-04-27
Updating verbiage for (lack of) SQL FCI support

joshts0:
Currently we say this:
Do not install R Services on a failover cluster. The security mechanism used for isolating R processes is not compatible with a Windows Server failover cluster environment.

But our 2017 docs say this:
Installing Machine Learning Services isn't supported on a failover cluster in SQL Server 2017. It's supported with SQL Server 2019.

When we say: "The security mechanism used for isolating R processes is not compatible with a Windows Server failover cluster environment." We really mean that is not supported in an FCI environment.

Since AGs use WSFC, this is misleading verbiage, when we really mean that it is not supported on SQL FCI.

Updating verbiage to be more clear.


PRMerger19:

PR #4682, opened 2020-05-01
Update load-data-to-from-excel-with-ssis.md

simes1:
Under "Export long text values" the second option to create the excel file using a sql statement and columns of type LONGTEXT does NOT appear to work.  Only the first option works - having at least one row with data in each column > 255 double byte characters (i.e. 512 normal ones) _already_ before writing anymore rows works.  So one can only make this work with an initial, manually created, dummy row.

Delete the second option - it may work for other data destinations but not for Excel.

### DIFF
--- a/docs/integration-services/load-data-to-from-excel-with-ssis.md
+++ b/docs/integration-services/load-data-to-from-excel-with-ssis.md
@@ -248,8 +248,6 @@ Before you can successfully save strings longer than 255 characters to an Excel 
 
 -   If an existing destination table already contains rows of data, then the first few rows that are sampled by the driver must contain at least one instance of a value longer than 255 characters in the memo column.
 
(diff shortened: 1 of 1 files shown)


PRMerger14:
@simes1 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@chugugrace Could you please take a look at this recommended change to the article? Look at the files changed tab. If this is correct, could you sign off? Thank you!

rothja:
@chugugrace Because this is removing information that @simes1 has verified to not work, I am going to sign off on the change. If this is incorrect, please let me know and we can work on adding it back with more clarification. Thank you @simes1, and apologies for the delay! #sign-off

PRMerger6:
@rothja: I'm sorry - only the author of this article, @chugugrace, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #4703, opened 2020-05-03
Adding in correct qualifiers for clusterless HA table item

SeanGallardy:
Added in the 5 and 6 call outs for clusterless (read-scale) items to match the changes made for the 2017 (PR 4702). This keeps them in line and identical to each other and should cause less confusion.


PRMerger17:

MikeRayMSFT:
Sean - including @kfarlee  for review - as I thought that readscale and clusterless are a little difference. Either way, you're right we need to use the same term if it's the same thing. Waiting for PM to confirm.

MikeRayMSFT:
Thanks @SeanGallardy 

#sign-off

PR #4773, opened 2020-05-13
Make table name in stored procedure consistent

c-w:
Both `RxPlotHistogram` and `RPlotHist` use the same source table `nyctaxi_sample` but reference it with a different name which can be confusing to people going through the tutorial. This change makes the table names consistent.


PRMerger12:
@c-w : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #4869, opened 2020-05-28
Update ss-linux-cluster-availability-group-create-prereq.md

suresh-kandoth:
update AlwaysOn XE with new docs link.

### DIFF
--- a/docs/includes/ss-linux-cluster-availability-group-create-prereq.md
+++ b/docs/includes/ss-linux-cluster-availability-group-create-prereq.md
@@ -77,7 +77,7 @@ ALTER EVENT SESSION  AlwaysOn_health ON SERVER WITH (STARTUP_STATE=ON);
 GO
 ```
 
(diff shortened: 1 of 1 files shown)


ktoliver:
Hi @MikeRayMSFT - I can't find an author attribute for this include file. You've updated it several times--could you review the PR? Thanks.

ktoliver:
If the update is approved, the PR review team can remove the locale (en-us) from the link that's added.

ktoliver:
Hi @MikeRayMSFT - Could you take a look at this PR? Thanks.

rothja:
@suresh-kandoth Thanks for the suggested fix. @MikeRayMSFT I think we need to make this a relative link. To make sure we do it right, can you make the change in a separate PR to make sure we put a valid relative link in there. Then we can close this without merging as fixed. Thanks!

MikeRayMSFT:
I had a problem with the link and am fixing it now.

PR #4918, opened 2020-06-05
Update with read-intent replica specific details

MSFTPawelM:
This is related to the discussion in RFC 13466925 - in read-intent only AG replicas connections querying sys.dm_exec_requests view can get a read-intent error if they don't have correct readintent specified in connection string, this will cause us to display an error instead of the expected details.


PRMerger14:
@MSFTPawelM : Thanks for your contribution! The author(s) have been notified to review your proposed change.

pmasl:
#assign:@kfarlee

PR #4951, opened 2020-06-11
SSMS can also be used as SQL query tool

chanmmn:
SSMS can also be used as SQL query tool. I completed the 4 parts using SSMS. 
So I added:
or [SQL Server Management Studio](https://docs.microsoft.com/en-us/sql/ssms/download-sql-server-management-studio-ssms?view=sql-server-ver15)


PRMerger6:
@chanmmn : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
Hi @cawrites 

Could you review the proposed changes to your article?

Thanks!

Admin note: If the PR is approved, remove the locale from the docs.microsoft.com URL before merging. The PR review team can do this.

dphansen:
@chanmmn Thank you for your contribution. While it is correct that the you can perform the tutorial using SSMS or any query tool (sqlcmd, for example), the tutorial is designed to use Azure Data Studio. We will therefore leave the pre-reqs as is.

PR #5000, opened 2020-06-19
Update alter function schemabinding explanation

kimani-m:
- correcting the schemabinding definition and making it consistent with the definition in the create function documentation.


PRMerger7:
@kimani-m : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5002, opened 2020-06-19
Update alter-index-transact-sql.md

suresh-kandoth:
Migrating content from KB article https://support.microsoft.com/en-us/help/2292737: Rebuilding an index can result in fragmented index when ALLOW_PAGE_LOCKS option is not enabled. We will be retiring this KB and instead use the information from this doc page.


PRMerger16:

craigg-msft:
Closing per conversation.

PR #5024, opened 2020-06-22
Improve examples

FrankVonDerHoeh:
- According to the documentation of CAST and CONVERT, 101 is U.S. style.
- The two last bullet points have the same introduction, so that it is unclear what to use. Maybe the penultimate bullet point should be deleted?


PRMerger12:
@FrankVonDerHoeh : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevestein:
@FrankVonDerHoeh - thanks for the feedback.

The first bullet is saying to use unseparated numeric strings, and the second bullet is saying if you need to convert it to a specific style, use the CONVERT statement with the style parameter you want.

The example is showing that if you want US, use 101 to convert the unseparated string to US format. 

Hope that clarifies. We're going to go ahead and close this PR.

Thanks!

#please-close

PR #5053, opened 2020-06-29
TSQL - Print Error

pra6hu:
- After the insert of record, there is series of `COMMIT`, So the text in the `PRINT` Statement to be `COMMIT` instead of `INSERT`


PRMerger15:
@pra6hu : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5076, opened 2020-07-01
Fixed bug in sample code

briangithex:
On a typical SQL configuration, sample code generates an error, that the table "contains records where end of period is not equal to MAX datetime."


PRMerger9:
@briangithex : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5129, opened 2020-07-06
Clarify MAXDOP and coop scheduling example

jadarnel27:
There is a really useful illustration on this docs page, but there are some confusing and inaccurate aspects to it.

My goal with this PR is to preserve the intent of the illustration while correcting some of those issues, specifically:

- remove the reference to setting CPU affinity for 8 cores. I don't think a MAXDOP 32 query makes much sense on a server with 8 cores available to SQL Server.  It feels like this distracts from the example
- clarify statements about the number of operators in the theoretical plan
  - "*The query plan has a join between two tables, with a total of 3 query plan operators*" - this isn't possible in a parellel plan - there would be at least one "Gather Streams" operator (making the total 4)
  - "*Each of the operators is executed in parallel, and spawns 32 tasks*" - for that to be true, each operator really needs to be in a separate branch - which would mean there are 6 operators total (2 scans, 1 join, and 3 exchanges)
  - I have included a screenshot in this PR of an AdventureWorks2014 query plan that demonstrates this plan shape.  I think it would go a long way towards clarifying what is meant about the operators / plan shape.  I don't know how to add it to the docs though, so if someone could help with that it would be appreciated :)

image

- It's more accurate (and I think more clear, in the presence of the screenshot) to say that each *branch* spawns 32 tasks
- "*At any given time, the request has at least 99 tasks*" - I think the math is just wrong here.  Each branch gets 32 tasks, which is 96 total.  Plus one for the coordinator is 97.  There isn't a coordinator per branch
- "*a single request can spawn multiple tasks, and each task is assigned to a worker, up to the MAXDOP limit*" - tasks aren't limited to MAXDOP for a parallel query (they are limited to MAXDOP per branch).  The number of schedulers available for the extra parallel worker threads is what's limited by MAXDOP, which I think it the point being made in this example

For reference (it's probably not useful for the example), I used this query on SQL Server 2019 (against the AdventureWorks2014 database at compat level 150) to produce that execution plan:

```
SELECT 
    sod.SalesOrderDetailID, 
    p.ProductID
FROM Sales.SalesOrderDetail sod
    INNER MERGE JOIN Production.Product p 
        ON sod.ProductID = p.ProductID
OPTION (USE HINT ('ENABLE_PARALLEL_PLAN_PREFERENCE'));
```


PRMerger9:
@jadarnel27 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

pmasl:
Thank you, but the docs are undergoing review right now, and should be published with new examples in the next 24h. Feel free to review after that change.

jadarnel27:
@pmasl I'm glad to hear it's being improved.  I put a decent amount of effort into this PR to try and help based on that Twitter exchange and the related blog posts, so I'm not sure I will invest more time to review again after the change - in case the goal posts are still moving.

pmasl:
I will review and see what I can incorporate from your PR. Really appreciate your input here.

jadarnel27:
Thanks, Pedro.  I appreciate that.

PR #5168, opened 2020-07-13
Correction to verbiage

way0utwest:
These are not nonlogged operations. They are minimally logged.


PRMerger14:

PR #5231, opened 2020-07-23
Removing slashes that shouldn't be there

joshts0:
Removing slashes that shouldn't be there - the "\" is wrong and launchpad will not start if it is there


PRMerger18:

PR #5270, opened 2020-07-30
Update sql-server-configuration-manager.md

noonand:
Small misspelling


PRMerger20:
@noonand : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@noonand Thanks so much for finding and correcting this. When I was looking at the files changed tab, it was a bit like a needle in the haystack to find what had changed, but I found it. Appreciate any quality improvements like this that you spot. #sign-off

PR #5308, opened 2020-08-08
Update to Query Governor

retracement:
This documentation misleadingly infers that the query cost is a measure of execution time when in fact this value is really the execution time on "Nick's machine". I have updated this document to reflect and clarify that the setting is not a measure of seconds on it's running instance to avoid the scenario where someone is trying to use it to prevent queries for running for X seconds. I have also underlined the fact that the cost is based on estimations, not actuals.


PRMerger8:
@retracement : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger8:
@markingmyname : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5309, opened 2020-08-08
Update set-query-governor-cost-limit-transact-sql.md

retracement:
Related to my previous PR #5308 for the server configuration setting. This session configuration documentation misleadingly infers that the query cost is a measure of execution time when in fact this value is really the execution time on "Nick's machine". I have updated this document to reflect and clarify that the setting is not a measure of seconds on it's running instance to avoid the scenario where someone is using it to prevent queries for running for X seconds. I have also underlined the fact that the cost is based on estimations, not actuals. It also does not cancel running queries (like the documentation suggests) but instead will not execute a query over the threshold. I have fixed these issues.

### DIFF
--- a/docs/t-sql/statements/set-query-governor-cost-limit-transact-sql.md
+++ b/docs/t-sql/statements/set-query-governor-cost-limit-transact-sql.md
@@ -41,9 +41,9 @@ SET QUERY_GOVERNOR_COST_LIMIT value
 
 ## Arguments
  *value*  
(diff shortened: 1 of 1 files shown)


PRMerger12:
@retracement : Thanks for your contribution! The author(s) have been notified to review your proposed change.

CarlRabeler:
#reassign @pmasl Pedro, please review / assign as appropriate

ktoliver:
Hi @pmasl -- Could you review this community contribution from August? Thanks!

rothja:
@pmasl Could you take a look at the suggested changes and see if you agree with the improvements that @retracement is recommending here? Thank you!

MikeRayMSFT:
Emailed PM

ktoliver:
Hi @MikeRayMSFT - Do you know whether the PM is still considering the proposed updates? Thanks.

MikeRayMSFT:
@uc-msft  reviewed and approved. Publishing through private repo due to of PR. Many thanks to @retracement

PR #5319, opened 2020-08-11
update sys-sp-cdc-add-job-transact-sql.md for pollinginterval setting

sizhonms:
CDC pollinginterval can only be greater than or equal to 0 and less than 24 hours(Max: 86399 seconds)


PRMerger14:
@sizhonms : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5349, opened 2020-08-20
Missing Optional Source Table Alias Definition

CheetahChrome:
- Inserted `[ [ AS ] table_alias ]` which was missing for the source table definition line.
 - Added description paragraph for the table alias definition.
 - Modified the description paragraph for both tables to be similar and provide specific info for that alias (its owner).


PRMerger15:
@CheetahChrome : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5351, opened 2020-08-20
Update merge-transact-sql.md

aisbergde:
invalid example using merge into an edge-table. The example code is not working in SQL Server 2017.
A valid example is required, but I have no idea which code could work. At least a warning should be until a valid code example will be provided.
the "ON MATCH (Person-(livesIn)->City" causes the error when used in the merge statement.


PRMerger18:
@aisbergde : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@aisbergde Thank you for this suggestion. Let me look into it more and I'll get back to you as soon as possible. #reassign:rothja

rothja:
@aisbergde Apologies for the delay. I verified that this does not work in 2017 but works in 2019. I altered your commit to put it in the form of a short note before the sample. Thank you for pointing this out and sorry for the confusion!

PR #5358, opened 2020-08-22
Typo correction and additional warning

way0utwest:
The original post had "can not" instead of cannot, which is a spelling error. This page also does not mention transactions, which is an important concept for variables.

### DIFF
--- a/docs/t-sql/language-elements/variables-transact-sql.md
+++ b/docs/t-sql/language-elements/variables-transact-sql.md
@@ -25,13 +25,9 @@ A Transact-SQL local variable is an object that can hold a single data value of 
 * To save a data value to be returned by a stored procedure return code or function return value.
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


ktoliver:
@rothja - Would you review the proposed changes? If the update is approved, I can change words to contractions where relevant per Microsoft style. Thanks.

rothja:
This additional note is correct. @ktoliver So, we should be using a contraction "can't" instead of "cannot" per the style guide? Also, in the PR as it stands now, the wrong space was removed, but I think you see that "can notbe". Thanks.

ktoliver:
@rothja I'd suggest this:

`> [!NOTE]`
`> - Variables can't be used in views.` [Or, rewrite to avoid passive voice: "You can't use variables in a view" or "Views can't contain variables"]
`> - Changes to variables aren't affected by the rollback of a transaction.`

rothja:
@ktoliver  That works. Can you make the changes and sign off? Thanks.

ktoliver:
> @ktoliver That works. Can you make the changes and sign off? Thanks.

Done. Thanks!

PR #5359, opened 2020-08-22
Added variable verbiage

way0utwest:
This statement does not effect local or table variables.


rothja:
@way0utwest Thanks for the suggested improvement. Do you have some sample T-SQL that shows the point you're making here? It looks like maybe something you've tested recently? That would help me confirm and approve. Thanks! -Jason

way0utwest:
This should help
```
DECLARE @founders VARCHAR(50) = 'way0utwest';
DROP TABLE IF EXISTS MyTable;
CREATE TABLE MyTable
(founder VARCHAR(50));
INSERT dbo.MyTable (founder) VALUES ('way0utwest');
BEGIN TRAN;
SET @founders = 'rothja';
UPDATE dbo.MyTable SET founder = 'rothja';
ROLLBACK;
SELECT @founders AS f;
SELECT * FROM dbo.MyTable AS mt;

```

The change inside the transaction and rollback do not apply as they would for live data.

rothja:
@way0utwest Thanks! I'll verify on my side with your sample and approve the PR. Appreciate your time.

rothja:
#sign-off

PR #5429, opened 2020-09-10
Add "priority boost" deprecation warning

eldarku:
There is a deprecation warning for priority boost here:
https://docs.microsoft.com/en-us/sql/database-engine/configure-windows/configure-the-priority-boost-server-configuration-option?view=sql-server-ver15
To make documentation consistent, I suggest adding the same warning after that topic.


ktoliver:
Hi @pmasl or @rothja - Could one of you take a look at this pull request? Thanks!

rothja:
This looks reasonable and helpful. Thanks, @eldarku. @pmasl Pedro, please confirm if you agree that the note is good to add to this conceptual article considering it is in the article Eldar referenced in the comment above.

pmasl:
Indeed. Thanks!

#sign-off

PR #5127, opened 2020-07-06
Removed the section "By default the SSIS service does not support del…

KrishnakumarRukmangathan:
…egation" which is incorrectly added to this page.

This page is about SSIS Windows Service (MsDtsSrvr.exe) which is used for managing packages which are deployed and stored under MSDB, File System & SSIS package store using the old legacy package deployment model. The content on this section (which is proposed for deletion) is not relevant to the page. This section talks about SSIS Execution Runtime process (ISSERVEREXEC.exe) which is not related to SSIS Windows Service (MsDtsSrvr.exe). ISSERVEREXEC.exe is used only when executing the packages stored under SSISDB catalog using the new project deployment model.

PR #5435, opened 2020-09-11
Update integration-services-service-ssis-service.md

KrishnakumarRukmangathan:
The proposed change to delete this section was implemented in the original DMC, looks like it came back again during the consolidation of DMCs.
Removed the section "By default the SSIS service does not support del… #5127

This page is about SSIS Windows Service (MsDtsSrvr.exe) which is used for managing packages which are deployed and stored under MSDB, File System & SSIS package store using the old legacy package deployment model. The content on this section (which is proposed for deletion) is not relevant to the page. This section talks about SSIS Execution Runtime process (ISSERVEREXEC.exe) which is not related to SSIS Windows Service (MsDtsSrvr.exe). ISSERVEREXEC.exe is used only when executing the packages stored under SSISDB catalog using the new project deployment model. The information about delegation is already added under https://docs.microsoft.com/en-us/sql/integration-services/catalog/ssis-catalog?view=sql-server-ver15#ssisdb-catalog-and-delegation-in-double-hop-scenarios.

Further Need to add: < as available under https://docs.microsoft.com/en-us/previous-versions/sql/sql-server-2012/aa337083(v=sql.110)>
Delegation Is Not Supported
SQL Server Integration Services does not support the delegation of credentials, sometimes referred to as a double hop. In this scenario, you are working on a client computer, Integration Services is installed on a second computer, and SQL Server is installed on a third computer. Although SQL Server Management Studio successfully passes your credentials from the client computer to the second computer on which Integration Services is running, Integration Services cannot delegate your credentials from the second computer to the third computer on which SQL Server is running.

### DIFF
--- a/docs/integration-services/service/integration-services-service-ssis-service.md
+++ b/docs/integration-services/service/integration-services-service-ssis-service.md
@@ -365,13 +365,6 @@ When you install [!INCLUDE[ssISnoversion](../../includes/ssisnoversion-md.md)], 
 ### Connecting by using a Local Account  
  If you are working in a local Windows account on a client computer, you can connect to the [!INCLUDE[ssISnoversion](../../includes/ssisnoversion-md.md)] service on a remote computer only if a local account that has the same name and password and the appropriate rights exists on the remote computer.  
   
-### By default the SSIS service does not support delegation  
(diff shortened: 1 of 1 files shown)


PRMerger13:
@KrishnakumarRukmangathan : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5478, opened 2020-09-22
Clarifying that Domain Name should not be included in the IDENTITY ar…

EricBuMSFT:
…gument of the command.

We are encountering CSS cases where customers are using "DomainName\UserName" for the IDENTITY argument which causes CREATE EXTERNAL table to fail with
Msg 105019, Level 16, State 1, Line 83
EXTERNAL TABLE access failed due to internal error: 'Java exception raised on call to HdfsBridge_Connect: 
Error [Kerberos login failed (LoginContext.login) with provided principal/password.] occurred while accessing external file.'


PRMerger14:
@EricBuMSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

VanMSFT (approved):

VanMSFT:
#sign-off

PR #5490, opened 2020-09-25
Fix sample notebook download script

BobPusateri:
Single quotes causes curl to throw an error about https protocol not being supported or disabled in libcurl. Double quotes does not cause this issue.


PRMerger18:
@BobPusateri : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MikeRayMSFT:
@BobPusateri  - Bob, thanks for the submission. The behavior is a little bit different depending on the console application that is used. Some consoles use single quotes - others use double. I'll work the my team to figure out what we should use as a standard, and if there's a better way to explain it.

MikeRayMSFT:
@BobPusateri - thanks for this submission. My team and I are going to drop the quotes (both single and double) becasue curl appears to work fine without it. 

LMK if you still have a problem.

PRMerger17:

PR #5515, opened 2020-10-01
Clarification of NUMBER Oracle datatype

NathanMSFT:
I changed the Oracle type mapping reference section to reflect how SQL Server datatype for NUMBER Oracle datatype depends on if NUMBER is using precision or not. If Oracle datatype is NUMBER (no precision), we need to use Float for SQL Server. If Oracle NUMBER is defined with precision (NUMBER (p, s)), then SQL Server needs to use Decimal (p, s).


PRMerger20:
@NathanMSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MikeRayMSFT (approved):

PR #5551, opened 2020-10-08
Remove password parameter from documentation

ymekesser:
The `-P` command line parameter was removed with SSMS 18.0 due to security concerns. See [this post](https://cloudblogs.microsoft.com/sqlserver/2019/04/24/sql-server-management-studio-ssms-18-0-released-for-general-availability/) on the SQL Server blog.

In the documentation, the parameter was still listed, which confused me while trying to automate opening SSMS connections.

Whether the extra paragraph with the explanation is necessary can be discussed, but it could make things clearer for people trying similar things.

### DIFF
--- a/docs/ssms/ssms-utility.md
+++ b/docs/ssms/ssms-utility.md
@@ -64,9 +64,6 @@ Connect using Active Directory Authentication. The type of connection is determi
 [**-U** _username_]
 User name when connecting with 'SQL Authentication'
 
-[**-P** _password_]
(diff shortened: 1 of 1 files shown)


PRMerger7:
@ymekesser : Thanks for your contribution! The author(s) have been notified to review your proposed change.

markingmyname:
@ymekesser

Sorry for the delay, but this change has already been made to this article.

Thank you for your suggestion.

We will continue to close this PR.

PR #5598, opened 2020-10-19
Remove 250MB from the list of supported sizes for DTU-base databases

sredman:
Trying to use this value was returning errors from the fail-fast path when trying to create a database.

Per examination of the backend code, this values does not appear to be actually supported.


PRMerger6:
@stevestein : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevestein:
@sredman - thanks for the update.

#sign-off

PRMerger7:
@stevestein: I'm sorry - only the author of this article, @markingmyname, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

markingmyname:
#sign-off

PR #5599, opened 2020-10-20
Update create-table-transact-sql.md

HugoKornelis:
The syntax diagram showed that a column default could be set with only a constraint name and no default; that should have been the other way around (the name is optional, but the default clause is really needed to supply a default)


PRMerger14:
@HugoKornelis : Thanks for your contribution! The author(s) have been notified to review your proposed change.

markingmyname (approved):
@HugoKornelis 

Approved.

markingmyname:
#sign-off

PR #5627, opened 2020-10-23
Update alter-partition-function-transact-sql.md

Rob-S:
Based on the rest of the description, both before and after that statement, the partition containing the range value specified is the one that is dropped, not the one "into which the values from the dropped partition are merged"


PRMerger13:

markingmyname (approved):
Thanks for your update.

approved

markingmyname:
#sign-off

PR #5658, opened 2020-11-01
Depending on physical_name it would render a conversion error

FreddyAngarita:
I Addded try_cast to substring part because it avoids to generate an error if it encounters a value that is not convertible to int

### DIFF
--- a/docs/relational-databases/system-catalog-views/sys-pdw-nodes-partitions-transact-sql.md
+++ b/docs/relational-databases/system-catalog-views/sys-pdw-nodes-partitions-transact-sql.md
@@ -57,7 +57,7 @@ JOIN sys.pdw_nodes_tables AS NTables
 AND pnp.pdw_node_id = NTables.pdw_node_id  
 JOIN sys.pdw_table_mappings AS TMap  
     ON NTables.name = TMap.physical_name 
-    AND substring(TMap.physical_name,40, 10) = pnp.distribution_id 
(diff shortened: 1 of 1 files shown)


PRMerger20:
@FreddyAngarita : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@FreddyAngarita Thank you for the suggestion. @ronortloff Ron, could you take a look at this public PR suggestion and see if you approve or have comments? Thank you!

ronortloff:
@FreddyAngarita have you hit an error using the example or trying to enhance the code to make it more robust?  Try_cast in this case could cause someone to miss data when an error may be a valid outcome.

FreddyAngarita:
> @FreddyAngarita have you hit an error using the example or trying to enhance the code to make it more robust? Try_cast in this case could cause someone to miss data when an error may be a valid outcome.

Yes When I tried this command on a table that may have "exotic" names it redenred an error. So i needed to add the try_cast to continue query execution

julieMSFT:
@FreddyAngarita Since this was a work around due to your "exotic" names we do not feel the article needs updating and may cause data to be missed. Thank you for submitting the change to Microsoft Docs. We hope to hear from you again. I will go ahead and close this PR.

julieMSFT:
#assign:julieMSFT

julieMSFT:
#please-close

PR #5659, opened 2020-11-02
compatibility for Windows 7 and Windows 2008 R2 differs

Karl-WE:
I have changed the doc, in case this is a mistake. Otherwise please comment and close PR.

### DIFF
--- a/docs/connect/odbc/windows/system-requirements-installation-and-driver-files.md
+++ b/docs/connect/odbc/windows/system-requirements-installation-and-driver-files.md
@@ -26,7 +26,7 @@ Compatibility indicates that a driver was tested for compatibility against exist
 |----|---|---|---|---|---|---|---|---|---|---|---|
 |17.6|Yes|Yes|Yes|Yes|Yes|Yes|Yes|Yes|   |   |   |
 |17.5|Yes|Yes|Yes|Yes|Yes|Yes|Yes|Yes|   |   |   |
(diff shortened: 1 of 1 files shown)


PRMerger7:
@Karl-WE : Thanks for your contribution! The author(s) have been notified to review your proposed change.

David-Engel:
@Karl-WE Thanks for the submission. But your edit has changed the SQL support table, not the operating system support table. Based on you comment, it sounds like you just confused the tables. Windows 7 and Windows 2008 R2 support is already identical in the OS table for version 17.4.

Feel free to re-open and update your PR if you meant to change something else.

#please-close

Karl-WE:
@David-Engel thank you for pointing this out. I think I have misread the table infact.
So 2008R2 is supported as OS with 17.4 but not the SQL DB of the same generation. It that correct? 
If so then the close is correct. Thank you for your reply.

David-Engel:
> So 2008R2 is supported as OS with 17.4 but not the SQL DB of the same generation. It that correct?

@Karl-WE Correct. Driver support for OS/database aligns with general/extended support for those products. [Support for SQL Server 2008 R2 ended July 9, 2019](https://support.microsoft.com/en-us/help/4456242/end-of-support-for-sql-server-2008-and-sql-server-2008-r2). ODBC Driver 17.4 released July 31, 2019. [Windows 2008 R2 support ended January 14, 2020](https://www.microsoft.com/en-us/cloud-platform/windows-server-2008).

PR #5710, opened 2020-11-12
Update distinctcount-mdx.md

gretchkosiy:
Duplicate entries for [Customer].[Customer Geography].[Country].&[Australia] and [Customer].[Customer Geography].[Country].&[United Kingdom] in example

### DIFF
--- a/docs/mdx/distinctcount-mdx.md
+++ b/docs/mdx/distinctcount-mdx.md
@@ -34,11 +34,11 @@ DistinctCount(Set_Expression)
   
  `WITH SET MySet AS`  
   
- `{[Customer].[Customer Geography].[Country].&[Australia],[Customer].[Customer Geography].[Country].&[Australia],`  
+ `{[Customer].[Customer Geography].[Country].&[Australia],`  
   
  `[Customer].[Customer Geography].[Country].&[Canada],[Customer].[Customer Geography].[Country].&[France],`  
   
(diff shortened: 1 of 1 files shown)


PRMerger10:
@gretchkosiy : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@gretchkosiy Thank you for this suggested change. @Minewiskan Owen, can you take a look at the suggestion. My only concern is that it looks like a closing `}` was removed but not readded at the end of the list when the duplicate was removed. Could one or both of you verify this? Thank you!

Minewiskan:
@gretchkosiy - Thanks for pointing this out. This likely happened during a filetype conversion. It appears to have affected more than just this DISTINCTCOUNT file. We're reviewing the whole reference and will update as soon as possible.

ktoliver:
@Minewiskan Would you like to keep the pull request open? Thanks.

Minewiskan:
@ktoliver - Yes. Let's keep open while I review other files. I'll likely resolve this as a separate PR, and then close this one.

ktoliver:
@Minewiskan Thanks.

Minewiskan:
@gretchkosiy - The expression example is correct. We also investigated this back in November 2020. The paragraph below the example was added at that time to provide additional clarity. That said, the _formatting_ of the expression block is still incorrect. We will address that as part of a separate work item. The formatting change will not affect the expression itself.
#please-close

PR #5714, opened 2020-11-12
Update report-server-executionlog-and-the-executionlog3-view.md

gbritton1:
Need instructions for enabling logging/verbose logging for a system where Reporting Services runs on a different node from the database server.

### DIFF
--- a/docs/reporting-services/report-server/report-server-executionlog-and-the-executionlog3-view.md
+++ b/docs/reporting-services/report-server/report-server-executionlog-and-the-executionlog3-view.md
@@ -67,6 +67,9 @@ ms.author: maggies
 2.  Change the **ExecutionLogLevel** to **verbose**. This field is a text entry field and the two possible values are **verbose** and **normal**.  
   
 ##  <a name="bkmk_native"></a> Configuration Settings for a Native Mode Report Server  
+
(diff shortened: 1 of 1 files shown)


PRMerger15:
@gbritton1 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@gbritton1 Gerald, thanks for pointing this out. Typically, the public pull requests are for specific changes you want to contribute rather than general documentation feedback. I would suggest that you instead use the Feedback link at the bottom of the article to create a git issue with your feedback that we can track. With that said, @maggiesMSFT is on this public pull request and sees your feedback here and can look at fixing this in the future. Apologies for the missing content as you were using the docs! I am going to close this pull request, but you are welcome to open a git issue or add more comments to this thread. Thank you!

maggiesMSFT:
Hi, @gbritton1 -- Sorry for the delay! Here's the answer I got to your query:
"I’m not following.  The process is the same regardless if the db is in the same machine or a different one."
I hope that helps! If not, can you ask a more specific question?
cc @rothja

rothja:
Also, Maggie pointed out that we don't have the git issue feedback turned on for these articles, so sorry for suggesting that route. However, Maggie can engage on this thread and has noted your feedback here. Thanks!!

PR #5747, opened 2020-11-21
Update sys-dm-os-sys-info-transact-sql.md

Rob-S:
[container_type] and {container_type_description] columns are showing up in SQL Server 2017 (and later, I presume). I'm not sure what version(s) those "Applies to". I suspect the "container" is probably something like "Docker" or "Kubernetes" and _not_ the same sense in which a database is a "contained" database.

The "Applies to" and descriptions need to be completed before merging this PR.


PRMerger20:

markingmyname:
@Rob-S 

At this time, we cannot approve this change.

However, we are thankful for bringing this to our attention to look at refreshing the table.

We will investigate and possibly make edits to this in the future.

Thank you.

PR #5749, opened 2020-11-22
Update setup-steps-for-extensible-key-management-using-the-azure-key-…

dbaduck:
…vault.md

Changed wording in sql example that was adding the credential. It was the same wording as the above example dropping it. So I added better wording.


PRMerger14:
@dbaduck : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Rupp29 (approved):

PR #5804, opened 2020-12-14
(Encrypting Connections to SQL Server on Linux): Add missing permissions and restarting of server afterwards to examples

alph4numb3r:
1. Add mysql user to ssl-cert group so it has permission to open the /etc/ssl/private/ folder the private key is in. (Could use ACL instead, but this is a simpler setup to debug if necessary)
2. Add misssing sudo to cat command of config files
3. Restart the mssql-server service after configuring

### DIFF
--- a/docs/linux/sql-server-linux-encrypted-connections.md
+++ b/docs/linux/sql-server-linux-encrypted-connections.md
@@ -41,6 +41,7 @@ TLS is used to encrypt connections from a client application to [!INCLUDE[ssNoVe
 ```bash
 openssl req -x509 -nodes -newkey rsa:2048 -subj '/CN=mssql.contoso.com' -keyout mssql.key -out mssql.pem -days 365 
 sudo chown mssql:mssql mssql.pem mssql.key 
+sudo usermod -a -G ssl-cert mssql 
 sudo chmod 600 mssql.pem mssql.key   
 sudo mv mssql.pem /etc/ssl/certs/ 
 sudo mv mssql.key /etc/ssl/private/ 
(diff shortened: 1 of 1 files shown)


PRMerger13:
@alph4numb3r : Thanks for your contribution! The author(s) have been notified to review your proposed change.

vin-yu:
+ @tejasaks - can you help take a look at this? looks like a few admin permissions were added to the steps in order for this to work.

tejasaks:
@amvin87 to do the research who was looking into SSL very recently. At first glance changes look ok but need to be validated.

amvin87:
Yup this is another way to solve it but for now we already have added a note making the current sample workable, the note added is this. So no correction needed at this point. 
"Ensure that the folder/s you save your certs and Private keys are accessible by the mssql user/group and has permission set to 700 (drwx-----). You can create folders manually with permission set to 700 (drwx------) and owned by mssql user/group or set the permission to 755(drwxr-xr-x) and owned by other user but still accessible to mssql user group. Example, you could create a folder called 'sslcert' under the path '/var/opt/mssql/' and then save the certificate and the private key with permissions on the files set to 600 as show below."

ktoliver:
> Yup this is another way to solve it but for now we already have added a note making the current sample workable, the note added is this. So no correction needed at this point.

@amvin87 Can the pull request be closed, or is it still in progress? Thanks.

alph4numb3r:
> Yup this is another way to solve it but for now we already have added a note making the current sample workable, the note added is this. So no correction needed at this point.
> "Ensure that the folder/s you save your certs and Private keys are accessible by the mssql user/group and has permission set to 700 (drwx-----). You can create folders manually with permission set to 700 (drwx------) and owned by mssql user/group or set the permission to 755(drwxr-xr-x) and owned by other user but still accessible to mssql user group. Example, you could create a folder called 'sslcert' under the path '/var/opt/mssql/' and then save the certificate and the private key with permissions on the files set to 600 as show below."

What about the other two changes in this pull request?

tejasaks:
@amvin87 to look into it.

amvin87:
yes the other two changes are correct. Let me update the document right away. 
1) Adding the sudo command to read the mssql.conf and 
2) Starting the SQL Server service at the end

Thank you for the feedback @alph4numb3r

amvin87:
@VanMSFT created this internal [pull](https://github.com/MicrosoftDocs/sql-docs-pr/pull/19489)) request for this please merge that with the live and close this PR.

ktoliver:
#please-close

PR #5805, opened 2020-12-14
Sqlbindcol strlen or ind ptr

mversic:
Convention is that pointer argument names end in `Ptr`


PRMerger19:
@mversic : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #5838, opened 2020-12-22
Update sys-databases-transact-sql.md

Rob-S:
is_memory_optimized_elevate_to_snapshot_on applies to SQL Server 2014 and later, the same as the two columns for containment before it.


PRMerger13:

markingmyname:
@Rob-S, thank you for your contribution.

We have moved our 2014 content to another set.

As such, we do not mention 2014 in this new docset.

However, we still have that content live.

[Previsou versions](https://docs.microsoft.com/en-us/sql/sql-server/previous-versions-sql-server)

PR #5846, opened 2020-12-24
Update create-database-scoped-credential-transact-sql.md

NandanHegde15:
Added an additional scenario of Secret length restriction in Azure Synapse

### DIFF
--- a/docs/t-sql/statements/create-database-scoped-credential-transact-sql.md
+++ b/docs/t-sql/statements/create-database-scoped-credential-transact-sql.md
@@ -58,7 +58,7 @@ Specifies the name of the account to be used when connecting outside the server.
 > WITH IDENTITY is not required if the container in Azure Blob storage is enabled for anonymous access. For an example querying Azure Blob storage, see [Importing into a table from a file stored on Azure Blob storage](../functions/openrowset-transact-sql.md#j-importing-into-a-table-from-a-file-stored-on-azure-blob-storage).
 
 SECRET **='**_secret_**'**
-Specifies the secret required for outgoing authentication. `SECRET` is required to import a file from Azure Blob storage. To load from Azure Blob storage into Azure Synapse Analytics or Parallel Data Warehouse, the Secret must be the Azure Storage Key.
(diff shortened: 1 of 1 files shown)


PRMerger16:
@NandanHegde15 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

VanMSFT:
Hi @NandanHegde15 - I apologize for the delay.

@julieMSFT - Hi Julie, is this something you can take a look at for Azure Synapse?

julieMSFT:
@nanditavalsan Please review this update.

nanditavalsan:
Azure Synapse Analytics and other Azure SQL services have the same secret length for database scoped credentials. Please re-submit your change without specifying the service name if you believe that additional details will be useful. The new PR will be directed to the correct team.
#please-close

NandanHegde15:
> Azure Synapse Analytics and other Azure SQL services have the same secret length for database scoped credentials. Please re-submit your change without specifying the service name if you believe that additional details will be useful. The new PR will be directed to the correct team.
> #please-close

Hey @nanditavalsan,
The issue is only for Azure SQL DW and not for Azure SQL database, hence had explicitly called out.
We were able to create a database scoped credential via SAS value in Azure SQL database but the same SAS value when we tried in synapse it failed with the restriction error.

nanditavalsan:
Are you attempting to create a database scoped credential for use with Polybase or load data from Azure Storage? This will not work for SAS as noted here in the document. https://docs.microsoft.com/en-us/sql/t-sql/statements/create-database-scoped-credential-transact-sql?view=sql-server-ver15#b-creating-a-database-scoped-credential-for-a-shared-access-signature.
The sentence ahead of the change you are proposing states that you must use an Azure Storage key, "To load from Azure Blob storage into Azure Synapse Analytics or Parallel Data Warehouse, the Secret must be the Azure Storage Key.".

NandanHegde15:
Hey @nanditavalsan ,
Yes, we had seen that Synapse doesn't support use of SAS and we were testing out what all properties are different between SQL database and synapse and there we found that synapse has a length restriction whereas Azure SQL database doesn't.
So I added that note in this PR highlighting Synapse

PR #5848, opened 2020-12-26
Update sys-dm-io-virtual-file-stats-transact-sql.md to reflect that it's a function, not a view

BrentOzar:
This is a dynamic management function, not a view, because it requires parameters.

You can run a view by just saying SELECT * FROM myview, but that doesn't work here with this dynamic management function.


PRMerger7:
@BrentOzar : Thanks for your contribution! The author(s) have been notified to review your proposed change.

markingmyname:
@BrentOzar, thanks for your contribution.

approved.

#sign-off

PR #5863, opened 2021-01-04
Make it more clear that 'tedious' is not just a word

richardpark-msft:
Found this page while working on a test project and I was confused. I got the joke after reading the entire thing but having some formatting in the file would help a bit to call out that this is a term (ie, package name) rather than just being...the word tedious. :)


PRMerger19:
@richardpark-msft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

David-Engel (approved):

David-Engel:
#sign-off

PR #5965, opened 2021-01-25
Typo "Azure Blob storage service"→"Azure Blob Storage Service"

hyoshioka0128:
https://docs.microsoft.com/en-us/sql/t-sql/statements/create-credential-transact-sql?view=sql-server-ver15


PRMerger19:

VanMSFT:
Hi @hyoshioka0128 - Thanks for your feedback! However, Azure Blob storage service is defined correctly here. `storage service` should be lower case. Please let us know if there are any concerns.

PR #5966, opened 2021-01-25
Typo "Azure Blob storage service"→"Azure Blob Storage Service"

hyoshioka0128:
https://docs.microsoft.com/en-us/sql/relational-databases/system-functions/sys-fn-db-backup-file-snapshots-transact-sql?view=sql-server-ver15


PRMerger6:

ktoliver:
@hyoshioka0128 - The updated text refers to the Azure service Azure Blob Storage. The initial capitalization in the service name "Azure Blob Storage" looks good; thanks. The word _service_ isn't part of the service name, so it should be lowercase; I've made the edit and will merge.

PR #5976, opened 2021-01-27
Updating the CASE topic for Synapse SQL

TwoUnder:
Synapse SQL supports both case options.


PRMerger9:
@TwoUnder : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6030, opened 2021-02-01
SELECT...OVER - ORDER BY: Typographical Error

LarnuUK:
*"and in ROWS/RANGE is not specified"* corrected to *"and a ROWS/RANGE is not specified"*.


PRMerger19:

ktoliver:
Looks good; merging. Thanks for the fix @LarnuUK.

PR #6081, opened 2021-02-12
Update step-3-proof-of-concept-connecting-to-sql-using-pyodbc.md

OrrinEdenfield:
I tried to use the INSERT code in my Python script and kept running into an issue where the record DOES get inserted into my table but I keep getting this error with pyodbc:

    row = cursor.fetchone()
pyodbc.ProgrammingError: No results.  Previous SQL was not a query.

In the case of this insert, I really don't expect anything to be returned - I just want to ensure the data is inserted into my table (which I've verified and can confirm). So I used a slightly different technique where I removed the cursor.fetchone() and instead have pyodbc return something that I can handle easier with my script. Using the code above I get a "0" returned which is good enough (it's not erroring anymore).

### DIFF
--- a/docs/connect/python/pyodbc/step-3-proof-of-concept-connecting-to-sql-using-pyodbc.md
+++ b/docs/connect/python/pyodbc/step-3-proof-of-concept-connecting-to-sql-using-pyodbc.md
@@ -61,16 +61,12 @@ In this example, you see how to run an [INSERT](../../../t-sql/statements/insert
   
 ```python
 #Sample insert query
-cursor.execute("""
+count = cursor.execute("""
 INSERT INTO SalesLT.Product (Name, ProductNumber, StandardCost, ListPrice, SellStartDate) 
 VALUES (?,?,?,?,?)""",
 'SQL Server Express New 20', 'SQLEXPRESS New 20', 0, 0, CURRENT_TIMESTAMP) 
 cnxn.commit()
-row = cursor.fetchone()
-
-while row: 
-    print('Inserted Product key is ' + str(row[0]))
-    row = cursor.fetchone()
+print(count)
 ```  
 
 ## Azure Active Directory and the connection string


PRMerger7:
@OrrinEdenfield : Thanks for your contribution! The author(s) have been notified to review your proposed change.

David-Engel:
@v-makouz Can you validate the changes in the PR? Thanks!

v-makouz:
When I test this change I don't get a count of 0, but "<pyodbc.Cursor object at SOMEADDRESS>"

docs/connect/python/pyodbc/step-3-proof-of-concept-connecting-to-sql-using-pyodbc.md
```diff
@@ -61,16 +61,12 @@ In this example, you see how to run an [INSERT](../../../t-sql/statements/insert
   
 ```python
 #Sample insert query
-cursor.execute("""
+count = cursor.execute("""
 INSERT INTO SalesLT.Product (Name, ProductNumber, StandardCost, ListPrice, SellStartDate) 
 VALUES (?,?,?,?,?)""",
 'SQL Server Express New 20', 'SQLEXPRESS New 20', 0, 0, CURRENT_TIMESTAMP)
```
David-Engel, line 67: This will accomplish what you want.
```suggestion
'SQL Server Express New 20', 'SQLEXPRESS New 20', 0, 0, CURRENT_TIMESTAMP).rowcount 
```

docs/connect/python/pyodbc/step-3-proof-of-concept-connecting-to-sql-using-pyodbc.md
```diff
@@ -61,16 +61,12 @@ In this example, you see how to run an [INSERT](../../../t-sql/statements/insert
...
 INSERT INTO SalesLT.Product (Name, ProductNumber, StandardCost, ListPrice, SellStartDate) 
 VALUES (?,?,?,?,?)""",
 'SQL Server Express New 20', 'SQLEXPRESS New 20', 0, 0, CURRENT_TIMESTAMP) 
 cnxn.commit()
-row = cursor.fetchone()
-
-while row: 
-    print('Inserted Product key is ' + str(row[0]))
-    row = cursor.fetchone()
+print(count)
```
David-Engel, line 69: And to clean things up a bit:
```suggestion
print('Rows inserted: ' + str(count))
```

David-Engel (commented):
It looks like the original sample probably also included a function that was trying to fetch the last PK ID that was created by the insert. But it was likely removed since that function is not reliable. We can go with your suggestion if you apply my suggestions to successfully return the inserted row count.

David-Engel:
This change has been made via an internal PR. Thanks for the suggestion!

#please-close

PR #6097, opened 2021-02-17
Formatting changes

KoprowskiT:
For having the same formatting as for documentation about the master database, I suggest:
- remove empty rows between row 63 and 71
- adding dash "-" for lines 63 71  for part Related Content

It will make page formatted (for part ##Related Content) in the same way (without empty lines and with dashes) as on master-database.md

### DIFF
--- a/docs/relational-databases/databases/resource-database.md
+++ b/docs/relational-databases/databases/resource-database.md
@@ -60,14 +60,9 @@ GO
 ```  
   
 ## Related Content  
- [System Databases](../../relational-databases/databases/system-databases.md)  
-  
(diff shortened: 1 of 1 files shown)


PRMerger14:
@KoprowskiT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rothja:
@KoprowskiT Thanks for this suggestion. However, we use this non-bulleted style quite a bit in docs, and it can be tricky to remove the spaces and get the markdown formatting right. So I won't be approving this change. Thanks for understanding.

PR #6140, opened 2021-03-02
Update sys-dm-hadr-availability-group-states-transact-sql.md

Rob-S:
In "Not the primary replica OR unable to communicate with the WSFC failover cluster." the "OR" means if either condition is true. So when it's not the primary replica, I would expect to see NULL, However, I'm seeing the primary replica name when selecting the column on the secondary replica. In this case, since it _can_ communicate with the WSFC, it must be getting the primary replica name from the cluster. Therefore I believe it's NULL when 1) it's not the primary replica AND 2) it can't get the primary replica name from the WSFC failover cluster because it's unable to communicate with it.

### DIFF
--- a/docs/relational-databases/system-dynamic-management-views/sys-dm-hadr-availability-group-states-transact-sql.md
+++ b/docs/relational-databases/system-dynamic-management-views/sys-dm-hadr-availability-group-states-transact-sql.md
@@ -32,7 +32,7 @@ ms.author: wiassaf
 |Column name|Data type|Description|  
 |-----------------|---------------|-----------------|  
 |**group_id**|**uniqueidentifier**|Unique identifier of the availability group.|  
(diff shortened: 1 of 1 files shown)


PRMerger13:

WilliamDAssafMSFT:
Thanks @Rob-S for contacting Microsoft with feedback about our product documentation. Depending on the complexity, it could take from a few business days to a few weeks to look further into your question, issue, or suggestion. We'll contact you if we need further information, and you'll receive an email message each time we "@mention" you in a comment.

Allow a day or more for the article's author to review and approve your change or offer an alternative solution.

By opening a pull request, you're helping the author add your approved change directly to the documentation. After the updated article is published, GitHub lists you as an article contributor.

cawrites:
#sign-off

PRMerger12:
@cawrites: I'm sorry - only the author of this article, @WilliamDAssafMSFT, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

WilliamDAssafMSFT:
#sign-off

PR #6176, opened 2021-03-12
Include the string representation of root

mrsegen:
The intention here is to allow people new to the HierarchyId concept to be more rooted 🥁 by seeing the concrete human-friendly representation of the root.


PRMerger16:
@mrsegen : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MikeRayMSFT:
@mrsegen thank you for this submission. 

We have reviewed the submission and decided not to include it in the docs at this time. It is not in the scope of the current article. 

The article for the ToString method is in the same table of contents location.

PR #6181, opened 2021-03-15
Improved memory-related wording

greglow-sdu:
The current wording is incorrect. It suggests that as soon as you start using sp_xml_preparedocument that you lose 1/8 of SQL Server's memory. It's easy to show that this is not true. That's a limit. In addition, a suggestion is added that the nodes() method might be a better alternative in many cases, and avoid this issue in the first place. A reference to it was also added.


PRMerger12:
@greglow-sdu : Thanks for your contribution! The author(s) have been notified to review your proposed change.

markingmyname (approved):
changes approved

markingmyname:
#sign-off

PR #6199, opened 2021-03-18
Update missing verb in "when you a database"

zo0o0ot:
I found six instances of a missing verb in the documentation.  I assume that the verb is "publish to", but I'm not sure, since that's the reason I was trying to go through these docs.


PRMerger9:
@zo0o0ot : Thanks for your contribution! The author(s) have been notified to review your proposed change.

dzsquared (approved):

dzsquared:
Thank you for catching this and fixing!

#sign-off

zo0o0ot:
So I was right that the verb that was missing was "publish to"? Good to hear.  I was still trying to understand the subject matter when I ran into the documentation error.

PR #6224, opened 2021-03-26
Incorrect key name

stimms:
Looks like an old value was used for the column encryption key value. The one created in the previous statement is called `CEK1`


PRMerger14:
@stimms : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jaszymas:
#sign-off

PR #6229, opened 2021-03-27
Add clarification for TF 1448

taroarai:
Our customer confused that "synchronous secondary replicas" as a setting, but as tested that SQL Server service for "synchronous secondary replicas" was stopped, it become asynchronous, and log reader will advance with TF 1448.
This synchronous replica become asynchronous is documented in the Note section for following Docs, so adding link to this Docs should help understand this TF 1448 better.

Differences between availability modes for an Always On availability group
https://docs.microsoft.com/en-us/sql/database-engine/availability-groups/windows/availability-modes-always-on-availability-groups?view=sql-server-ver15


PRMerger9:

cawrites:
@taroarai thank you for supporting and providing this update for our customer.

PR #6237, opened 2021-03-30
Update alter-database-transact-sql-set-options.md

kodaksmile:
Note there are many options that are not supported in SQL MI per this link:  https://docs.microsoft.com/en-us/azure/azure-sql/managed-instance/transact-sql-tsql-differences-sql-server#alter-database-statement  including RESTRICTED_USER and READ_ONLY.  There are several others so this document should be updated to accurately reflect those options we do not support.  Please consider reviewing the full list of what is not supported and including that here to reduce confusion for our customers.

### DIFF
--- a/docs/t-sql/statements/alter-database-transact-sql-set-options.md
+++ b/docs/t-sql/statements/alter-database-transact-sql-set-options.md
@@ -178,7 +178,7 @@ SET
     { READ_ONLY | READ_WRITE }
 
 <db_user_access_option> ::=
-    { SINGLE_USER | RESTRICTED_USER | MULTI_USER }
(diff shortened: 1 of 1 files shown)


PRMerger12:
@kodaksmile : Thanks for your contribution! The author(s) have been notified to review your proposed change.

WilliamDAssafMSFT:
Thanks @kodaksmile for contacting Microsoft with feedback about our product documentation. Depending on the complexity, it could take from a few business days to a few weeks to look further into your suggestion. We'll contact you if we need further information, and you'll receive an email message each time we "@mention" you in a comment.

PR #6246, opened 2021-04-01
Update row-level-security with lookup table example

gloxton:
I have had the same question from a number of my clients about row level security – “based on these examples, do we need the store security predicate on the table being queried, or can we use a lookup?” This new example provides a code example and shows how easy it is to perform the security lookup using a lookup table.

### DIFF
--- a/docs/relational-databases/security/row-level-security.md
+++ b/docs/relational-databases/security/row-level-security.md
@@ -530,6 +530,149 @@ DROP FUNCTION Security.fn_securitypredicate;
 DROP SCHEMA Security;
 ```
 
+### <a name="Lookup"></a> D. Scenario for using a lookup table for the security predicate
+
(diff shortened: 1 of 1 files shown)


PRMerger16:
@gloxton : Thanks for your contribution! The author(s) have been notified to review your proposed change.

VanMSFT:
Hi @gloxton! Thanks for these additions! We appreciate your help with adding examples.

@AndreasWolter - Hey Andreas, can you please help review these additions and see if they are appropriate to add to the documentation? https://docs.microsoft.com/en-us/sql/relational-databases/security/row-level-security

#reassign:AndreasWolter

AndreasWolter:
Hi @gloxton Thanks for the additional example. it certainly is useful. Just one minor thing: Could you for the user table please use a different schema than dbo? We are trying to educate users to avoid the dbo schema unless there are strong reasons to use it. Thank you

PRMerger20:
@gloxton : Thanks for your contribution! The author(s) have been notified to review your proposed change.

VanMSFT:
#sign-off

VanMSFT:
Thanks @gloxton and @AndreasWolter!

PR #6265, opened 2021-04-07
Update sql-server-linux-editions-and-components-2019.md

DanielAdeniji:
On the Microsoft Windows platform, successful and failed logins are auditable.

Whereas, on the Linux platform, it does not appear that logins are auditable.

### DIFF
--- a/docs/linux/sql-server-linux-editions-and-components-2019.md
+++ b/docs/linux/sql-server-linux-editions-and-components-2019.md
@@ -227,6 +227,7 @@ The following features and services are not available for SQL Server 2019 on Lin
 | **Security** | Extensible Key Management |
 | &nbsp; | AD Authentication for Linked Servers |
 | &nbsp; | AD Authentication for Availability Group (AG) Endpoints |
+| &nbsp; | Login Auditing |
 | **Services** | SQL Server Browser |
 | &nbsp; | SQL Server R services<sup>1</sup> |
 | &nbsp; | StreamInsight |


PRMerger13:
@DanielAdeniji : Thanks for your contribution! The author(s) have been notified to review your proposed change.

VanMSFT:
Hi @DanielAdeniji - Thanks for this submission. Can you please elaborate on what you mean by auditable? What are you using to audit?

https://docs.microsoft.com/sql/linux/sql-server-linux-editions-and-components-2019

VanMSFT:
Thanks for the clarification, @DanielAdeniji.

@amvin87 @tejasaks - Hey Amit/Tejas, what do you think of this change? I suppose there are other ways to audit logins, so I wouldn't call this an unsupported feature, but let's hear the Product Team's input.

DanielAdeniji:
Van To:-

My personal Take is that the availability of other ways does mean the option is available.

The other ways mentioned are a bit more laborious and may not be available in an increasingly cloud-first world.

The other means includes:-

  1.  Server Audit
     *   One has to create a Server Audit and Server Audit Specification
        *   Server Audit ( Access to File System - Read directories, create and write files )
        *   Server Audit Specification

The original pathway appears to be more easily accessible via SSMS.

It was a simple radio button and that we can access and choose which login attempts ( failed and successful ) that we want included in the default trace.

A lot of security conscious administrators will use the option to track failed login attempts.

It is truly useful when troubleshooting failed login attempts.

On Linux, if one does not see failed login attempts in the sql server error log, one might assume they are not occurring.

One will be the worse for it.

My 2 cents,

Daniel

tejasaks:
@amvin87 to provide final word on this. 

If we want to clarify about the "default" login failed/succeeded messages in errorlog, that's fine. We can add that statement. But that statement cannot state that login auditing is not available. We have to be very specific that this particular "dated" way of checking for failed login is not available in Linux for today. The preference for future would obviously be to use server auditing which is far more scalable and usable compared to writing, reading and parsing errorlog files. In general use of errorlog file as login audits isn't the primary or ideal use case. Login failed messages in errorlog are there to help with more login infrastructure related issues (such as kerberos auth not working etc.).  I am not sure what part cloud plays here since server audit should be available regardless of SQL on Linux/containers being deployed on-prem or cloud.

DanielAdeniji:
Tejas:-

Stumbled Upon:-


  1.  I was surprised to notice the change.
  2.  In the midst of troubleshooting another functionality, I tried to see if the issue was sql server login related
  3.  Agreeably, the error reported to the client is sparse ( security through obfuscation, if you will )
  4.  One can access is SQL Server Error Log and Event Viewer for more depth
  5.  To me the depth is quite good

On Linux:-


  1.  SQL Server Management Studio ( SSMS )
     *   Is aware of this issue
        *   SQL Server Instance \ Property \ Server \ Security exposed
        *   Server Policy
           *   MS Windows => Supports logging to event viewer ( File System,  Event Viewer - Security & Application )
           *   Linux => Supports logging to event viewer ( File System,  Event Viewer - Security & Application )

My hope is that the dissimilarities between the platforms ( OS & Linux ) will be surfaced more readily.

And, those allow better reasoning by system architects.

Daniel


amvin87:
Thank you for the feedback @DanielAdeniji. I agree with @tejasaks here. Couple of more points:
1)  To the point "On Linux, if one does not see failed login attempts in the sql server error log, one might assume they are not occurring." - You can see that today by default for SQL on Linux as well. The failed login attempts do get recorded in the SQL Server error logs like it is for SQL on windows. There is no change in that behavior. 
The only difference is that on windows you can select to log both failed and successful logins, in Linux only failed logins get logged in error log. 
2) But just not having that option does not mean that we do not support login audits, you can do that using Server audits ( which is the preferred method of auditing) and another option is xevents. And both, these options are supported for SQL on Linux and Windows. 

So, in my opinion in this document we can state that to audit both successful and failed logins please use server audits.

VanMSFT:
@DanielAdeniji @tejasaks @amvin87 - Thanks for the input!

Daniel, we appreciate your feedback. For the document, it wouldn't be appropriate to add that line into the unsupported features. I'm going to deny this for now. Basic and fine grained auditing is listed as a feature here, and should include login auditing: https://docs.microsoft.com/en-us/sql/linux/sql-server-linux-editions-and-components-2019?view=sql-server-ver15#RDBMSS

As Tejas and Amit noted, there are various ways to get the auditing that you need, and it is not prudent to have this granular level of detail in our doc for this particular issue. Again, thank you for your feedback and suggestions to make our docs better.

PR #6289, opened 2021-04-13
Update sql-server-linux-active-directory-authentication.md

krishnamsft20:
We should not give "mssql" user as an example of creating new user in AD. This seems to be creating confusion between mssql local user on Linux vs mssql user in AD. Some customers do not seem to read the fine print and simply create mssql user in AD. 
I think replacing mssql with sqlsvc makes it clear to customer that this is service account we are talking about.


PRMerger12:
@krishnamsft20 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

VanMSFT:
Thanks @krishnamsft20!

Hey @amvin87 - Can you please help with this?

#reassign:amvin87

VanMSFT:
Hey @amvin87 - Following up on this.

amvin87 (approved):

amvin87:
@VanMSFT  yes this looks good and hence approved it.

PR #6306, opened 2021-04-15
Added an example with NULLs and Improved examples

greglow-sdu:
There was no example that showed the effect of NULL. I've added one and improved the final example. In particular, ensured that the INSERT statement uses a column list. (Having INSERT statements without column lists isn't good practice).  William Assaf mentioned that these should use AdventureWorksLT so changed that too.

### DIFF
--- a/docs/t-sql/functions/logical-functions-least-transact-sql.md
+++ b/docs/t-sql/functions/logical-functions-least-transact-sql.md
@@ -88,7 +88,27 @@ LeastVal
 (1 rows affected)  
 ```  
 
-### B. Simple example with character types
+### B. Simple example with NULL values
+
+Any NULL values are ignored. 
+
+ 
+```sql 
+SELECT LEAST ( 7.0000, NULL, 3, NULL, 9 ) AS LeastVal; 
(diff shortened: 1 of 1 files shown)


PRMerger15:
@greglow-sdu : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
@jmsteen Could you take a look at the proposed update to your article?

Thanks!

greglow-sdu:
Looks like this review never happened. @ktoliver

ktoliver:
@rothja @WilliamDAssafMSFT @jmsteen Could you review or reassign this pull request? Thanks.

WilliamDAssafMSFT:
Hi @greglow-sdu, sorry for the delay on this one. 
Switching to `AdventureWorksLT` breaks this sample, I like to use LT as much as anyone in the samples but it is limited. I'll take a look.

WilliamDAssafMSFT:
Addressed in https://github.com/MicrosoftDocs/sql-docs-pr/pull/21432

PR #6334, opened 2021-04-23
Don't recommend DATALENGTH as replacement for LEN

bdukes:
`LEN` counts the number of characters, while `DATALENGTH` counts the number of bytes. At the least, this requires logic bound to the type of the string (e.g. `varchar` vs. `nvarchar`), but also doesn't account for surrogate pairs that are counted as one character but multiple "characters" worth of bytes. `DATALENGTH` cannot be used to get the correct number of characters in a string containing surrogate pairs.


PRMerger9:
@bdukes : Thanks for your contribution! The author(s) have been notified to review your proposed change.

pmasl:
For non-Unicode encoding DATALENGTH is a suitable replacement, as it is for Unicode in the 1st 128 character range. The appropriate note that if processing a unicode string, DATALENGTH will return a number that may not be equal to the number of characters is there.

#please-close

PR #6389, opened 2021-05-10
Removing a confusing sentence which doesn't help the customer

monamaki:
Each specific external language can have one or more external language file spec. For one external language, there cannot be multiple external language files from the same platform (e.g. we cannot have TWO ext lang files BOTH for WINDOWS for a specific ext lang). This was already mentioned under file_spec in this doc.

Removing the confusing sentence: Platform and language name will be the unique key per external language.


PRMerger12:
@monamaki : Thanks for your contribution! The author(s) have been notified to review your proposed change.

dphansen (approved):
Thank you for your contribution. I agree with the change.

PR #6395, opened 2021-05-11
Updated restrictions for servername unqiueness

SeanGallardy:
Availability Replicas must have unique names in both Windows and SQL Server. Updating the documentation under AG restrictions to explicitly call this out. Availability Replicas and Groups will have issues if all instances are renamed and do not match their Windows name. For example, using sp_dropserver and sp_addserver to make all instances return the same @@SERVERNAME value.


PRMerger16:

kfarlee:
That works for me.

cawrites (approved):

PR #6426, opened 2021-05-20
Update create-database-scoped-credential-transact-sql.md

stevepmo:
Hi If you are using a database scoped credential (or possibly more than 1) the name of the credential needs to be the path to the blob storage
If you just use a random identifier it will not work - took me way to long to find out what was wrong - so updating the example as proposed in BoL would be very helpful

[https://MyStorageAccount.blob.core.windows.net/MyStorageContainerSqlEvents]

### DIFF
--- a/docs/t-sql/statements/create-database-scoped-credential-transact-sql.md
+++ b/docs/t-sql/statements/create-database-scoped-credential-transact-sql.md
@@ -114,7 +114,7 @@ The following example creates a database scoped credential that can be used to c
 CREATE MASTER KEY ENCRYPTION BY PASSWORD='<EnterStrongPasswordHere>';
 
 -- Create a database scoped credential.
-CREATE DATABASE SCOPED CREDENTIAL MyCredentials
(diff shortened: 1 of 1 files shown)


PRMerger16:
@stevepmo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
Note that publishing in this repository is suspended until Tuesday, May 25. Merging will resume Tuesday. Authors can still review and sign off on PRs in the interim. Thanks.

VanMSFT:
Hi @stevepmo - Thanks for your contribution. Can you please elaborate on this change as we're getting different results. The *credential_name* does not need to be the URL of the Azure Blob storage when creating the Database Scoped Credential for an external data source. Please see our example: https://docs.microsoft.com/en-us/sql/t-sql/statements/create-external-data-source-transact-sql?view=sql-server-ver15&tabs=dedicated#i-create-an-external-data-source-for-bulk-operations-retrieving-data-from-azure-storage

stevepmo:
Hi

yes I think I misinterpreted the documentation / mainly followed an incomplete blog post

my apologies - you should close the issue

/Stephen

VanMSFT:
Thanks for follow-up, @stevepmo!

PR #6464, opened 2021-06-04
Update with more Support Info vs Linux

EricBuMSFT:
We had a case come into CSS about this feature on a Linux OS (SQL Linux), and since this article didn't mention Linux directly it was hard to find that it was not supported.


PRMerger6:
@EricBuMSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MikeRayMSFT:
#sign-off

PR #6470, opened 2021-06-08
Update alter-login-transact-sql.md

toastjoqo:
Remove below part. Because CHECK_POLICY is not supported in synapse  "Msg 40517, Level 16, State 1, Line 6 Keyword or statement option 'check_policy' is not supported in this version of SQL Server."

To unlock a login without changing the password, turn the check policy off and then on again.

```sql
ALTER LOGIN [Mary5] WITH CHECK_POLICY = OFF;
ALTER LOGIN [Mary5] WITH CHECK_POLICY = ON;
GO
```

### DIFF
--- a/docs/t-sql/statements/alter-login-transact-sql.md
+++ b/docs/t-sql/statements/alter-login-transact-sql.md
@@ -257,13 +257,6 @@ ALTER LOGIN [Mary5] WITH PASSWORD = '****' UNLOCK ;
 GO
 ```
 
-To unlock a login without changing the password, turn the check policy off and then on again.
-
-```sql
-ALTER LOGIN [Mary5] WITH CHECK_POLICY = OFF;
-ALTER LOGIN [Mary5] WITH CHECK_POLICY = ON;
-GO
-```
 
 ### G. Changing the password of a login using HASHED
 


PRMerger14:
@toastjoqo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

VanMSFT:
@WilliamDAssafMSFT and @julieMSFT to check on the Azure Synapse portion

WilliamDAssafMSFT:
Hi @toastjoqo! 
Thanks for bringing this to our attention. After confirming with the product team, the ALTER LOGIN doc correctly only shows the password and enable options when selecting Synapse. The second example in Section F should be deleted. However, your change is made in the SQL server region of the document, not the SQL DW region of the doc. So I have pulled this commit into [a different PR](https://github.com/MicrosoftDocs/sql-docs-pr/pull/19495), to preserve your contributor status, and modified the change to occur in the right moniker range, under ## Azure Synapse Analytics. This isn't always obvious - the doc has multiple versions for _similar_ content for different SQL products, and we needed to make this change down around line 900. Your suggestions help improve our documentation. Your commit has been merged and this article update is in flight. After the updated article is published, GitHub will list you as an article contributor. 

PR #6491, opened 2021-06-16
Fixed tiny error in path to cert

x509cert:
I made this fix, but also, on line 217, there's a ref to Azure SQL DB not supporting a file for the cert, that's the case for Synapse, too, right? I didn't make an edit to this effect.


PRMerger6:
@x509cert : Thanks for your contribution! The author(s) have been notified to review your proposed change.

VanMSFT:
Thanks @x509cert - I'm approving this. Thank you so much for your help with our documentation!

#sign-off

PR #6514, opened 2021-06-17
Clarification on permissions

taroarai:
I suspect it is "CONNECT SQL" as server level, not "CONNECT" as database level, right?
However, at prerequisites, you must be able to connect mean you mast have "CONNECT SQL," so is it something different?


PRMerger12:

cawrites:
@taroarai we appreciate your input.  As of SQL 2012 it is CONNECT SQL at the instance level, but prior to that it was just CONNECT.

PR #6552, opened 2021-06-29
IFI is only available for data files

StefDBA:
In the way it is written, clients believe that it applies also to log files, especially if they do not read the sentence to the end.
https://techcommunity.microsoft.com/t5/core-infrastructure-and-security/how-and-why-to-enable-instant-file-initialization/ba-p/370329


PRMerger12:
@StefDBA : Thanks for your contribution! The author(s) have been notified to review your proposed change.

WilliamDAssafMSFT:
#sign-off

PR #6557, opened 2021-06-30
Update configure-reporting-services-to-use-a-subject-alternative-name.md

aweber-ql:
When I followed the directions as is, I was able to get to my /ReportServer, but not /Reports.  The new step 5 solved this problem.  I could not get the rest of the list to automatically renumber, so I left it in case there is something easy I am missing.
I also added the tip for non-command line people with a warning about the hard return and extra pieces that don't need to be typed from step 9.


PRMerger17:
@aweber-ql : Thanks for your contribution! The author(s) have been notified to review your proposed change.

maggiesMSFT:
Hi, @aweber-ql -- thanks so much for your suggested edits! I ran it by the devs, and we came up with slightly different edits, based on yours. Should be live sometime today, July 1.
#please-close

PR #6560, opened 2021-07-01
Update create-xml-index-selective-xml-indexes.md

mpa1234567:
There is no need for namespaces here.


PRMerger18:
@mpa1234567 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

WilliamDAssafMSFT:
@mpa1234567 Can you clarify the suggestion that the XMLNAMESPACES syntax arguments are not valid for CREATE SELECTIVE XML INDEX?

PR #6561, opened 2021-07-01
Update xml-schemas-xml-type-system-catalog-views-transact-sql.md

mpa1234567:
There is no document for system view sys.selective_xml_index_namespaces


PRMerger9:
@mpa1234567 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6603, opened 2021-07-16
Adding_QS_cannot_disable

ahmedmahmoudMSFT:
Adding a note that customer cannot disable QS


PRMerger7:
@ahmedmahmoudMSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
@WilliamDAssafMSFT Could you review the proposed update? Avoiding "stacked" alerts (two alerts without content between them) is a preferred practice. Is there a way to take that into consideration if the information should be added to the article? Thanks.

WilliamDAssafMSFT:
Thanks @ahmedmahmoudMSFT for the suggestion! We already have this limitation and error message documented in the [appropriate place here in ALTER DATABASE](https://docs.microsoft.com/en-us/sql/t-sql/statements/alter-database-transact-sql-set-options#query-store). I believe that suffices and this doesn't need to be repeated in a performance article.

PR #6607, opened 2021-07-17
Fix int bitness to 32

slepmog:
See https://stackoverflow.com/questions/68420639/bitwise-not-example-on-sql-server


PRMerger8:
@slepmog : Thanks for your contribution! The author(s) have been notified to review your proposed change.

cawrites (approved):

cawrites:
@slepmog appreciate your input on this. You made a great catch here and I am sure it will benefit others as well. Thank you.

PR #6609, opened 2021-07-19
Update sql-server-linux-encrypted-connections.md

MitchellSternke:
Add documentation related to recent Linux distribution behavior with regards to OpenSSL certificates.

Users may encounter this issue out-of-the box if they are using their own custom certificates. This makes this issue searchable and documents the solution.

Mentioning @tejasaks / @amvin87 for review.


PRMerger17:
@MitchellSternke : Thanks for your contribution! The author(s) have been notified to review your proposed change.

amvin87 (approved):

amvin87:
#sign-off 
@VanMSFT could you please help merge this change to the document.

ktoliver:
Hi @amvin87 Before we merge, "note" should be removed from the new H2 heading. We can either change the current H2 heading to "Ubuntu 20.04 and other recent Linux distribution releases," or use that same text string, but make the heading an H3, to demote it in accordance with the meaning of "note." Which do you prefer, or is there another option you'd like to use? 

Thanks.

amvin87:
Thanks @ktoliver for the suggestion, let us please go with the latter option, which is : "use that same text string, but make the heading an H3, to demote it in accordance with the meaning of "note""

PRMerger10:

ktoliver:
Thanks @amvin87; we'll merge this now.

PR #6620, opened 2021-07-21
Highlight the fact that there is no undo for this action

jesulink2514:
Marking the limitations of MEMORY OPTIMIZED TABLES with a Caution label because there is no undo for this.


PRMerger8:
@jesulink2514 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

markingmyname:
#sign-off

Issue #6634, opened 2021-07-23
Why promote a scan when a seek would be far better?


LarnuUK:
In the example it seems odd that the `ON` clause contains the following expressions:
```
      AND DATEPART(mm,OrderDate)= 3
      AND DATEPART(yy,OrderDate) = 2002
```
Though true in the comment at the start of the query, that the query will use the indexed view, it *scans* the index. If proper date boundaries are used and not `DATEPART`, then a **seek** can be used; far more performant.

I don't really understand why this document is teaching a bad habit here. When referencing columns in the `WHERE` or `ON` applying functions on them should be avoided to maintain SARGability. The above cause would be far better as the following, which maintains SARGability and promotes good query writing:
```
      AND o.OrderDate >= '20020301'
      AND o.OrderDate < '20020401'
```

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0a452207-5ac2-1a41-c5b1-3baf085a79b4
* Version Independent ID: a590e66c-ed9f-4d80-1511-1864128d2e8e
* Content: [Create Indexed Views - SQL Server](https://docs.microsoft.com/en-us/sql/relational-databases/views/create-indexed-views?view=sql-server-ver15)
* Content Source: [docs/relational-databases/views/create-indexed-views.md](https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/relational-databases/views/create-indexed-views.md)
* Product: **sql**
* Technology: **table-view-index**
* GitHub Login: @WilliamDAssafMSFT
* Microsoft Alias: **wiassaf**

WilliamDAssafMSFT:
Thanks @LarnuUK for contacting Microsoft with feedback about our product documentation.

This is a good edit! Perhaps these ranges:

```
      AND o.OrderDate >= '20110101'
      AND o.OrderDate < '20120101'
````

So that data is returned in the latest versions of the AdventureWorks sample databases.

Would you like to open a PR to make the change yourself, and be recognized as a contributor? Let me know.

LarnuUK:
That's done for you, @WilliamDAssafMSFT . I did a small amount other "tidying" as well while I was there.

PR #6635, opened 2021-07-23
Clean up examples - SARGability & clause positions

LarnuUK:
Update the example to use proper date boundaries, `>=` and `<` logic, in the final example. This changes the query to use a seek on the example index, rather than a scan of it; which would be much more performant as the query is SARGable. 

Also changed the date ranges in both example queries, as proposed in the related [issue](https://github.com/MicrosoftDocs/sql-docs/issues/6634) so that it's relevant for more recent versions of the Adventure Works database. Though I prefer, personally, to use unambiguous date formats (`yyyyMMdd`, I have used `CONVERT` in the second query with an explicit style to be consistent with the prior query.

Finally, I move the clause `ProductID BETWEEN 700 and 800` to the `WHERE`. `ProductID` is in the table `SalesOrderDetail` so it doesn't make much sense being in the `ON` clause for the `JOIN` to the table `SalesOrderHeader`. Technically the date boundaries in the `ON` clause in the second example could be moved to the `WHERE` (not if it were a `LEFT JOIN`) but it's not as unintuitive in its current position.

### DIFF
--- a/docs/relational-databases/views/create-indexed-views.md
+++ b/docs/relational-databases/views/create-indexed-views.md
@@ -2,7 +2,7 @@
 description: "Create Indexed Views"
 title: "Create Indexed Views | Microsoft Docs"
 ms.custom: ""
-ms.date: "11/19/2018"
+ms.date: "07/23/2021"
 ms.prod: sql
 ms.prod_service: "database-engine, sql-database, synapse-analytics, pdw"
 ms.reviewer: ""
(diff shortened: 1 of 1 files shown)


PRMerger9:

PRMerger8:
@WilliamDAssafMSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

WilliamDAssafMSFT:
Good edit @LarnuUK and thanks for contributing! Your commit will be merged and this article update is in flight. After the updated article is published, GitHub will list you as an article contributor. 

#sign-off

PR #6638, opened 2021-07-25
Update sys-dm-db-partition-stats-transact-sql.md

alberto-morillo:
I just want to add a useful example applicable to Azure SQL Database.

### DIFF
--- a/docs/relational-databases/system-dynamic-management-views/sys-dm-db-partition-stats-transact-sql.md
+++ b/docs/relational-databases/system-dynamic-management-views/sys-dm-db-partition-stats-transact-sql.md
@@ -98,6 +98,12 @@ FROM sys.dm_db_partition_stats
 WHERE object_id=OBJECT_ID('HumanResources.Employee')    AND (index_id=0 or index_id=1);  
 GO  
 ```  
+### D. Get the reserved space of an Azure SQL Database  
+ The following example uses the sys.dm_db_partition_stats DMV to get the current size of an Azure SQL Database.  
+  
+```sql  
+SELECT SUM(reserved_page_count) * 8.0 / 1024 AS 'Database size [MB]' FROM sys.dm_db_partition_stats
+```  
   
 ## See Also  
  [Dynamic Management Views and Functions &#40;Transact-SQL&#41;](~/relational-databases/system-dynamic-management-views/system-dynamic-management-views.md)   


PRMerger9:
@alberto-morillo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

WilliamDAssafMSFT:
Thanks @alberto-morillo, your suggestions help improve our documentation. 

Adding @dimitri-furman to review and approve. #assign:dimitri-furman

docs/relational-databases/system-dynamic-management-views/sys-dm-db-partition-stats-transact-sql.md
```diff
@@ -98,6 +98,12 @@ FROM sys.dm_db_partition_stats
 WHERE object_id=OBJECT_ID('HumanResources.Employee')    AND (index_id=0 or index_id=1);  
 GO  
 ```  
+### D. Get the reserved space of an Azure SQL Database  
+ The following example uses the sys.dm_db_partition_stats DMV to get the current size of an Azure SQL Database.
```
dimitri-furman, line 102: "reserved space" and "current database size" are not the same thing though. What most customers consider as "current database size" is the sum of sizes of all files on disk, also known as "allocated size" in SQL DB documentation and in the portal. Reserved space can be less than that if there is empty space in files, which is not uncommon.

We can add this example, but we need to be careful to consistently refer to "reserved space" rather than "database size". Also, this example is not really specific to Azure SQL Database, but applies to any MSSQL database.

docs/relational-databases/system-dynamic-management-views/sys-dm-db-partition-stats-transact-sql.md
```diff
@@ -98,6 +98,12 @@ FROM sys.dm_db_partition_stats
 WHERE object_id=OBJECT_ID('HumanResources.Employee')    AND (index_id=0 or index_id=1);  
 GO  
 ```  
+### D. Get the reserved space of an Azure SQL Database  
+ The following example uses the sys.dm_db_partition_stats DMV to get the current size of an Azure SQL Database.  
+  
+```sql  
+SELECT SUM(reserved_page_count) * 8.0 / 1024 AS 'Database size [MB]' FROM sys.dm_db_partition_stats
```
dimitri-furman, line 105: Similar comment, column name should be "Reserved space MB", not "Database size".

dimitri-furman (changes requested):

docs/relational-databases/system-dynamic-management-views/sys-dm-db-partition-stats-transact-sql.md
WilliamDAssafMSFT: I'd add that this calculation does not match what we display in sys.database_files.size, sp_spaceused, or in Azure Portal. Perhaps this code snippet and some caveat/discussion about what it actually measures belongs in this article: https://docs.microsoft.com/sql/relational-databases/databases/display-data-and-log-space-information-for-a-database?view=sql-server-ver15

alberto-morillo: @dimitri-furman I fully agree about the allocated space vs DB size. Once I posted the suggestion I did not know how to make the adjustment. I also know about that example applies to SQL Server also, but as you can see on the QA forums, the azure-sql-database tag is a cinderella to me and the same I see on Microsoft Docs. That is why you will always find me adding Azure SQL examples on Microsoft Docs.
@WilliamDAssafMSFT The vision of what Microsoft would like to see in this section of the Docs that I do not have it. I will leave that in your hands. If you feel it is not appropriate here I am fine with that, If you feel I could submit this example on the URL you provided above, it is fine with me also. If the example is not worthy do not add it. I think if I remember well if you try sp_spaceused on Azure SQL you will receive an error, back on those days this query was very useful.

Thanks both for your time and considerations provided. If you need any action from me, please let me know. Have a great day.

dimitri-furman: sp_spaceused does work in Azure SQL.

alberto-morillo: sp_spaceused was not supported 6 years ago as it was reported [here](https://stackoverflow.com/questions/29636621/how-i-can-fix-this-error-migrate-to-sql-azure), but the example I provided has always work

WilliamDAssafMSFT:
Thanks @alberto-morillo for this suggestion. Will consider it for future Azure SQL-specific documentation improvements and review internally. Thanks again for this feedback, your suggestions help improve our documentation.

PR #6639, opened 2021-07-25
Update sys-dm-db-wait-stats-azure-sql-database.md

alberto-morillo:
This DMV requires an useful example of how to use the DMV with Azure SQL Database.


PRMerger10:
@alberto-morillo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

WilliamDAssafMSFT:
Thanks @alberto-morillo this is a good query. I don't believe we're moving in the directly of hosting a lot of these samples in our syntax reference, as we have other places where this type of query and its place in troubleshooting can be better explained, such as https://docs.microsoft.com/azure/azure-sql/database/understand-resolve-blocking. Will close with a consideration to expand on this type of thing in other troubleshooting articles.

PR #6642, opened 2021-07-26
Add and scalability to the description for dev ed.

MarkIannucci:
I noticed that developer edition isn't listed in the tables below.  I suspect that's because it is the same as Enterprise with the exception of the legal licensing scenarios.  There's some question though if it has the same unlimited scaling possibilities that SQL Server Enterprise does because we recall previous editions of SQL Server Developer lacking that.  I'm not sure if my change is warranted in the documentation.  If it is not, can Microsoft clarify this point for us?


PRMerger16:
@MarkIannucci : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MikeRayMSFT:
@Marklannucci - thanks for the submission. Verifying.

MikeRayMSFT:
@MarkIannucci - I'm still following up on this. Just FYI.

MikeRayMSFT:
@MarkIannucci  - at this time we're not going to update this. There are cases where scalability can be different in SQL Server Analysis Services, and as the product group partners have not explicitly approved this update we'll keep the content as is.

PR #6667, opened 2021-08-06
Added Clarity on Pre-reqs

SeanGallardy:
Added additional clarity on pre-reqs as it wasn't apparent what "needed" to be connected and why. Added additional requirement that the secondary is healthy, showing the secondary role, and connected to the primary (which is a requirement for being healthy and in the secondary role).


PRMerger9:

cawrites (commented):
reviewed

PR #6679, opened 2021-08-11
Added callout to a limitation in Azure Synapse Analytics

Illuminae:
Schema binding is currently not supported by views in Synapse SQL Pools. I think this should be called out explicitly and linked to avoid confusion.


PRMerger7:
@Illuminae : Thanks for your contribution! The author(s) have been notified to review your proposed change.

WilliamDAssafMSFT:
Thanks @Illuminae for taking the time to note this change. Your commit has been cherrypicked into another PR and will be merged as this article update is in flight. After the updated article is published, GitHub will list you as an article contributor. 

PR #6683, opened 2021-08-13
use http for xmlns

maciejpokorski:
With https version, SQL throws an error:
Msg 4855, Level 16, State 1, Line 3
Line 2 in format file "C:\t_floatformat-c-xml.xml": unexpected element "BCPFORMAT".

Solution:
https://stackoverflow.com/questions/54373618/why-does-format-file-cause-error-during-sql-server-bulk-insert

### DIFF
--- a/docs/t-sql/statements/bulk-insert-transact-sql.md
+++ b/docs/t-sql/statements/bulk-insert-transact-sql.md
@@ -285,7 +285,7 @@ The following format file uses the `SQLFLT8` data type to map the second data fi
 
 ```xml
 <?xml version="1.0"?>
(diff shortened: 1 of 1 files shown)


PRMerger7:
@maciejpokorski : Thanks for your contribution! The author(s) have been notified to review your proposed change.

WilliamDAssafMSFT:
@maciejpokorski Thanks for taking the time to submit this PR! This needed to be fixed for certain. Your commit has been cherrypicked as part of a large PR and will be merged as this article update is in flight. After the updated article is published, GitHub will list you as an article contributor. 

PR #6687, opened 2021-08-13
Using KILL in Active Database Deletion

saitorhan:


PRMerger20:
@saitorhan : Thanks for your contribution! The author(s) have been notified to review your proposed change.

cawrites:
@saitorhan thank you for contacting Microsoft. We appreciate it when we get suggestions from the community. After careful consideration of your suggestion, we have decided not to apply the update. This script can introduce risk and create an unstable environment. We are aware that many developers use scripts similar to this, but it is not something that Microsoft can endorse as a solution.

PR #6688, opened 2021-08-13
Update sys-sql-modules-transact-sql.md - Unnecessary object_name invocation 

DanielAdeniji:
Peers:-

It is preferable to directly use the object name from sys.objects rather than to invoke object_name function.

The reasons include:-

1) Metadata locks during object creation
2) The existence of sys.objects in the query, as is

Also, added sys.schemas so that we can include the schema name, as well.

### DIFF
--- a/docs/relational-databases/system-catalog-views/sys-sql-modules-transact-sql.md
+++ b/docs/relational-databases/system-catalog-views/sys-sql-modules-transact-sql.md
@@ -1,8 +1,8 @@
 ---
 description: "sys.sql_modules (Transact-SQL)"
-title: "sys.sql_modules (Transact-SQL) | Microsoft Docs"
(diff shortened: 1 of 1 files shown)


PRMerger17:
@DanielAdeniji : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger9:
@WilliamDAssafMSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

WilliamDAssafMSFT:
Thanks @DanielAdeniji for submitting this PR! I have added a commit to support your updates to the sample query and refresh other style things in the article. Your commit will be merged and this article update is in flight. After the updated article is published, GitHub will list you as an article contributor.  

#sign-off

PR #6695, opened 2021-08-16
Update break-transact-sql.md

MarkPryceMaherMSFT:
I've added an example for Synapse dedicated SQL Pool because waitfor is not supported.   So running the example will just fail.


PRMerger10:
@MarkPryceMaherMSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

cawrites (approved):

PR #6702, opened 2021-08-17
Update alter-database-scoped-configuration-transact-sql.md

vtpombei:
Add to the documentation that this feature is also supported by SQL MI and removed the indication that it was not supported by PDW.
The information I've is that if it's not supported then it should be visible, we should not use the cross to tell that it's not supported.


PRMerger10:
@vtpombei : Thanks for your contribution! The author(s) have been notified to review your proposed change.

WilliamDAssafMSFT:
Thanks @vtpombei for submitting this PR. Looks good. Your commit will be reviewed for merge and this article update is in flight. After the updated article is published, GitHub will list you as an article contributor. 

#sign-off

PR #6738, opened 2021-08-26
Update download page to release SSDT for VS2017 15.9.9

fan130:
Update download page to release SSDT for VS2017 15.9.9


dzsquared:
#sign-off

PR #6753, opened 2021-08-28
Update r-clustering-model-deploy.md

wvaughn409:
prevent implied authentication from blocking R script execution from within SQL Stored Procedure


PRMerger16:
@wvaughn409 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

wvaughn409:
updated instructions resolving potential issue with implied authentication

garyericson:
@wvaughn409 - Thanks for taking the time to work out the details of this solution. Can you explain a little more about why this is needed? Is it only for SQL Server Express Edition and only if the user wants to use Windows authentication? What happens if they try it without this workaround? And does doing this configuration restrict the user to only using Windows authentication or could they still use SQL authentication?

PR #6774, opened 2021-09-02
Update cast-and-convert-transact-sql.md

jasonhorner:
added a section to the table to call out the rounding rules when casting a datetime to a time.. The rounding behavior seems to me to be undocumented or a bug. It definitely was unexpected see this post on SO for further discussion: https://stackoverflow.com/questions/69033081/t-sql-truncate-time-not-round-to-the-nearest-minute it's proably worth adding an example to the PR to demonstrate the behavior. If I can get confirmation that this behavior is by design and not a bug from the product team I will be glad to amend the PR.


PRMerger20:
@jasonhorner : Thanks for your contribution! The author(s) have been notified to review your proposed change.

cawrites:
@jasonhorner I appreciate you taking the time to provide this insight. I am going to share this insight with my team and see what next steps will be.

rothja:
@WilliamDAssafMSFT William, could you take a look at this? Thank you! #reassign:williamdassafmsft

WilliamDAssafMSFT:
Hi @jasonhorner! Thanks for taking the time to make this PR and apologize for the delay in a resource properly addressing it. I don't think this is quite the intent of the table - the conversion of datetime to time, and other date/time types, isn't a matter of decimal places, but of scale and possible the style involved. That much is documented elsewhere in this same article, or should be made plain by the document capabilities of each data type.

PR #6779, opened 2021-09-04
Fix some numbers and Add some internal configurations

taroarai:
Here are my research result for each configuration that is different from SQL Server 2019.

- access check cache bucket count
From the code, it use 16384 for x86 and 65536 for x64, but as all supported versions are x64 only now, I change it to 65536.

- ADR cleaner retry timeout (min)
Actual default values is 0 when first started after clean setup, but I'm still investigating if 0 is ok or not.

- ADR Preallocation Factor
Actual default values is 0 when first started after clean setup, but when 0 is set, it will use 4 from the code.

- allow filesystem enumeration
I think it is internal, but customers can see from sp_configure or sys.configurations, we should document this, so some customers happen to changed, they can change back to default.

- column encryption enclave type
Max values is 2 from sp_configure or sys.configurations, but 1 is supported value as VSM.
I guess 2 is used for Azure SQL Database with SGX.

- min server memory (MB)
Run_value will use 16 MB if Configure_values is lower, so I added a comment.

- version high part of SQL Server / version low part of SQL Server
It is also internal, and used for replicated master, but the same as "allow filesystem enumeration", customers can see it, so it is better to have default values on here.

### DIFF
--- a/docs/database-engine/configure-windows/server-configuration-options-sql-server.md
+++ b/docs/database-engine/configure-windows/server-configuration-options-sql-server.md
@@ -70,16 +70,17 @@ The following table lists all available configuration options, the range of poss
 
 | Configuration option | Minimum value | Maximum value | Default |
 |--|--|--|--|
-| [access check cache bucket count](../../database-engine/configure-windows/access-check-cache-server-configuration-options.md) (A) | 0 | 16384 | 0 |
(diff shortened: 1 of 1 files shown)


PRMerger8:

MikeRayMSFT:
Requesting review or reviewer recommendation from @pmasl

PR #6781, opened 2021-09-06
Polices are missing from the default location

kasarava:
Microsoft Best Practice policies are missing on the default installation location. There is no Polices folder under C:\Program Files (x86)\Microsoft SQL Server\140\Tools\.  Please let us know if I can download the polices from any other location.


PRMerger15:
@kasarava : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6783, opened 2021-09-06
Update sys-dm-db-task-space-usage-transact-sql.md

sqlgt0:
|From|To|Relationship|  
|----------|--------|------------------|  
|dm_db_task_space_usage.request_id|dm_exec_requests.request_id|Many-to-one|  

The combination of session_id + request_id in sys.dm_db_task_space_usage might not always be unique in the scenario when parallel thread is spun using the same request_id as per the attached screenshot. In this situation, you would need to filter the sys.dm_db_task_space_usage DMV with exec_context_id = 0 to show only the active row.

### DIFF
--- a/docs/relational-databases/system-dynamic-management-views/sys-dm-db-task-space-usage-transact-sql.md
+++ b/docs/relational-databases/system-dynamic-management-views/sys-dm-db-task-space-usage-transact-sql.md
@@ -87,9 +87,7 @@ On SQL Database **Basic**, **S0**, and **S1** service objectives, and for databa
   
 ## Relationship Cardinalities  
   
-|From|To|Relationship|  
-|----------|--------|------------------|  
-|dm_db_task_space_usage.request_id|dm_exec_requests.request_id|One-to-one|  
+![Capture](https://user-images.githubusercontent.com/73323764/132151558-1440cc03-4a35-42f4-9fb1-9d8e399af4d3.PNG)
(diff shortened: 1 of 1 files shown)


PRMerger9:
@sqlgt0 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

sqlgt0:
The combination of session_id + request_id in sys.dm_db_task_space_usage might not always be unique in the scenario when parallel thread is spun using the same request_id as per the attached screenshot. In this situation, you would need to filter the sys.dm_db_task_space_usage DMV with exec_context_id = 0 to show only the active row.

image

WilliamDAssafMSFT:
Thanks @sqlgt0 for contacting Microsoft with feedback about our product documentation. We'll consider a suggestion like this for applied query samples where this scenario would come into play. We'll contact you if we need further information, and you'll receive an email message each time we "@mention" you in a comment.

sqlgt0:
[dm_db_task_space_usage.docx](https://github.com/MicrosoftDocs/sql-docs/files/7125936/dm_db_task_space_usage.docx)

Put a more descriptive statement where the session_id + request_id relationship between sys.dm_db_task_space_usage and sys.dm_exec_requests will not be 1-to-1 when the same task runs in parallel. A lot of our company monitoring tool assume the 1-to1 relationship and we're getting lots of errors due to duplicates. Unless the statement is more clear, our engineers are having a hard time to explain why we need to update our code because it goes against this official Microsoft documentation. I'd really appreciate if this scenario is properly reflected in the documentation.

PR #6803, opened 2021-09-11
Fix grammar in understanding-isolation-levels.md

ericmutta:
The following sentence:

> Whether **a read operation** referencing rows modified by another transaction:

Should read:

> Whether **read operations** referencing rows modified by another transaction:

...due to the way the sentences in the list below that line are formed (e.g. without this fix, the sentences would read "Whether a read operation...block until" which is incorrect. With the fix that now reads as "Whether read operations...block until").

### DIFF
--- a/docs/connect/jdbc/understanding-isolation-levels.md
+++ b/docs/connect/jdbc/understanding-isolation-levels.md
@@ -24,7 +24,7 @@ Transaction isolation levels control the following effects:
 
 - How long the read locks are held.
 
-- Whether a read operation referencing rows modified by another transaction:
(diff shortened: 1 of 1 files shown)


PRMerger17:
@ericmutta : Thanks for your contribution! The author(s) have been notified to review your proposed change.

David-Engel (approved):
Looks good to me. Thanks for catching that!

David-Engel:
#sign-off

ericmutta:
Many thanks @David-Engel and @ktoliver for following up :+1:

PR #6823, opened 2021-09-17
Specify temp table sessions

ronoudgenoeg:
The original phrasing makes it sound like the same user across various sessions would re-use the same local temp table, but this is not the case. The same user across various sessions will create different local tables.


PRMerger14:
@ronoudgenoeg : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

WilliamDAssafMSFT:
Thanks @ronoudgenoeg for submitting this PR. It's a good edit, I could see how this improves the understanding. Your commit will be merged and this article update is in flight. After the updated article is published, GitHub will list you as an article contributor. 

#sign-off

PR #6838, opened 2021-09-20
Fix typo in backup-overview-sql-server.md

ericmutta:
The following sentence:

> Regardless of which of the **conflicting operation** began first...

Should read:

> Regardless of which of the **conflicting operations** began first...


ktoliver:
Looks good; thanks.

PR #6850, opened 2021-09-22
Update alter-table-transact-sql.md

arunguru-msft:
filter column in data_deletion is a required parameter


PRMerger4:
@arunguru-msft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

arunguru-msft:
Apologize looks like the current description is correct. My bad

PR #6854, opened 2021-09-23
XQuery is case-sensitive and the code included multiple UPPER CASE in…

pituach:
…stead of LOWER CASE

XQuery is case-sensitive and the code included multiple UPPER CASE instead of LOWER CASE
for example XQuery modify function uses "insert" instead of "INSERT"


PRMerger16:

ktoliver:
#label:"aq-pr-triaged"

rothja:
@pituach Ronen, thanks for this fix, and sorry for the delayed response! #sign-off

PR #6862, opened 2021-09-25
Update start-sql-server-in-single-user-mode.md

suresh-kandoth:
The code example used -m instead of /m. NET START will not accept -m.
See https://docs.microsoft.com/en-us/sql/database-engine/configure-windows/database-engine-service-startup-options?view=sql-server-ver15#using-startup-options-for-troubleshooting


PRMerger8:

PR #6881, opened 2021-09-29
Added scenarios which will not work disabling distributor_admin account

sbhuwalka:
Added scenarios which will not work in case distributor_admin account at distributor is disabled both for remote and local distributor.


PRMerger4:
@sbhuwalka : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6890, opened 2021-09-30
Update sp-configure-transact-sql.md

SassiJ:
While 'show advanced option' will parse, the more correct example should be 'show advanced options' as this is the setting name in sp_configure.


PRMerger13:
@SassiJ : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6908, opened 2021-10-06
Update system-user-transact-sql.md

jc-odot:
Replaced C: with C. so it matches the A. B. etc notation in this document.


PRMerger12:
@jc-odot : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6957, opened 2021-10-19
Corrected terminology and other things

allanhirt:
1. There is no such thing as a WSFC cluster. It's a WSFC - Windows Server Failover Cluster. I've fixed this elsewhere. I know I wrote this right when I wrote it for docs so somethign happened.
2. Added a reference to Pacemaker since it's applicable there, too.


PRMerger7:

allanhirt:
Also, it looks like you pulled the note which needs to come back - Never use DAG for a distributed AG. Why was that section deleted? That is a recent deletion.

PR #7003, opened 2021-11-04
where clause is optional

tjeske:
I guess something like `SELECT * FROM ( SELECT * FROM myCube )` is allowed, or?

### DIFF
--- a/docs/mdx/mdx-data-manipulation-select.md
+++ b/docs/mdx/mdx-data-manipulation-select.md
@@ -59,8 +59,8 @@ FROM
            <SELECT query axis clause>,...n ] )   
          ]   
             FROM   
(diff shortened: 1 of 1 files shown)


PRMerger19:
@tjeske : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

Minewiskan:
@tjeske - It's hard to tell if you're asking a question about this clause, or making a change based on a use case. If the latter, can you show how you came to that determination? The original writing of this is assumed correct, so we'll need more to go on to before committing resources to vet your change. If you have a question regarding this use of this clause, please use support channels.

Minewiskan:
Closing due to no response.

PR #7005, opened 2021-11-04
`MEASURE` is not a keyword

tjeske:


PRMerger14:
@tjeske : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

tjeske:
@Minewiskan I think there was a misunderstanding. `MEASURE` is an MDX keyword but using `MEASURE` as a keyword in this context doesn't make sense, or? In the docs `MEASURE` is written as a keyword but I think it is about cell properties if I look e.g. at https://github.com/MicrosoftDocs/bi-shared-docs/blob/master/docs/analysis-services/multidimensional-models/mdx/mdx-cell-properties-using-cell-properties.md?plain=1#L87.

PR #7061, opened 2021-11-29
fix missing line feed and useless <br>

jlecordier:
- A line feed was missing, making two links to appear next to one another, instead of being on two separate lines.
- `<br>` tags were used in the `Related Content` section, which is not coherent with the `Related Tasks`


PRMerger6:
@jlecordier : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Jak-MS:
@jlecordier - thanks.

PR #7070, opened 2021-11-30
Change >= to >

zikato:
Otherwise, the boundary value 500 would qualify for two rows of the table.
Changed it to match the pre compatibility level 130 table.


PRMerger13:
@zikato : Thanks for your contribution! The author(s) have been notified to review your proposed change.

WilliamDAssafMSFT:
Hi @zikato, this is correct, the 20% logic comes into play when the Object > 500. Thanks for submitting this PR. Your commit will be merged and this article update is in flight. After the updated article is published, GitHub will list you as an article contributor. 

#sign-off

PR #7073, opened 2021-11-30
Fixed Page title

way0utwest:
The HTML title shows proper case, but not the documentation inline. Updated the H1 to be consistent


PRMerger4:

MikeRayMSFT:
@way0utwest - thank you for this (and all of your thoughtful community activity). 

Microsoft uses [sentence-style capitalization](https://docs.microsoft.com/style-guide/capitalization). We're going to keep the heading as is. 

I recognize that our current presentation is inconsistent, because when SQL Server Books Online was introduced it uses a different capitalization style and we've never gone through and updated everything. In this case "Data type precedence" is the correct terminology. 

I've submitted https://github.com/MicrosoftDocs/sql-docs-pr/pull/21052 against the private repo to fix some of the inconsistencies with this article (metadata and headings). That'll be live shortly.

Thanks,
Mike

#please-close

PR #7096, opened 2021-12-06
Update Ubuntu installation instructions with lsb_release -rs

jlaundry:
Short PR: the installation instructions can be greatly simplified and automated by using $(lsb_release -rs) instead of manually selecting the version number.

### DIFF
--- a/docs/connect/odbc/linux-mac/installing-the-microsoft-odbc-driver-for-sql-server.md
+++ b/docs/connect/odbc/linux-mac/installing-the-microsoft-odbc-driver-for-sql-server.md
@@ -159,20 +159,7 @@ sudo zypper install -y unixODBC-devel
 sudo su
 curl https://packages.microsoft.com/keys/microsoft.asc | apt-key add -
 
-#Download appropriate package for the OS version
-#Choose only ONE of the following, corresponding to your OS version
-
-#Ubuntu 16.04
-curl https://packages.microsoft.com/config/ubuntu/16.04/prod.list > /etc/apt/sources.list.d/mssql-release.list
-
-#Ubuntu 18.04
-curl https://packages.microsoft.com/config/ubuntu/18.04/prod.list > /etc/apt/sources.list.d/mssql-release.list
-
-#Ubuntu 20.04
-curl https://packages.microsoft.com/config/ubuntu/20.04/prod.list > /etc/apt/sources.list.d/mssql-release.list
-
-#Ubuntu 21.04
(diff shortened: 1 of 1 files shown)


PRMerger20:
@jlaundry : Thanks for your contribution! The author(s) have been notified to review your proposed change.

David-Engel:
That certainly does shorten the instructions. However, I would not support doing this. It would lead to questions from a lot of users about why their installation fails on an unsupported version. People often skip the system requirements and assume the latest version (or an old version) of Ubuntu (and others) is supported as soon as Ubuntu releases it. Keeping the version reference in the installation instructions makes it more obvious when users try to walk through them and realize instructions don't exist for their version. We already get questions from people who ignore even the obviousness of the current instructions when they lack a reference to a version they want to install on.

Regards,
David

jlaundry:
Hi @David-Engel, I understand your position, but I don't think we can solve for users who blatantly ignore prerequisites. However, not being able to 'obviously' automate deployment is a real challenge for teams I work with.

If you do want to have a hard check in the deployment process, you could add a check like:

```
if ! [[ "16.04 18.04 20.04 21.04" == *"$(lsb_release -rs)"* ]];
then
    echo "Ubuntu $(lsb_release -rs) is not currently supported.";
    exit;
fi
```

Also, please consider that other Microsoft teams have standardised on the dynamic approach:

- https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-linux?pivots=apt#option-2-step-by-step-installation-instructions
- https://docs.microsoft.com/en-us/azure/azure-functions/functions-run-local?tabs=v4%2Clinux%2Ccsharp%2Cportal%2Cbash%2Ckeda#install-the-azure-functions-core-tools
- https://docs.microsoft.com/en-us/defender-cloud-apps/discovery-docker-ubuntu?tabs=ubuntu#step-2--on-premises-deployment-of-your-machine

David-Engel:
> If you do want to have a hard check in the deployment process, you could add a check like:
> 
> ```
> if ! [[ "16.04 18.04 20.04 21.04" == *"$(lsb_release -rs)"* ]];
> then
>     echo "Ubuntu $(lsb_release -rs) is not currently supported.";
>     exit;
> fi
> ```

I like that. If you update your PR, I'll approve the change. Thanks!

PRMerger7:
@jlaundry : Thanks for your contribution! The author(s) have been notified to review your proposed change.

David-Engel (approved):

David-Engel:
#sign-off

PR #7151, opened 2021-12-31
Adding feature line for Lock Pages In Memory

dbrownems:
It's documented on the feature page https://docs.microsoft.com/en-us/sql/database-engine/configure-windows/server-memory-server-configuration-options?view=sql-server-ver15#lock-pages-in-memory-lpim  But to follow the practice of consolidating edition feature differences on this page it should be here too.

### DIFF
--- a/docs/sql-server/editions-and-components-of-sql-server-version-15.md
+++ b/docs/sql-server/editions-and-components-of-sql-server-version-15.md
@@ -193,6 +193,7 @@ The Developer edition continues to support only 1 client for [[!INCLUDE[ssNoVers
 |In-Memory Database: hybrid buffer pool|Yes|Yes|No|No|No|
 |In-Memory Database: memory-optimized tempdb metadata|Yes|No|No|No|No|
 |In-Memory Database: persistent memory support|Yes|Yes|Yes|Yes|Yes|
+|Lock Pages In Memory|Yes|Yes|No|No|No|
 |Stretch database|Yes|Yes|Yes|Yes|Yes|
 |Multi-instance support|50|50|50|50|50|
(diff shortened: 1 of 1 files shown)


PRMerger20:
@dbrownems : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

MikeRayMSFT:
Thank you @dbrownems - #sign-off

MikeRayMSFT:
@dbrownems David, after reviewing this with the product group we have determined that we will not update the article. Lock pages in memory is not a SQL Server feature or component. It's an operating system feature that you can configure for SQL Server to prevent paging. 


The change is rolled back in https://github.com/MicrosoftDocs/sql-docs-pr/pull/21053/commits/68fc1166c8b287957ee15fcf6cea3078003baebf


CC: @pmasl

davidbaxterbrowne:
> 

But _actually_ using the AWE APIs to allocate locked pages is a feature exclusive to Enterprise Edition.  Other editions won't use Locked Pages even if the service has the privilege.  It's just that SQL Server doesn't have a knob to enable the behavior, instead relying on the presence or absence of the OS privilege.

pmasl:
David I understand your point, but "locked pages" is an OS feature that allows locking physical memory, preventing OS paging of the locked memory in use by the sql mem manager. As such it's not a SQL feature (like merry-go-round scan is for example). We can find other ways of exposing applicable editions in documentation, but not a SQL feature list. What was your intent, or problem space you're trying to address, so we can better help contextualize?

dbrownems:
This doc page is where people go to evaluate the value proposition of Enterprise Edition.  It should not omit any significant behavior difference between the editions.

PR #7162, opened 2022-01-05
Update .acrolinx-config.edn

v-savila:
update to reflect default branch name change


ktoliver:
@v-savila Should this config file be updated via repo sync with the private repo? (The pull request was opened in the public repo; Acrolinx doesn't run in the public repo, but the file should be synced with the file in the private repo.)

v-savila:
@ktoliver thanks, I am in the process of updating all the acrolinx config file to add "main" branch for the default branch renaming from master to main. I do see this repo is already setup to sync with the private repo live branch so we may be good.

PR #7213, opened 2022-01-19
correct URI to reference correct http protocol

jephcotj:
Note the URI here is acting as a unique identifier that references the values in the sys.services table, and not as a web URL and so needs to be a text match the value in the sys.services table, that is a value with http and not https


PRMerger10:
@jephcotj : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

markingmyname (approved):
#sign-off

PR #7215, opened 2022-01-19
Update dml-triggers.md

AntonC9018:
On line 49, I'm confused as to whether the trigger actually gets to do the final insert itself, or if it just allows or denies it. I feel like "before inserting the record into the payroll table" implies that, at the end of the day, trigger is the one doing the inserting. It should be changed to something like "before the insertion of the record happens/takes place/finishes/succeeds/goes through" if it doesn't do the insert itself. Am I wrong?


PRMerger20:
@AntonC9018 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

MikeRayMSFT (approved):

MikeRayMSFT:
@AntonC9018 - thank you for the submission. It corrects some language problems with the article as is. 

In the case of an `INSTEAD OF` trigger, the command of the trigger is executed in lieu of the command that launched the trigger, so the current example: 

> a trigger can be defined to either produce an error message and roll back the transaction, or insert a new record into an audit trail before inserting the record into the payroll table

The trigger completes the defined command. 

If that doesn't answer the question, please ask at SQL Server [Q&A](https://docs.microsoft.com/en-us/answers/topics/sql-server-general.html).

#sign-off

AntonC9018:
@MikeRayMSFT makes sense, thank you!

PR #7285, opened 2022-02-07
Temporal table data range should be OR instead of AND

tinkujv:
The temporal table data range in the document mentions AND but it should be OR instead


PRMerger13:
@tinkujv : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

rwestMSFT:
Hi @tinkujv, I hope to clarify your misunderstanding.

Firstly, the article was updated this week to make the table a little clearer. We are now using `ValidFrom` and `ValidTo` in the examples throughout our docs as well.

As written, the `FROM..TO` and `BETWEEN` sub-clauses for temporal queries must include both boundaries as noted (which requires the AND condition for the logical explanation of the PERIOD columns).

The main difference is that the `BETWEEN` includes the upper boundary, while the `FROM..TO` does not.

Consider this small table for example:

    CREATE TABLE dbo.Employee
    (
    	[EmployeeID] INT NOT NULL IDENTITY(1,1) PRIMARY KEY CLUSTERED,
    	[Name] NVARCHAR(100) NOT NULL,
    	[Position] NVARCHAR(100) NOT NULL,
    	[Department] NVARCHAR(100) NOT NULL,
    	[Address] NVARCHAR(1024) NOT NULL,
    	[AnnualSalary] DECIMAL(10, 2) NOT NULL,
    	[ValidFrom] DATETIME2(3) GENERATED ALWAYS AS ROW START,
    	[ValidTo] DATETIME2(3) GENERATED ALWAYS AS ROW END,
    	PERIOD FOR SYSTEM_TIME(ValidFrom, ValidTo)
    )
    WITH (SYSTEM_VERSIONING = ON);
    GO

If you run a temporal query against this table using `FROM..TO`, providing a period range of a start and end date, the results will include all rows where (logically speaking) `ValidFrom` is less than the `<end_date_time>` provided, **and** `ValidTo` is greater than the `<start_date_time>` value provided.

Running the same temporal query using `BETWEEN` is only slightly different. The upper bound will **include** the `<end_date_time>` provided. So you will see all rows where `ValidFrom` is less than and equal to the `<end_date_time>` provided, **and** `ValidTo` is greater than the `<start_date_time>` value provided.

Where I think the confusion comes from, is that `<end_date_time>` is related to `ValidFrom`, and `<start_date_time>` is related to `ValidTo`.

Please reach out if you have any other questions.

rwestMSFT:
#please-close

markingmyname:
resolved.

PR #7317, opened 2022-02-17
 update '-hostname' to '--hostname'

dlumpp:
The long version of the `-h` option should be `--hostname` rather than `-hostname`
See [docker run reference](https://docs.docker.com/engine/reference/commandline/run/)


PRMerger18:
@dlumpp : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
Looks good; merging. Thanks @dlumpp.

PR #7374, opened 2022-03-10
change wording of a sentence to clarify the effect of disabling a nonclustered index on a table

agg06:
I propose a slight modification for a sentence in the first paragraph to make it clear that disabling a nonclustered index on a table also deletes the index data. Current version could be understood as if only indexes on a view are referenced.


PRMerger8:
@agg06 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #7382, opened 2022-03-12
Update try-catch-transact-sql.md

jBarrineau:
The document has a section labeled "Examples: Azure Synapse Analytics and Analytics Platform System (PDW)" but its completely out of place as nothing in the document is related to that.

Change: removed that section header

### DIFF
--- a/docs/t-sql/language-elements/try-catch-transact-sql.md
+++ b/docs/t-sql/language-elements/try-catch-transact-sql.md
@@ -325,9 +325,7 @@ BEGIN CATCH
 END CATCH;  
 GO  
 ```  
-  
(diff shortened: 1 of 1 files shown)


PRMerger13:
@jBarrineau : Thanks for your contribution! The author(s) have been notified to review your proposed change.

LitKnd:
Hi,

Thanks for this PR, @jBarrineau , and apologies for the delay in responding. Looking at the article, I agree that example D isn't specific to Synapse, and that heading shouldn't be there. Looking in the history of commits for the article, I can't see a reason why the heading was added-- I believe it was simply a mistake.

Thanks for helping us make the docs better!
Kendra Little from the Database Docs team

#sign-off

PR #7414, opened 2022-03-21

RaufAsadov:
I did experiments and came to the conclusion that in order to backup database or create mdf file onto shared folder on different machine,  SQL Server service startup account must be granted SE_MANAGE_VOLUME_NAME on that remote machine as well. I performed the following steps in order to enable IFI on the remote server:

1. Adding SQL Server service startup account to Grant Perform Volume Maintenance security policy on local machine.
2. Restart SQL Server engine service.
3. Adding SQL Server service startup account to Grant Perform Volume Maintenance security policy on remote machine (restarting SQL Server engine service is not required).

### DIFF
--- a/docs/relational-databases/databases/database-instant-file-initialization.md
+++ b/docs/relational-databases/databases/database-instant-file-initialization.md
@@ -40,7 +40,7 @@ Instant file initialization is only available if the [!INCLUDE[ssNoVersion](../.
 > Some feature usage, such as [Transparent Data Encryption (TDE)](../../relational-databases/security/encryption/transparent-data-encryption.md), can prevent Instant File Initialization.  
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


PRMerger3:
@RaufAsadov : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Jak-MS:
#label:"aq-pr-triaged"

RaufAsadov:
@WilliamDAssafMSFT  Thank you for your response. I agree with avoiding creation of mdf files on remote network shares. In my scenario, we can't store backups locally, so we take backups into remote dedicated server where we also keep other backups. I am aware that Ideally we need to take backups locally and then copy those backups in asynchronous way into remote server, but as I said before, we have different policies where we also need to consider DPM. Anyway, by default whenever I start backup process, backup file will firstly be initialized with zeros which takes much time. The question came into my mind about how can I avoid zeroing backup file on remote server. I couldn't find anything about it in official docs or anywhere else, I figured it out by myself. I guess it can't be included as a recommendation but rather an optional step for performing backups into remote server like in my scenario. What do you think about it?

PR #7427, opened 2022-03-27
Update creating-a-system-versioned-temporal-table.md

rjk:
Better constraint names in this example, since they must be unique within the database.

Previously I copied and used the existing example and later realised it was a bad constraint name. Better to have a good example, i.e. a properly unique constraint name.


PRMerger7:
@rjk : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rwestMSFT:
@rjk Thank you for catching this. As this issue is likely to impact other code samples in this section, I will probably do the work in an internal PR.

PR #7457, opened 2022-04-05
updating doc to clarify that MAXDOP is not a property of an index, bu…

dbrownems:
…t a property of an index modification DDL statement

And removing the description for changing the MAXDOP for an index in SSMS, for while the GUI allows the operation, it actually is a noop.  As MAXDOP is simply not a persistent attribute of an index, but only a property of an index modification DDL statement.


PRMerger3:
@dbrownems : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Court72:
#label:"aq-pr-triaged"

MikeRayMSFT:
#please-close

Migrating to private repo for review.

PR #7565, opened 2022-05-12
Update hadr-cluster-best-practices.md

jasonplante1:
Customer brought this to my attention that we call out 20 seconds (20000 ms) for our relaxed monitoring, but then give T-SQL for 15 seconds.

My gut take on this is that common sense would win out here and you'd update the value based on what you feel is best, but that didn't happen on at least a case I've had. My opinion following this case is that this T-SQL example would be better if the value we provide is set to match the value we provide for relaxed monitoring so there's no confusion at all.

### DIFF
--- a/azure-sql/virtual-machines/windows/hadr-cluster-best-practices.md
+++ b/azure-sql/virtual-machines/windows/hadr-cluster-best-practices.md
@@ -262,7 +262,7 @@ Use Transact-SQL (T-SQL) to modify the **session timeout** for an availability g
 
 ```sql
 ALTER AVAILABILITY GROUP AG1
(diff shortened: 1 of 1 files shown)


PRMerger4:
@jasonplante1 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Court72:
@rajeshsetlem

#label:"aq-pr-triaged"

rothja:
@MashaMSFT This PR is not assigned to anyone. I am assigning you for now to look into it. Thanks! #assign:mashamsft

tarynpratt:
#sign-off

MashaMSFT:
#sign-off

MashaMSFT:
#reassign:MashaMSFT

PR #7607, opened 2022-05-24
Updated 'Tip' and WinRM req.

StewartHandsMSFT:
CSS Case 2205180050000597.

This document led to some confusion on deploying a multi-subnet cluster. You cannot do this from the Azure Portal at present and even leaves the AG unable to failover (due to second subnet IP not being requested or added to the cluster dependencies). Although the article states it's for single subnet, the tip does confuse matters.

Also, the deployment requires WinRM ports to be open for the NSG of the VMs for successful deployment.


PRMerger20:
@StewartHandsMSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rwestMSFT:
#reassign:rwestMSFT

PR #7620, opened 2022-05-25
Updating the semantics section

mchrupczalski:
"One image can express thousand words".
Added images to better explain where each value in the hierarchy can be set/updated.
Converted list to table (used html tags for better formatting - not sure if permitted)


PRMerger7:
@mchrupczalski : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jborsecnik:
@MikeRayMSFT, can you review the proposed changes? Note that for this to be merged, the new file names would have to lose their capital letters, and the new html table would have to be converted to Markdown.

IMPORTANT: When the changes are ready for publication, add a `#sign-off` comment to signal that the PR is ready for the review team to merge. Thanks.

MikeRayMSFT:
[mchrupczalski](https://github.com/mchrupczalski) - thank you so much for the images and the update. Unfortunately, we can't accept public contributions in html. The table needs to be presented in markdown. 

In addition, file names need to be in lower case.

For now, I'm going to close this pull request, so it won't be merged.

You may resubmit with markdown instead of HTML if you wish.

I will also share your update with some product group partners to evaluate if we should update the layout based on your thoughtful recommendations.

Mike Ray

#please-close

PR #7631, opened 2022-05-31
Update alter-role-transact-sql.md

charithcaldera:
ALTER ROLE only supported for Azure Synapse Analytics Serverless SQL Pool.
This is not mentioned in Syntax. (Since customers copy / paste the syntax from the public documents, this needs to be specified under syntax comments.

### DIFF
--- a/docs/t-sql/statements/alter-role-transact-sql.md
+++ b/docs/t-sql/statements/alter-role-transact-sql.md
@@ -50,7 +50,7 @@ ALTER ROLE  role_name
   
  
 ```syntaxsql
--- Syntax for SQL Server 2008, Azure Synapse Analytics and Parallel Data Warehouse
+-- Syntax for SQL Server 2008, Azure Synapse Analytics Serverless SQL Pool and Parallel Data Warehouse
   
 -- Change the name of a user-defined database role  
 ALTER ROLE role_name   


PRMerger19:
@charithcaldera : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ShannonLeavitt:
@VanMSFT – Can you review the proposed changes? IMPORTANT: When the changes are ready for publication, add a `#sign-off` comment to signal that the PR is ready for the review team to merge. 

#label:"aq-pr-triaged"

VanMSFT:
@WilliamDAssafMSFT - Hey William, can you please verify and approve? Thanks!

#assign:WilliamDAssafMSFT

WilliamDAssafMSFT:
Hi @charithcaldera thanks for taking the time to make a suggestion. In this case we shouldn't just provide this Applies To information in a code comment - those aren't translated - but in the Arguments/Remarks, or if it doesn't apply to a whole product, potentially in the Applies To line and moniker section of the metadata. 

However, I just tested and I was able to create and rename a role in a dedicated SQL pool in both standalone and in a Synapse workspace. The article appears to be correct in that you cannot add a member in a dedicated SQL pool in both standalone and in a Synapse workspace via the ALTER ROLE. Can you be more specific as to what platforms and what behavior is not supported?

charithcaldera:
> Hi @charithcaldera thanks for taking the time to make a suggestion. In this case we shouldn't just provide this Applies To information in a code comment - those aren't translated - but in the Arguments/Remarks, or if it doesn't apply to a whole product, potentially in the Applies To line and moniker section of the metadata.
> 
> However, I just tested and I was able to create and rename a role in a dedicated SQL pool in both standalone and in a Synapse workspace. The article appears to be correct in that you cannot add a member in a dedicated SQL pool in both standalone and in a Synapse workspace via the ALTER ROLE. Can you be more specific as to what platforms and what behavior is not supported?

Hi @WilliamDAssafMSFT ,
Thank you for the clarification. 

"ALTER ROLE ADD MEMBER / DROP MEMBER" does not work with Synapse Dedicated SQL Pools.
As you mentioned, "ALTER ROLE ALTER NAME" works as expected. 

I noticed already its mentioned in the document already, but I was having the concern about the syntax area, which you already answered it. 
I will close this PR since there is no further actions. 

Thank you @WilliamDAssafMSFT.

PR #7655, opened 2022-06-07
Quote default in examples

zikato:
The default has to be quoted otherwise this error shows up
The keyword DEFAULT is not allowed in DBCC commands.


PRMerger7:
@zikato : Thanks for your contribution! The author(s) have been notified to review your proposed change.

rwestMSFT:
#sign-off

PR #7668, opened 2022-06-09
Update creating-a-system-versioned-temporal-table.md

himanshu-ram-punjabi:
Adding a space between the and PERIOD on Line 46.


rwestMSFT:
@himanshu-ram-punjabi Thanks for the eagle-eyed fix!

#sign-off

PR #7694, opened 2022-06-16
Added an additional preview limitation

srinia:
Added an additional limitation on the private end points usage on the Job Agents target databases and pools


WilliamDAssafMSFT (commented):
Approved.

PR #7701, opened 2022-06-17
Nondeterminism of STRING_SPLIT

lehiester:
In [this commit](https://github.com/MicrosoftDocs/sql-docs/commit/f12ff508b69db557347846748ef7f92a62905ddc), the output of `STRING_SPLIT` was clarified to be nondeterministic:

> The output rows might be in any order.

However, `string-functions-transact-sql.md` has not been updated and currently suggests that `STRING_SPLIT` is deterministic:

> All built-in string functions except `FORMAT` are deterministic.

This change simply updates the statement to read instead

> All built-in string functions except `FORMAT` and `STRING_SPLIT` are deterministic.

### DIFF
--- a/docs/t-sql/functions/string-functions-transact-sql.md
+++ b/docs/t-sql/functions/string-functions-transact-sql.md
@@ -145,7 +145,7 @@ The following scalar functions perform an operation on a string input value and 
 :::row-end:::
 
   
- All built-in string functions except `FORMAT` are deterministic. This means they return the same value any time they are called with a specific set of input values. For more information about function determinism, see [Deterministic and Nondeterministic Functions](../../relational-databases/user-defined-functions/deterministic-and-nondeterministic-functions.md).  
+ All built-in string functions except `FORMAT` and `STRING_SPLIT` are deterministic. This means they return the same value any time they are called with a specific set of input values. For more information about function determinism, see [Deterministic and Nondeterministic Functions](../../relational-databases/user-defined-functions/deterministic-and-nondeterministic-functions.md).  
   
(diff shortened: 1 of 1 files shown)


PRMerger12:
@lehiester : Thanks for your contribution! The author(s) have been notified to review your proposed change.

lehiester:
A little more context: I almost got code into production that could have broken at any moment because I was relying on `STRING_SPLIT` being deterministic, because I hadn't yet found the more recent statement indicating that it is not--not just being pedantic.

I was doing e.g.

`SELECT` [value], ROW_NUMBER() OVER (ORDER BY (SELECT NULL)) FROM STRING_SPLIT('foo/bar', '/')`

in order to attempt to compensate for the lack of the recently added `ordinal` column in the version of SQL Server I'm using.

Court72:
@mikeraymsft

#label:"aq-pr-triaged"

lehiester:
I suppose this updated wording might need a correction depending on the implementation--I'm aware that once a set of rows is stored in a table that there's no way to retrieve them in a deterministic order without an explicit ORDER BY, but in this case what matters is the ordering of rows returned by the function itself _before_ it's stored--execution plans I've looked at so far have their own block for `STRING_SPLIT` as though the query optimizer treats it as a black box that it can't modify and it's invariant to whatever the query optimizer decides to do.  I suppose it's possible, though, that `STRING_SPLIT` could return rows in a deterministic order strictly speaking but there is no way to _use_ those rows in a way that preserves the order.  Not having the code, I'm unable to investigate this.

MikeRayMSFT:
Per discussion with engineering:
Determinism property applies to scalar UDFs or built-ins. Since STRING_SPLIT is a relational operator that returns N rows, this property doesn't apply  or it is not relevant.

#please-close

PR #7708, opened 2022-06-20
20220622 polybase configure csv doc

HugoMSFT:
New doc, showing how to virtualize csv doc with SQL Server 2022, the only example we had previously was exclusive to BDC and it was only using GUI.


ShannonLeavitt:
Hi @HugoMSFT  - This pull request was opened in the public repo. PMs should work in the private repo, per the Microsoft Docs contributor guide. Would you make this content update in the private repo instead? Thank you! We'll close this PR.

WilliamDAssafMSFT:
@HugoMSFT I'll take care of it.

PR #7717, opened 2022-06-22
Update sql-agent-extension-automatic-registration-all-vms.md

jasonplante1:
Discussing with Abdullah Mamun on this subject - the job doesn't run daily. It runs monthly. 

With that in mind, the documentation should reflect this, and the line below:

Automatic registration can take up to 4 days to detect newly created SQL Server VMs.

I doubt we can say that it will only take 4 days to pick up a VM with SQL Server installed. The process could take up to 30 depending on when SQL is installed, and when the job runs. 

Currently working a case now where this is the focus of the customer's issue, and again - if the job doesn't run daily, we can't have this out in the ether.

### DIFF
--- a/azure-sql/virtual-machines/windows/sql-agent-extension-automatic-registration-all-vms.md
+++ b/azure-sql/virtual-machines/windows/sql-agent-extension-automatic-registration-all-vms.md
@@ -27,7 +27,7 @@ This article teaches you to enable the automatic registration feature. Alternati
 
 Registering your SQL Server VM with the [SQL IaaS Agent extension](sql-server-iaas-agent-extension-automate-management.md) to unlock a full feature set of benefits. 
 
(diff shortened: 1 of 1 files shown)


Court72:
@adbadram 

#label:"aq-pr-triaged"

adbadram:
We run an Automated job daily to find VM candidates where SQL Server is present, if we don't see SQL Server we will come back and revisit this VM in a month/30 days. Hence daily in this context stands correct.

jasonplante1:
Thanks for that clarification - appreciate the explanation. Adding @Abdullah ***@***.***> for visibility/context.

PRMerger12:
Invalid command: '#sign-off'. Only the assigned author of one or more file in this PR can sign off. @adbadram

PRMerger17:
Invalid command: '#sign-off'. Only the assigned author of one or more file in this PR can sign off. @adbadram

PR #7767, opened 2022-07-06
Adjust wording for user_type_id

mjswart-d2l:
What's a system data type? 
The **is_user_defined** column seems to define it as types that are not user defined. In other words, any type shipped by Microsoft. 
But there are a few system types where **system_type_id** <> **user_type_id**, specifically **hierarchyid**, **geometry**, **geography**, and **sysname**

So it's a mistake to say 
> For system data types, **user_type_id** = **system_type_id**.

(A colleague used `user_type_id = system_type_id` as a filter when they really wanted `is_user_defined = 0`)

### DIFF
--- a/docs/relational-databases/system-catalog-views/sys-types-transact-sql.md
+++ b/docs/relational-databases/system-catalog-views/sys-types-transact-sql.md
@@ -30,7 +30,7 @@ monikerRange: ">=aps-pdw-2016||=azuresqldb-current||=azure-sqldw-latest||>=sql-s
 |-----------------|---------------|-----------------|  
 |**name**|**sysname**|Name of the type. Is unique within the schema.|  
 |**system_type_id**|**tinyint**|ID of the internal system-type of the type.|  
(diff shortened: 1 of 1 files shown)


PRMerger13:
@mjswart-d2l : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mjswart-d2l:
@rwestMSFT Not quite sure about the process on creating an issue and associating it with this PR.

mjswart-d2l:
I suggested adding "most" to the sentence (rather than removing the sentence) because it's still important to point that user_type_ids and system_type_ids are mostly the same for system types.

rwestMSFT:
@mjswart-d2l Thank you for your contribution! I will confer with our team to confirm your change.

rwestMSFT:
@mjswart-d2l After some investigation, I have confirmed that this article needed more detail. I have created an internal PR that will incorporate your suggestions as well as refresh the article, and once that is approved, this PR will automatically close.

Once again, we appreciate your feedback.

PR #7789, opened 2022-07-12
Changing Linux CA folder location

HugoMSFT:
As a CTP 2.1 improvement we changed the path of Linux CA location to /var/opt/mssql/security/ca-certificates this way it is easier for container SQL to persist and reuse the CA.


PRMerger8:
@HugoMSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Issue #7885, opened 2022-08-01
Confusion about SCHEMA_NAME, OBJECT_NAME and DISTRIBUTION


sommarskog:
[Enter feedback here]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: f7ac2fe2-31da-7db6-2d5f-5091022f41d2
* Version Independent ID: bbf07c2c-7006-7b94-f12f-800fa7bc67e4
* Content: [CREATE EXTERNAL TABLE (Transact-SQL) - SQL Server](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-external-table-transact-sql?view=sql-server-ver16&tabs=dedicated)
* Content Source: [docs/t-sql/statements/create-external-table-transact-sql.md](https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/statements/create-external-table-transact-sql.md)
* Product: **sql**
* Technology: **t-sql**
* GitHub Login: @WilliamDAssafMSFT
* Microsoft Alias: **wiassaf**

sommarskog:
Sorry, did not see where I was supposed to type.

On this page, SCHEMA_NAME, OBJECT_NAME and DISTRIBUTION are explained in the text, and they also appear in at least one example(G). But they don't appear in the initial syntax graph. Also, when I try Example G on SQL 2022 CTP 2.1, I get a syntax error. I believe this these three arguments are only valid in Azure SQL Database.

Also, the syntax graph suggests that FILE_FORMAT is mandatory, but that is reasonably not the case when the external data source is a relational data source.

It appears that this topic is in need of some review. Or being split in separate topics, given the differences between platforms.

TimShererWithAquent:
@sommarskog Thanks for your input. We'll look at the issues you found.

WilliamDAssafMSFT:
Appears to be correct on all accounts, thanks @sommarskog. It looks like some of this content belongs in the Azure SQL DB version of the article, thanks for taking the time to point this out.  A doc update is in flight. Thanks again for taking the time to provide this feedback, your suggestions help improve our documentation.

PR #7938, opened 2022-08-14
Update create-external-table-transact-sql.md

sommarskog:
Removed SCHEMA_NAME and OBJECT_NAME, as they only apply to Azure SQL Database. (See also Issue #7885)


Jak-MS:
@WilliamDAssafMSFT - Can you review this PR?

IMPORTANT: When this content is ready to merge, you must add `#sign-off` in a comment or the approval may get overlooked.


#label:"aq-pr-triaged"

WilliamDAssafMSFT:
@sommarskog FYI I was unable to confirm that SCHEMA_NAME and OBJECT_NAME are not supported in SQL Server the last time around. I am still investigating this but at this time I don't have enough information. If you have other confirmation that these two arguments are not supported in _any_ version of SQL Server, please add. 
Closing as I am tracking in an internal effort.

sommarskog:
SCHEMA_NAME and OBJECT_NAME are used in Azure SQL Database, but it is not present in the box product - I get syntax errors when I try to use them. When it comes to Azure Managed Instance, I cannot vouch for anything, since I don't have access to MI. However, SCHEMA_NAME and OBJECT_NAME are not present in the syntax graph for SQL Server.

If my edit would remove that text also from Azure SQL Database, my apologies. But I don't think they should appear in the version for SQL Server.

WilliamDAssafMSFT:
An update: Confirmed that the arguments SCHEMA_NAME and OBJECT_NAME are currently used only for the SQL DB sharing feature. Looks like we do have some places to clean these up in the arguments list. Thanks @sommarskog for pointing this out. A doc update is in flight. Your feedback makes the docs better for everyone, once again!

PR #7942, opened 2022-08-15
Update download-sql-server-management-studio-ssms.md

emilvissing:


erinstellato (approved):
thanks for the catch!

PR #7949, opened 2022-08-16
Include Azure SQL Managed Instance in Applies to

ingebeumer:
As most SQL Server Agent features are supported in Azure SQL Managed Instance, please include MI in Applies to.
 
Kind regards, Inge


Jak-MS:
@ingebeumer - to be able to test the includes, you'll need to use the private repo. Thanks.

PR #7952, opened 2022-08-16
increase Azure SQL Managed Instance visibility

ingebeumer:
increase Azure SQL Managed Instance visibility (favored over Analytics Platform System (PDW) in Applies to section)


Jak-MS:
@ingebeumer - for changes to includes, you'll need to use the private repo. thanks.

PR #8010, opened 2022-08-29

drsqlgithub:
It said you could grant granular permissions to users, but it should be either "database principals" or "users or database roles" so that is clear. I wish the term principals had been pushed harder in 2005 and beyond, but users or roles is probably good enough of a clarification in my mind (even though you can also grant to certificates :))


Jak-MS:
@VanMSFT 
- Can you review this PR?

IMPORTANT: When this content is ready to merge, you must add `#sign-off` in a comment or the approval may get overlooked.


#label:"aq-pr-triaged"

PR #8051, opened 2022-09-07
Update optimize-json-processing-with-in-memory-oltp.md

ph-ict:
Preceding text implies that both columns are persisted, but in the sample code the `Cost` computed column is not defined as `PERSISTED`. Compare to the next code sample down in which both columns are defined as persisted.


rwestMSFT:
@ph-ict Thanks for your feedback. We are refreshing the article in an internal PR, and incorporating your changes. This PR will automatically close when that article is merged.

PR #8053, opened 2022-09-08
correct dmv name

SQLDoubleG:
The query to get supported caches should be [sys].[dm_resource_governor_resource_pools] and not [sys].[dm_os_memory_clerks]


rwestMSFT:
Thanks, @SQLDoubleG. We'll get this merged into an internal PR because the article was in need of a refresh. Look for it in the next day or so.

PR #8061, opened 2022-09-11
Removing ambiguity with date data type and date parameter

greglow-sdu:
The syntax and description as currently written suggests that the parameter data type is a date. That's then confusing when other date-related data types come into use. I think it's important to make it clear that the 2nd parameter is not just a date, and to avoid mixing (throughout the discussion), the concepts of that date parameter and the date data type.

### DIFF
--- a/docs/t-sql/functions/datetrunc-transact-sql.md
+++ b/docs/t-sql/functions/datetrunc-transact-sql.md
@@ -1,6 +1,6 @@
 ---
 title: "DATETRUNC (Transact-SQL)"
-description: "Transact-SQL reference for the DATETRUNC function. This function returns an input date truncated to a specified datepart."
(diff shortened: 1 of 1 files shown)


Court72:
@aashnabafna-ms 

#label:"aq-pr-triaged"

rwestMSFT:
Greg, unfortunately we can't unilaterally change the parameter name in a function like this. However, I will add a note to the Remarks section that makes it more clear. I'll adopt this public PR into an internal update so that you get appropriate credit for your contribution.

rwestMSFT:
@greglow-sdu Thanks again for your contribution. Your feedback has been incorporated into an updated article that will go live later today. I reviewed every word and checked the formatting to make sure the ambiguity is removed, and your PR will be closed automatically when it's signed off.

PR #8071, opened 2022-09-13
Update sql-server-backup-to-url-best-practices-and-troubleshooting.md

jasonplante1:
I don't know that this is the best means of reminding readers that the trace flag referenced is only applicable for page blobs, and not block blobs, but the number of times lately I've had customers AND customer engineers/CSAs point to that trace flag and then be stunned when they get no additional logging is pretty high. 

Which is what I'd expect knowing that the trace flag in question will only work for page blobs, not block blobs - which is what more and more of our customers have switched into.

### DIFF
--- a/docs/relational-databases/backup-restore/sql-server-backup-to-url-best-practices-and-troubleshooting.md
+++ b/docs/relational-databases/backup-restore/sql-server-backup-to-url-best-practices-and-troubleshooting.md
@@ -60,7 +60,7 @@ Parallel backups to the same blob cause one of the backups to fail with an **Ini
 
 If you're using page blobs, for example, `BACKUP... TO URL... WITH CREDENTIAL`, use the following error logs to help with troubleshooting backup errors:  
   
-Set trace flag 3051 to turn on logging to a specific error log with the following format in:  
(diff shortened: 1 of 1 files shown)


jborsecnik:
@WilliamDAssafMSFT, 

#label:"aq-pr-triaged"

WilliamDAssafMSFT:
Thanks @jasonplante1 for this feedback! Sorry you've had that experience. I have to say, the previous sentence is pretty clear, "If you're using page blobs," but this could get better documented in other places, including the main Trace Flags article. That would help awareness. I'll pull in this commit and add this note in other places with a different internal PR.  After the updated article is published, GitHub will list you as an article contributor. Thanks for bringing this to attention! Your feedback helps improve Microsoft Docs for the whole community.

WilliamDAssafMSFT:
Just an update @jasonplante1, TF3051 definitely only applies to page blobs and there is not a trace flag that applies similarly to block blobs, instead, Extended Events can be used for similar troubleshooting purposes. We are in the process of documenting new recommendations for this. Thanks again for drawing attention to this.

PR #8076, opened 2022-09-14
Update failover-clustering-and-always-on-availability-groups-sql-serv…

tjarosz-stdair:
…er.md

Spelling correction.


Jak-MS:
@tjarosz-stdair - thanks for edit.

PR #8081, opened 2022-09-15
Fix create external table example query

fbeltrao:
Based on my experimentation the example query to create external table isn't correct. The correct syntax is to have the `DATA_SOURCE` in `WITH` section.

See [here](https://docs.microsoft.com/sql/t-sql/statements/create-external-table-transact-sql?view=sql-server-ver16&tabs=dedicated#syntax)


scoriani (approved):

scoriani:
#sign-off

PR #8154, opened 2022-10-04
Update sys-query-store-plan-transact-sql.md

awickham10:
Fixed typo in the "find the reason SQL Server couldn't force a plan via QDS" query (missing the "F" in "FROM").


rwestMSFT:
@awickham10 Thanks Andrew!

#sign-off

PR #8159, opened 2022-10-07
Update loginproperty-transact-sql.md

DanielAdeniji:
Currently reads "Indicates whether the login has expired".

I propose it should read ""Indicates whether the principal's password has expired".

The former is a bit ambiguous.

That is the reader may wonder whether an account can actually have expiration date in SQL Server.

Akin to an account that will auto expire ( disappear ) upon its expiration date.

Personally, my current read is that an account cannot have expiration date.

But, an account's password can have an expiration date.  

It is subtle read.

An implicit or inferred statement.

But, I think the explanation/documentation can be more explicit as to what is expiring.


Jak-MS:
@VanMSFT 
- Can you review this PR?


#label:"aq-pr-triaged"

VanMSFT:
Thanks @DanielAdeniji! I agree, and will approve this.

#sign-off

PR #8270, opened 2022-11-17
Update create-index-transact-sql.md

williamadba:


williamadba:
Just me demo'ing at PASS.

PR #8281, opened 2022-11-19
Add SELECT to auditable statements

cairneym:
In my testing, issuing a SELECT statement that includes columns with sensitivity labels added results in the sensitivity information being added to the audit record for that query. The list of activities that are auditable does not include SELECT which would suggest to readers that the sensitivity labels would not be included in the audit record for a SELECT statement


Madhumitatripathy:
#sign-off

PR #8331, opened 2022-11-30
Change summary to "may be passed" instead of "is"

sean-bloch:
Makes header summary semantically in-line with further details below on "TRY...CATCH constructs do not trap the following conditions"


rwestMSFT:
I agree with your intent. What do you think about "is usually passed" instead?

sean-bloch:
@rwestMSFT , that fits and I do think better acknowledges common behavior. 

I was originally inspired by wording usually seen relating to RFC 2119;( however, even re-referencing that, I think "should" would fit better than "may"...) while your suggestion of "is usually passed" is more beneficial to this case, AND plays nicely with diff-comparisons.  Should I update in the PR

Jak-MS:
@rwestMSFT 
- Does this look good to merge?  If so, can you add `#sign-off` ?
Thanks.


#label:"aq-pr-triaged"

rwestMSFT:
#sign-off

rwestMSFT:
Thanks, @sean-bloch, appreciate the attention to detail you bring!

PR #8344, opened 2022-12-05
Remove extra space

oneoneonepig:
For example, to query the `HumanResources. Employee` table for all employees

=>

For example, to query the `HumanResources.Employee` table for all employees


rwestMSFT:
#sign-off

Thanks @oneoneonepig

PR #8347, opened 2022-12-05
Update dbcc-traceon-trace-flags-transact-sql.md

pmasl:
Clarified 2312 and 9481. These don't enable the user to set a CE model (reads as there's some subsequent choice/action to be made). These actually immediately set the intended CE model.

Also clarified 2446, that somehow was changed to read as applicable to "SQL 2017 and SQL 2022 and later" instead of the norm "SQL 2017 and SQL 2019 and later" - where SQL22 is therefore included. Same as for example 2422, 1800, etc


rwestMSFT:
@pmasl I'm working on another PR which touches the DBCC articles so I'll incorporate some of your changes there. Note that the inclusion of SQL Server 2019 won't be included, however.

pmasl:
@rwestMSFT ack, so that means you're merging this one and then it'll be partially overwritten with your newer PR?

Or I should resubmit with the SQL 2019 reference separately? Bc now it reads as not applicable to SQL 2019, which is incorrect. The proposed form aligns with the remainder of the page (min applicable version in CU + min applicable RTM version)

rwestMSFT:
This PR will be closed by my merge, then you can create a new PR based on the latest version of the article once it's published, which will be in the next day or so.

PR #8354, opened 2022-12-06
Update sys-dm-db-resource-stats-azure-sql-database.md

daschall:
Fix typo (memory instead of CPU)


rwestMSFT:
@daschall Thanks for the edit, Daniel!

rwestMSFT:
#sign-off

PR #8370, opened 2022-12-10
Update percent-character-wildcard-character-s-to-match-transact-sql.md

max-webster:
Clarified that LIKE parameters such as 'Jo%n' and '%ari%' would also match. The current description makes it seem like a single % has to go at the beginning or end of the string.

I also deleted 2 trailing spaces from several lines. I presume those aren't significant.


rwestMSFT:
#assign:rwestMSFT

rwestMSFT:
@max-webster Please sign the Contributor License Agreement as soon as possible, so we can merge your change. I have an internal PR waiting, which includes the edit. If you don't sign the CLA, we cannot give you credit for the change.

max-webster:
Sorry, I was offline all through Christmas and didn't see the CLA requirement. I will track down the required permission at my employer anyway in case I have other substantive suggestions in future.

PR #8376, opened 2022-12-13
Quotes in quoted string

Rob-S:
Inside a double quote-enclosed string, it's the double quotes (not single quotes) that need two quotes, to avoid being confused with the double quotes delimiting the string. For example:

```
create table test (
   "single ' quote" int, -- only need one single quote inside quoted string - it's not the delimiter
   "double "" quote" int -- need two double quotes inside the quoted string, since those are the delimiters
);
```


Rob-S:
Because of the "literal string" in the text, I suspect the text was originally copy-and-pasted from the similar remarks for literal strings - where the delimiter is a single quote - while the delimiter for identifiers is a double quote.

jborsecnik:
@markingmyname, 

#label:"aq-pr-triaged"

rwestMSFT:
@Rob-S You're not wrong about double quotation marks, but this appears to be a transcription error when moving the content over from Books Online.

I believe the text should have read as follows (italics added for emphasis):

> All strings delimited by double quotation marks (`"`) are interpreted as object identifiers. Quoted identifiers do not have to follow the Transact-SQL rules for identifiers. They can be keywords and can include characters not generally allowed in Transact-SQL identifiers. If a single quotation mark (`'`) is part of the literal string, it can be represented by ~double~ *two single* quotation marks (`''`).

That being said, you're not wrong. I think this paragraph needs disambiguation so as not to conflate the effect of single and double quotation marks, so I am going to reach out to the product group to make sure we word this correctly.

#reassign:rwestMSFT

Rob-S:
I believe the entire discussion is about identifiers; nothing really about literal strings. In the last sentence, "_the_ literal string" has nothing to refer back to, so "a literal string" would be grammatically correct. But then the question of how to escape double quotes within a double-quoted identifier does not get answered. With the suggested change:

**If a double quotation mark (") is part of the identifier, it can be represented by two double quotation marks ("").**

 "the identifier" refers back to the identifiers being discussed and it answers that question, which is important information needed for the identifier for the second column name in the example I provided:

```
create table test (
   "single ' quote" int, -- only need one single quote inside quoted string - it's not the delimiter
   "double "" quote" int -- need two double quotes inside the quoted string, since those are the delimiters
);
```

The whole thing appears to have been copied from a similar statement regarding literal strings, which are enclosed in single quotes, but since identifiers are enclosed in double quotes, every where that original statement had mentioned single quotes, it should have been changed to "double quotes".

rwestMSFT:
We're with you on this, @Rob-S. It's an old mistake that carried through from BOL. I've already brought it to the attention of the product group, and we're going to figure out if any other articles need to be updated too.

Rob-S:
Thanks!

Rob-S:
>> we're going to figure out if any other articles need to be updated too.
>> But then the question of how to escape double quotes within a double-quoted identifier does not get answered.

There's probably two cases to look for, when figuring out if any other articles need to be updated too.

1. The paragraph is discussing identifiers, but explains how to escape literal strings _instead_ of identifiers - such as for the ALTER DATABASE SET QUOTED_IDENTIFIER.
2. The paragraph is discussing identifiers first, but omits the info on how to escape literal strings, then discusses literal strings and how to escape them - such as for the SET QUOTED_IDENTIFIER statement for setting the option in a session. This is the change I just added.

rwestMSFT:
@Rob-S We are tracking this item internally ([AB#70883](https://dev.azure.com/msft-skilling/cebd7ef5-4282-448b-9701-88c8637581b7/_workitems/edit/70883)), so this issue is being closed. Thanks again for your feedback. When we have an update to share, we will let you know here. In the meanwhile, please feel free to leave additional feedback.

#reassign:rwestMSFT

PR #8381, opened 2022-12-15
Update checksum-agg-transact-sql.md

richardswinbank:
Adds extract white space to ensure newline for readability in "see also".

(Current effect reads as if `BINARY_CHECKSUM ... OVER`... is a supported window function).


Jak-MS:
@richardswinbank - thanks for your submission!

PR #8383, opened 2022-12-15
Update and rename sssql14-md.md to sssql12-md.md

sqlworldwide:
This should reflect SQL 2012 for compat level 110

### DIFF
--- a/docs/includes/sssql12-md.md
+++ b/docs/includes/sssql12-md.md
@@ -1,0 +1,1 @@
+ SQL Server 2012 (11.x)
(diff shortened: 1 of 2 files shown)


MashaMSFT:
Hi @sqlworldwide , thanks so much for your submission! The include files are used as references and should not be updated. We will review your suggestion to determine if there's a business case but cannot accept this PR at this time. 

Thanks again for helping improve our docs, and hope you have a wonderful rest of your day!
Masha from the SQL Docs team

rwestMSFT:
Hi @sqlworldwide, to follow up on my colleague's comments, we're reworking the include files to have more understandable names, but it is a major endeavour. Stay tuned.

rwestMSFT:
@sqlworldwide We already have an include file for 2012, which happens to be called `sssql11-md.md`.

PR #8425, opened 2022-12-25
regular tables can't be created in ledger DB

emad-almousa:
Remark: Its worth stating that when you create a database with flag ledger=ON, you can ONLY create ledger tables within the database system. If you try to create a regular table within the database and specify LEDGER = OFF this will not work and the following error message will be thrown:

LEDGER = OFF cannot be specified for tables in databases that were created with LEDGER = ON.


jborsecnik:
@VanMSFT, 

Can you review the proposed changes? ***Unless that "Remark:" is standard for this content set, we should use common wording/formatting, such as a Docs Markdown "note".***

#label:"aq-pr-triaged"

VanMSFT:
Thanks, @emad-almousa! I reworded the addition and moved it to an appropriate section.

FYI @Pietervanhove 

#sign-off

PR #8450, opened 2023-01-03
Empty string to "0" is valid for integer

sean-bloch:
Empty string for numeric datatypes results in error. Propose changing "This causes a zero to be inserted into the numeric column" to "This causes a zero to be inserted into the integer column" under "Inserting or modifying data in a column set" In the query/DML operation involving  column set 'cs',  conversion failed when converting from the data type 'nvarchar' to the data type 'numeric' 

This behavior can be seen in CAST:

```sql
SELECT CAST('' AS int)
SELECT CAST('' AS NUMERIC(5,2))
```


Jak-MS:
@WilliamDAssafMSFT 
- Can you review this PR?


#label:"aq-pr-triaged"

WilliamDAssafMSFT:
Hi @sean-bloch, nice catch. You are correct that numeric/decimal data types don't get the NULL->0 translation if omitted in a column set. We should specify that. However the substitution to 0 does apply to other generic "numeric" data types as well: bigint, int, smallint, tinyint, bit, and also applies to float/real. So we need to say a little more here. Will take your commit into consideration for a future update to this page. Thanks!

WilliamDAssafMSFT:
Hi @sean-bloch, you should know a doc update which includes your commit is in flight. Thanks again for taking the time to propose this commit.

PR #8470, opened 2023-01-09
Fix include in Synpase ACID note

richardswinbank:
Missing space/square brackets causes rendering error:

image


Court72:
Looks good; thanks, @richardswinbank

PR #8489, opened 2023-01-13
Fix include syntax

richardswinbank:
Fix syntax for ssazuresynapse-md include in [ORDER section](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-columnstore-index-transact-sql?view=sql-server-ver16#order). Current syntax renders as:

image


Court72:
Looks good; thanks, @richardswinbank

PR #8499, opened 2023-01-16
clarification of requirement

way0utwest:
IF all expressions are NULL explicitly, then error 8133 is returned.


rwestMSFT:
Steve, I'll take this over with an internal PR. Thanks.

#assign:rwestMSFT

way0utwest:
Thanks

PR #8509, opened 2023-01-20
Update log-replay-service-overview.md

neetusingh-ms:
Wrong header for mode.


Court72:
Looks good; thanks, @neetusingh-ms

PR #8537, opened 2023-01-27
Update faq.yml - fix typo

ColinPM:
Sentence started with "it's" instead of "It's" - change the first letter "i" to upper case.


dzsquared:
thanks @ColinPM!
#sign-off

PR #8538, opened 2023-01-27
Update sp-addpublication-transact-sql.md

0x7FFFFFFFFFFFFFFF:
Make it more clear. 

This parameter pertains solely to the manner in which the log reader agent operates. The present explanation inadequately illustrates this and employs apparent fact as the characterization. Consider, if the value is "snapshot", would the statement "Publisher provides output of all log-based transactions" still hold true? Indeed!


Court72:
@markingmyname

Can you review the proposed changes?  Note that the validation error must be resolved before merging.

#label:"aq-pr-triaged"

rwestMSFT:
Thanks, we're merging this change into an internal PR.

PR #8592, opened 2023-02-10
Update resource-limits.md

MaxMelcher:
The azure portal allows to deploy up to 16 TB if you deploy SQL MI with 64 cores - the 5.5 TB is the old value.  Please verify


Court72:
@vladai78

#label:"aq-pr-triaged"

vladai78:
@MaxMelcher the correct limit for the maximum storage in Business Critical service tier for the premium-series hardware generation is 5.5 TB. 16 TB of storage in Business Critical is only supported on the memory optimized premium-series hardware generation. If you tried to create a 16 TB BC instance on premium-series, the request would fail.

What you ran into is actually a bug in the portal - thanks for letting us know!

With this in mind, the proposed docs update is incorrect, 

@Court72 what's the easiest way to reject/cancel this change?

PR #8612, opened 2023-02-17
GZIP and ZIP

ClaudioESSilva:
Use `GZIP` on both points `DECOMPRESS`and `Return Types`

In the 1st part, it's mentioned GZIP but on the `Return Types` it mentions ZIP.

Checked the `COMPRESS` page (https://learn.microsoft.com/en-us/sql/t-sql/functions/compress-transact-sql?view=sql-server-ver16) and it's ok only one mention and it is GZIP.


Court72:
@markingmyname

#label:"aq-pr-triaged"

rwestMSFT:
Thanks for identifying this article for review. Along with your change, a refresh was performed in an internal PR.

This PR will be automatically closed when the internal PR is merged.

PR #8619, opened 2023-02-20
Update sql-server-and-client-encryption-summary.md

brandonpires:
Only edited line 46. The sys.dm_exec_connections DMV doesn't actually have an is_encrypted column, but it does have an encrypt_option column with true/false values. The encrypt_option column ultimately provides a user with the same information about whether a connection is encrypted or not so no further changes are needed.


Court72:
@padmajayaraman

#label:"aq-pr-triaged"

ramakoni1 (approved):
Thanks for submitting the correction.

PR #8621, opened 2023-02-20
Update set-a-database-to-single-user-mode.md

himanshu-ram-punjabi:
Grammatical correction.


Court72:
Looks good; thanks, @himanshu-ram-punjabi

PR #8633, opened 2023-02-23
Fix typo

Sun-Kevin:
Modified 'though' to 'through', which I think it is a typo.


Court72:
Looks good; thanks @sun-kevin

PR #8644, opened 2023-02-24
Linked Server Named instance requires port number specified in SQL on Linux

inarocki:
SQL Server on Linux client library does not discover and utilize SQL Browser to get the port for Named Instance


inarocki:
Opened by mistake

PR #8645, opened 2023-02-24
Eliminate duplicate | wording - Update cardinality-estimation-sql-server.md

ClaudioESSilva:
There was a duplicate statement.
The wording can be adapted for "database scoped configuration"

The red rectangle is a copy of the previous bullet.
The green it's a copy of 2 bullets before. Please confirm but I think that text only applies to the `compatibility level` change and not to the `database scoped configuration`
image


WilliamDAssafMSFT:
Thanks @claudioessilva for taking the time to submit this pull request. Good catch! 

Your commit will be merged and this article update is in flight. After the updated article is published, GitHub will list you as an article contributor. Thanks again for this feedback, your suggestions help improve our documentation.

#sign-off

PR #8647, opened 2023-02-25
Update elastic-query-overview.md

kevinmartintech:


Court72:
Looks good; thanks, @kevinmartintech

PR #8656, opened 2023-02-28
SP name is sp_help_change_feed_table

im-microsoft:
Updated SP name from being sp_change_feed_table (incorrect) to sp_help_change_feed_table (correct)


im-microsoft:
renamed Synapse Link for SQL system SP from sp_change_feed_table to sp_help_change_feed_table

PR #8697, opened 2023-03-09
Add missing space in "format nul"

sqlHippo:
**format nul** is referenced as **formatnul** in several places. Sometimes bcp would successfully create a non-XML format file with the space missing, but it would not create the XML format file, even with the **-x** option.


rwestMSFT:
Great find!

#sign-off

PR #8732, opened 2023-03-14
adding additional info regarding sys.sysxlgns table

emad-almousa:
Hi,
system table sys.sysxlgns contains password hashes, which should be mentioned so admins are aware of this fact.

sys.sysxlgns**|Exists in the **master** database only. Contains a row for each server principal and its corresponding password hash.

Regards,
Emad


VanMSFT:
Hi @emad-almousa! Thanks for adding to our docs! Although what you're adding is correct, there are many more values to that table. We are purposely not adding any additional information to System Base Table, and do not recommend them for general customer use.

image

Unfortunately, I'll deny this one, but please feel free to let us know if you have any feedback. Thanks!

PR #8746, opened 2023-03-20
Not implemented for Elastic Query

cairneym:
After feedback from the Product Team (Vlado Tosev) I got confirmation that this is not implemented with Elastic Query. In addition to the removal of Elastic Query, I suggest that the 2 images that show SQL Servers as the FQDN are also changed to only be Blob Storage Endpoint FQDNs


Court72:
@rohitnayakmsft

#label:"aq-pr-triaged"

rohitnayakmsft:
#sign-off

PR #8750, opened 2023-03-20
use SINGLE_BLOB to import XML, rather than CLOB

jnahmias:
Per the guidance in the OPENROWSET() documentation, use `SINGLE_BLOB` when loading XML. ref: https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/functions/openrowset-transact-sql.md#single_blob


rwestMSFT:
Great eye, and thanks for the feedback, @jnahmias.

PR #8755, opened 2023-03-21
Update database-files-and-filegroups.md

asavioliMSFT:
primary word is missing when describing the primary filegroup


WilliamDAssafMSFT:
Thanks @asavioli for taking the time to submit this PR! Good catch. Your commit will be merged and this article update is in flight. After the updated article is published, GitHub will list you as an article contributor. 

#sign-off

PR #8763, opened 2023-03-24
Update at-time-zone-transact-sql.md

JediSQL:
Modify Example B so as to highlight that an argument for the "timezone" parameter can be provided as an expression other than a string literal.

PR #8764, opened 2023-03-25
Update querying-data-in-a-system-versioned-temporal-table.md

JediSQL:
Modified the "AS OF" example to demonstrate the value of using AT TIME ZONE to convert a local time zone value for the SYSTEM_TIME expression. also added AT TIME ZONE for Next Steps (please check I did that correctly).

### DIFF
--- a/docs/relational-databases/tables/querying-data-in-a-system-versioned-temporal-table.md
+++ b/docs/relational-databases/tables/querying-data-in-a-system-versioned-temporal-table.md
@@ -79,9 +79,11 @@ LEFT JOIN [dbo].[LocationDepartments]
 LEFT JOIN [dbo].[Department]
    ON LocationDepartments.DeptID = [Department].DeptID;
 GO
-/* Querying view AS OF */
+/* Querying view AS OF with local time*/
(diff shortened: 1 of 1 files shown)


rwestMSFT:
@JediSQL Thanks - if you can break that out into a completely separate example, I think it'll be an even better contribution!

Court72:
@rwestMSFT 

Thanks for reviewing the proposed changes. 

#label:"aq-pr-triaged"

rwestMSFT:
@JediSQL I am closing this PR, because we identified that both articles you reference (see #8763) needed refreshment. Your changes will be incorporated into the refresh. You can expect to see the updated articles in the next day or so.

rwestMSFT:
@JediSQL Thanks - I'll still comment here once I've refreshed all the articles in this area, because some of them haven't been looked at in a while.

rwestMSFT:
@JediSQL We ran into an issue with the syntax and I've had to roll these changes back. I'll comment here once I've refreshed all the articles in this area, because some of them haven't been looked at in a while.

PR #8782, opened 2023-03-29
SA password change when lost the password

qxsch:
In this scenario, you have lost/forgotten your password for the **SA** account and have to set a new one. When using a dev container, that lies around, this can happen.


American-Dipper:
This PR, which does not seem to include any actual change, targets the live branch, which is not allowed.

PR #8786, opened 2023-03-30
Create Push Subscription - RMO Documentation updates

jpomfret:
Working on using RMO to create subscriptions and I found that I needed to set the `WindowsAuthentication` setting to False for the change to be picked up.

Following the docs I ended up with True still for WindowsAuth and it didn't show the change in SSMS\replication monitor. The second results set shows setting it to False.
image

Then it shows up here - believe that true\false is that radio button.
image

It's also on [create-a-pull-subscription](https://learn.microsoft.com/en-us/sql/relational-databases/replication/create-a-pull-subscription?view=sql-server-ver16) but I haven't tested that yet - I can test and get back to you on this.


ShannonLeavitt:
@MashaMSFT 

#label:"aq-pr-triaged"

MashaMSFT:
@jpomfret , thank you so much for helping improve our documentation! #sign-off

PR #8792, opened 2023-04-01
paragraph update

emad-almousa:
in section "Limitations and restrictions":
 In addition, its worth stating that you can’t drop procedures owned by “sys” schema even with accounts granted SYSADMIN role.


WilliamDAssafMSFT:
Thanks @emad-almousa for contacting Microsoft with feedback about our product documentation. I believe this is sufficiently understood. Thanks for submitting this feedback, your suggestions help improve our documentation.

PR #8806, opened 2023-04-04
Remove the IsAlive /LooksAlive ref in the table and updates to lease timeout

PiJoCoder:
These are not meaningful for lease timeout and session timeout and leads customers to confusion. Also they don't give practical application. So best is to remove them


PiJoCoder:
#hold-off - more changes coming

PR #8814, opened 2023-04-05
Updated Scale value : precision > 38

UttamKumarReddy:
According to document The resulting scale is reduced to min(scale, 38 - (precision-scale)) but it is applicable when precision is grater than 38. So this would result in min(scale , (38-precision) + scale) and 38 - precession is always negative and 38 - (precision-scale) will be new scale . So Instead of  min(scale, 38 - (precision-scale)) 38 - (precision-scale) can be used.


Jak-MS:
@MikeRayMSFT 
- Can you review this PR?


#label:"aq-pr-triaged"

rwestMSFT:
@UttamKumarReddy While it is true that if the min precision is greater than 38, your suggested edit is correct. However, if the precision is less than 38, your logic doesn't seem to hold up.

Can you provide more insight into your proposed edit so that I can understand it better?

rwestMSFT:
@UttamKumarReddy I took a look at the source code and can confirm that the documentation is correct as currently written.

Thank you for your contribution. Although we decided not to use it, you did give us the opportunity to refresh the article and make it easier to read. The updated article will be available in the next day or so.

UttamKumarReddy:
@rwestMSFT We will get into the 3 cases provided in the documentation only if precision is greater than 38 only. If the precision is less than 38, we just us result precision is p1+p2+1 and scale is s1+s2 and we do not enter to those rules. So, min precision can never be less than 38 because if it is less than 38 if do not get into these rules.

Also, Can I get a look into the source code?

rwestMSFT:
Please refer to the first example in the article, where the integral part is < 32. This is the exact scenario you are arguing against. The text below that example describes the calculation as documented, which is confirmed in the source code.

Needless to say, I am not at liberty to provide access to SQL Server source code, no matter how nicely anyone asks.

Thanks again for your feedback.

PR #8824, opened 2023-04-06
Update connectivity-architecture-overview.md

zoran-rilak-msft:
Fix public port


zoran-rilak-msft:
Will do this against sql-docs-pr instead

PR #8853, opened 2023-04-15
Changed unique index advice to unique constraint

greglow-sdu:
In the advice given here, if you need to enforce uniqueness for a column that has a sequence value, the first choice should be a unique constraint, not a unique index i.e. it should be an attribute of the table design, not an index external to the table.

### DIFF
--- a/docs/relational-databases/sequence-numbers/sequence-numbers.md
+++ b/docs/relational-databases/sequence-numbers/sequence-numbers.md
@@ -42,7 +42,7 @@ monikerRange: "=azuresqldb-current||>=sql-server-2016||>=sql-server-linux-2017||
 ## Limitations  
  Unlike identity columns, whose values cannot be changed, sequence values are not automatically protected after insertion into the table. To prevent sequence values from being changed, use an update trigger on the table to roll back changes.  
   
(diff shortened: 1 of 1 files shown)


Court72:
@MashaMSFT

#label:"aq-pr-triaged"

rwestMSFT:
Greg, since a unique constraint is implemented as a unique index anyway, do you not consider this a semantic difference?

greglow-sdu:
Hi Randolph,

Not, not at all. Regardless of the implementation, uniqueness is a business/logical requirement that should be part of the table design, and an attribute of the table, not an attribute of an external object like an index.

In general, I think that writing CREATE UNIQUE INDEX is a design flaw, unless you are specifically working around one of the limitations of how SQL Server implements uniqueness, which are unlikely to apply here.

We should be encouraging better design in the docs.

Regards,

Greg

Dr Greg Low

1300SQLSQL (1300 775 775) office 
SQL Down Under | Web: https://sqldownunder.com<https://sqldownunder.com/> | About Greg:  https://about.me/greg.low

MashaMSFT:
Hi @greglow-sdu , thanks so much for your submission! #sign-off

PR #8857, opened 2023-04-17
Update sys-dm-hadr-availability-replica-cluster-states-transact-sql.md

0x7FFFFFFFFFFFFFFF:
From my actual server, the join_state_desc is "JOINED_FCI".


rwestMSFT:
Thanks for this. I get the feeling this was changed since the article was written, so let me investigate.

rwestMSFT:
We have an internal PR that includes this change, which you'll see published in the next day or so. Thank you for your contribution.

PR #8858, opened 2023-04-17
Update sys-dm-hadr-database-replica-cluster-states-transact-sql.md

0x7FFFFFFFFFFFFFFF:
Fix error. You should query dm_hadr_database_replica_cluster_states to answer the following questions, not dm_hadr_database_replica_states.


rwestMSFT:
This will be incorporated into the internal PR we have for #8857.

PR #8862, opened 2023-04-18
Update sql-database-project-extension-sdk-style-projects.md

kchenery:
Correct typo in SQL script

BEING should have been BEGIN


Court72:
@kchenery - thanks for the typo fix!

PR #8867, opened 2023-04-19
Update sp-add-category-transact-sql.md

cafealternativo:
The store procedure can be executed by users not belonging to the sysadmin fixed role by being explicitly granted exec permissions. Example:

USE MSDB; 
GO
GRANT EXEC ON dbo.sp_add_category TO <USER>
GO

I am using the same wording used here: https://learn.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-delete-backuphistory-transact-sql#permissions

Please confirm this is the intended behavior with Product Group.

### DIFF
--- a/docs/relational-databases/system-stored-procedures/sp-add-category-transact-sql.md
+++ b/docs/relational-databases/system-stored-procedures/sp-add-category-transact-sql.md
@@ -67,7 +67,7 @@ sp_add_category
  **sp_add_category** must be run from the **msdb** database.  
   
 ## Permissions  
- Only members of the **sysadmin** fixed server role can execute **sp_add_category**.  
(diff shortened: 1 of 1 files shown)


rwestMSFT:
@cafealternativo A couple things:

1. Assigning explicit permissions, which should only be considered as part of a broader security posture, is outside of the scope when discussing default permissions for SQL Server system objects. I don't think the PRs you submitted warrant a specific change to the wording for these three articles, because the same could be argued for *all* of them. That's the kind of change that is definitely not in scope.

2. Microsoft employees are encouraged to use the private repo to submit changes, per the contributor guide.

I'm closing your PRs, and you're welcome to engage with us internally to continue this discussion.

PR #8868, opened 2023-04-19
Update sp-delete-category-transact-sql.md

cafealternativo:
The store procedure can be executed by users not belonging to the sysadmin fixed role by being explicitly granted exec permissions. Example:

USE MSDB;
GO
GRANT EXEC ON dbo.sp_delete_category TO USER
GO

I am using the same wording used here: https://learn.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-delete-backuphistory-transact-sql#permissions

Please confirm this is the intended behavior with Product Group.


rwestMSFT:
Closed with #8867.

PR #8879, opened 2023-04-25
Corrected possible causes and fixes for the error

VladDBA:
Since this error pertains to Shared Memory, the original explanation stating that "the instance is not configured for remote connections" was incorrect due to shared memory being a local protocol (meaning that the client is on the same host as the instance it's trying to connect to).

### DIFF
--- a/docs/relational-databases/errors-events/mssqlserver-233-database-engine-error.md
+++ b/docs/relational-databases/errors-events/mssqlserver-233-database-engine-error.md
@@ -30,14 +30,12 @@ helpviewer_keywords:
 The [!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)] client cannot connect to an instance residing on the same host as the client. 
  
 ### Potential causes:
-[!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)]
 - Unable to connect through the Shared Memory protocol.
 - The instance has reached its maximum concurrent user connections limit
   
 ## User Action 
-[!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)]
 Depending on the cause:
-- Try to disable the Shared Memory protocol in Configuration Manager and use TCP/IP instead.
+- Try to disable the Shared Memory protocol in [!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)] Configuration Manager and use TCP/IP instead.
 - Restart the instance in single user mode and make sure that the maximum number of concurrent user connections has not been changed from its default value of 0 (0 = unlimited)
   
 ## See Also  
--- a/docs/relational-databases/errors-events/mssqlserver-233-database-engine-error.md
+++ b/docs/relational-databases/errors-events/mssqlserver-233-database-engine-error.md
@@ -26,15 +26,25 @@ helpviewer_keywords:
 |Symbolic Name||  
 |Message Text|A connection was successfully established with the server, but then an error occurred during the login process. (provider: Shared Memory Provider, error: 0 - No process is on the other end of the pipe.) (Microsoft SQL Server, Error: 233)|  
   
-## Explanation  
-The [!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)] client cannot connect to the server. This error could occur because the server is not configured to accept remote connections.  
+## Explanations  
+The [!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)] client cannot connect to an instance residing on the same host as the client. 
+ 
+### Potential causes:
+[!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)]
+- Unable to connect through the Shared Memory protocol.
+- The instance has reached its maximum concurrent user connections limit
   
-## User Action  
-Use the [!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)] Configuration Manager tool to allow [!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)] to accept remote connections.  
+## User Action 
+[!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)]
+Depending on the cause:
+- Try to disable the Shared Memory protocol in Configuration Manager and use TCP/IP instead.
+- Restart the instance in single user mode and make sure that the maximum number of concurrent user connections has not been changed from its default value of 0 (0 = unlimited)
   
 ## See Also  
-[Network Protocols and Network Libraries](~/sql-server/install/network-protocols-and-network-libraries.md)  
+[Network Protocols and Network Libraries](~/sql-server/install/network-protocols-and-network-libraries.md)
+[Shared Memory Properties](~/tools/configuration-manager/shared-memory-properties.md)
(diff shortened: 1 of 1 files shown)


VladDBA:

VladDBA:
Hello,
I've just realized that I've left some `[!INCLUDE[ssNoVersion](../../includes/ssnoversion-md.md)]` elements in my changes that I should have removed.
If It's not ok, I can redo the pull request.

Court72:
@MashaMSFT

#label:"aq-pr-triaged"

MashaMSFT:
Hi @VladDBA , thanks for bringing this to our attention! After investigating this further, it seems the documentation for error 233 is actually incorrect. We've created a backlog item to get this addressed. However, unfortunately, we were unable to reproduce the behavior you're describing, nor do your potential causes and resolutions seem correct. As such, my apologies, but we're going to have reject this PR. 

We're working on getting this fixed though so please stay tuned. 

Thanks again for reaching out, and I hope you have a wonderful rest of your day!
Masha from the SQL Docs team

MashaMSFT:
Hi @VladDBA , if you're able to, it may be good to open a support case with Microsoft so we're better able to understand the circumstances that are leading to the error that you're seeing, as we're struggling to reproduce it on our end. 

Thanks!
Masha from the SQL Docs team

VladDBA:
Hi @MashaMSFT,

Thank you for the input.
Sorry about that, I realized now that I was too sparse on the information initially.
### Steps to reproduce:
1. On a machine with SQL Server installed enable, if it's not already enabled, the Shared Memory protocol for the instance (now local connections will tend to use that as opposed to TCP/IP).
2. Set user connections to a low number like 2.
```SQL
EXEC sys.sp_configure N'show advanced options', N'1';
GO
RECONFIGURE WITH OVERRIDE;
GO
EXEC sys.sp_configure N'user connections', N'2';
GO
RECONFIGURE WITH OVERRIDE;
GO
EXEC sys.sp_configure N'show advanced options', N'0';
GO
RECONFIGURE WITH OVERRIDE;
GO
```
3. Restart the instance
4. Fire up one or two connections via sqlcmd to make sure those 2 connections are used up
5. Attempt to connect  the instance via SSMS from the same host

- At this point the error message returned by SSMS and which links to this docs page is exactly like this one 

- There's another variation of this error which is reproducible when SSMS connects ok initially and Object Explorer is also connected, but the query editor (either the first one opened or any subsequent tab) will be left with no available connection slots to use, in this case the error message will be this one:

This can be reproduced by restarting the SQL Server instance, skipping step 4, and immediately connecting with SSMS. When the connection is established and OE is displayed,  just open query editor tabs until the connection limit is reached.

- Attempting to connect remotely or via IP/TCP to the instance will result in the following error message with 64 error code

And in one of my tests, connecting remotely resulted in an 10053 error code, but I only got this one out of 10 tries.


### Fixing it
In my scenario, the users changed the user connections limit and were unable to get back on the instance, so the fix was to stop SQL Server, start it in single user mode, connect via sqlcmd as soon as the instance was up, and set the limit back to 0.
Example using administrative command prompt:
```
net stop "SQL Server (SQL2019)"
net start "SQL Server (SQL2019)" /m"SQLCMD" && sqlcmd -S localhost\sql2019 -E
```
Once sqlcmd is connected to the instance:
```SQL
EXEC sys.sp_configure N'show advanced options', N'1';
GO
RECONFIGURE WITH OVERRIDE;
GO
EXEC sys.sp_configure N'user connections', N'0';
GO
RECONFIGURE WITH OVERRIDE;
GO
EXEC sys.sp_configure N'show advanced options', N'0';
GO
RECONFIGURE WITH OVERRIDE;
GO
```
And then restart SQL Server:
```
net stop "SQL Server (SQL2019)"
net start "SQL Server (SQL2019)"
```

Obviously reaching the max connections limit is just one of the potential causes, insufficient resources for SQL Server to allocate for a new connection can also lead o this. In this case DAC will be needed to successfully connect to the instance and troubleshoot without restarting it in single user mode.

### Test environment:
Microsoft SQL Server 2019 (RTM-CU19) (KB5023049) - 15.0.4298.1 (X64) 
	Jan 27 2023 16:44:09 
	Copyright (C) 2019 Microsoft Corporation
	Developer Edition (64-bit) on Windows Server 2022 Standard Evaluation 10.0 <X64> (Build 20348: ) (Hypervisor)


I hope this helps clear things up.

Thank you and have a wonderful day!

Vlad

VladDBA:
Hi @MashaMSFT,

I forgot to mention that the URL provided by the error message for code 64 leads to a page that doesn't exist yet.
```
TITLE: Connect to Server
------------------------------

Cannot connect to Winsrv2k22\SQL2022.

------------------------------
ADDITIONAL INFORMATION:

A connection was successfully established with the server, but then an error occurred during the pre-login handshake. (provider: TCP Provider, error: 0 - The specified network name is no longer available.) (Microsoft SQL Server, Error: 64)

For help, click: https://docs.microsoft.com/sql/relational-databases/errors-events/mssqlserver-64-database-engine-error

------------------------------

The specified network name is no longer available

------------------------------
BUTTONS:

OK
------------------------------
```

Regarding opening a support case: I don't think it's that big of a deal to warrant bothering the MS support team, and it's easily fixable. I've also put up [a blog post](https://vladdba.com/2023/04/26/fixing-sql-server-connection-errors-64-233/) to help out anyone else that might be dealing with this scenario. 

Thanks!

Vlad

MashaMSFT:
Hi Vlad, this is really helpful, thanks a bunch! Just to close the loop for you and anyone else who may stumble on this, error 233 in the product is actually: The column '%.*ls' in table '%.*ls' cannot be null. 
So the documentation is definitely wrong :) we're working to fix that. 

The error 233 you're seeing is actually a [Windows error](https://aka.ms/errors):  No process is on the other end of the pipe.

Hope that helps and once again, really appreciate you bringing this to our attention so we can fix the documentation!
Masha from the SQL Docs team

VladDBA:
Masha, thank you for clearing this up!
Does this mean that the error message details provided by SSMS will also need to be updated since the links they provide point to the wrong documentation pages?
For reference, this is the link provided in the details for error code 233
```
ADDITIONAL INFORMATION:

A connection was successfully established with the server, but then an error occurred during the pre-login handshake. (provider: Shared Memory Provider, error: 0 - No process is on the other end of the pipe.) (Microsoft SQL Server, Error: 233)

For help, click: https://docs.microsoft.com/sql/relational-databases/errors-events/mssqlserver-233-database-engine-error
```

VladDBA:
Hi @MashaMSFT,

Can you please answer my last question about the URL where SSMS is pointing when giving out the error details?
Because if it needs to point to the Windows error URL (which it looks like it should, but I'd rather have confirmation) I'll add it as [feedback](https://feedback.azure.com/d365community/forum/04fe6ee0-3b25-ec11-b6e6-000d3a4f0da0) for SSMS.

Thank you and have a great weekend!

Vlad

MashaMSFT:
Hi @VladDBA , thanks again for reaching out! Yup, we've filed a bug with the SSMS PG about it but please feel free to file it as well to increase the visibility :)

Thanks!
Masha from the SQL Docs team

PR #8901, opened 2023-05-03
Fixing a few typos

greglow-sdu:
Just fixing a few typos in two code examples


Court72:
Looks good; thanks, @greglow-sdu

PR #8937, opened 2023-05-10
Add language mode to the code example

Espand-Masoon:


rwestMSFT:
Thanks for your contribution. This change will be incorporated into an article refresh, which will publish later this week. This PR will automatically close when that change is merged.

PR #8978, opened 2023-05-24
Fix Misspelling in insert-transact-sql.md

jamesbascle:
vaule => value


ShannonLeavitt:
Looks good. Thanks @jamesbascle

PR #8986, opened 2023-05-25
Update configure-firewall-rules-before-running-the-tsql-debugger.md

sommarskog:
Fixed spelling error.


sommarskog:
But reading that paragraph that I edited, I am not sure that it is correct. "SQL Server Management Studio or SQL Server Data Tools must be running under a Windows account that is a member of the sysadmin fixed server role." But the sysadmin fixed server role is an SQL Server thing. Don't we want to talk about Windows permissions here? 

My interest for this topic was spawned by a Q&A thread where  the poster laments that he is not a local admin on his Windows box, and therefore cannot use the debugger. See https://learn.microsoft.com/en-us/answers/questions/1290827/store-procedure-debugging-in-visual-studio.

I should add that I don't use the debugger myself. Too much hassle to get it to work.

ShannonLeavitt:
@markingmyname 

#label:"aq-pr-triaged"

rwestMSFT:
@sommarskog Thanks Erland, we'll incorporate this into a refresh of the article.

PR #8991, opened 2023-05-29
Update managing-services-and-network-settings-by-using-wmi-provider.md

yochananrachamim:
Changed "applies to" to SQL Server only, the cloud services, Azure SQL DB, Managed instance and Synapse does not have RPC interface and do not support WMI remote connection at all.


American-Dipper:
@yochananrachamim - The path you've used doesn't work. It looks like maybe the ../ you removed needs to be readded, as the file is at the same level of the one it replaced. (Note also that Microsoft staff are generally expected to work in the private repo, not the public.)

@markingmyname - Can you reivew / approve?

#label:"aq-pr-triaged" 

PR #9010, opened 2023-06-01
Update execution-plans.md

pituach:
clarify that the execution plan (EP) is not neccererily "the most efficient", but the best between the EP which were created. I came here as a result of a question in the forum. The current description confuse people to think that there is no reason to monitor executions and improve queries using hints for example, since the server already always create "the most efficient" EP.

### DIFF
--- a/docs/relational-databases/performance/execution-plans.md
+++ b/docs/relational-databases/performance/execution-plans.md
@@ -18,7 +18,7 @@ helpviewer_keywords:
 # Execution Plans
 [!INCLUDE [SQL Server Azure SQL Database Azure SQL Managed Instance](../../includes/applies-to-version/sql-asdb-asdbmi.md)]
 
-To be able to execute queries, the [!INCLUDE[ssDEnoversion](../../includes/ssdenoversion-md.md)] must analyze the statement to determine the most efficient way to access the required data. This analysis is handled by a component called the Query Optimizer. The input to the Query Optimizer consists of the query, the database schema (table and index definitions), and the database statistics. The output of the Query Optimizer is a query execution plan, sometimes referred to as a query plan, or execution plan.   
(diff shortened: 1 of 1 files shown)


pituach:
Small typo in the description: 
"best between the **RP** which were created." -> "best between the **EP** which were created."

rwestMSFT:
@pituach Thanks, Ronen! I'll incorporate your feedback into an internal PR, as it looks like this article hasn't been updated in 3 years and could do with a refresh anyway.

Jak-MS:
@rwestMSFT - if these have been incorporated, this PR should be closed. thanks.

Jak-MS:
#label:"aq-pr-triaged"

rwestMSFT:
@Jak-MS This PR will be closed when the internal PR is merged, as per our usual process.

pituach:
Hi @rwestMSFT (Randolph West),

There is a tiny issue with this approach which I have just noticed. 
As a former MVP you should understand the implication better than others.

You are using my editing contents as part of another PR, which makes sense probably, but this means that all my the activities are not documented. For example in the current document I see that you embed my fix.
https://learn.microsoft.com/en-us/sql/relational-databases/performance/execution-plans
image

But in the list of contributors, my name does not appears since my content was added under another account.
image

It's happens a lot when we improve documents and the docs team choose to merge the improvement to another PR.
👆 Just something that someone should think about probably😀

Have a great day

rwestMSFT:
@pituach Your contributions are still included in the commit history of the article, even if you don't always show as a contributor. When we do merges between repositories, the commit history is sometimes squashed. You'll see your name here, and it still counts.

Observe: https://github.com/MicrosoftDocs/sql-docs/commit/50c7d54aca61b6afa2b4a770054bdf803cac58c6

PR #9019, opened 2023-06-02
Update sp-changedistributor-password-transact-sql.md

taiyebzakir:
Added this:
If this is a remote distributor, then it needs to be run on all the Publisher Servers that is using this Distributor. If Distribution or publisher database is in AG, then it needs to be run on all the Distributor and Publisher nodes. It does not matter if the node is primary or secondary.


Court72:
@markingmyname

#label:"aq-pr-triaged"

rwestMSFT:
@taiyebzakir Thanks, we're moving this to an internal PR to refresh the article too. Please take a look at https://github.com/MicrosoftDocs/sql-docs-pr/pull/27343.

PR #9052, opened 2023-06-12
Update ssis-catalog.md

asavioliMSFT:
If customers use automatic seeding or manual backup/restore method + join only in the wizard, the error described in the article below happens (I could reproduce the same issue on different SQL versions, 2017 and 2019). The only way to fix the issue is using the Full database and log backup method when adding the database to the AG.

https://techcommunity.microsoft.com/t5/sql-server-support-blog/ssis-always-on-ag-availability-group-and-error-please-create-a/ba-p/414236

Because of that, I think we must mention that in the article


Court72:
@chugugrace

Can you review the proposed changes? Note that the validation error must be resolved before merging.

#label:"aq-pr-triaged"

PR #9060, opened 2023-06-14
Update aad-security-configure-tutorial.md

miljann995:
Added a note that describes limitations when an Azure AD user, that is part of the Azure AD group which is used as the Azure AD admin on the SQL Managed Instance, logs in to SQL Managed Instance. In this case, the `SUSER_SNAME` call returns the Azure AD user's name, but the `SUSER_SID` call returns `NULL`, since the given Azure AD user is not listed in the `sys.server_principals` table. Therefore, access to certain functionalities may be restricted.

### DIFF
--- a/azure-sql/managed-instance/aad-security-configure-tutorial.md
+++ b/azure-sql/managed-instance/aad-security-configure-tutorial.md
@@ -220,6 +220,11 @@ For more information on granting database permissions, see [Getting Started with
 
 ### Create an Azure AD user and create a sample table
 
+> [!NOTE]
+> There are some limitations when the Azure AD user, that is part of the Azure AD group used as the Azure AD admin on SQL Managed Instance, logs in to SQL Managed Instance.
+> For example, a call to `SUSER_SID` returns `NULL`, since the given Azure AD user is not part of the `sys.server_principals` table.
+> Therefore, access to certain stored procedures or a list of granted permissions may be limited in this case.
+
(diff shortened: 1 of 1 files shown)


Court72:
@GithubMirek

#label:"aq-pr-triaged"

azure-sql/managed-instance/aad-security-configure-tutorial.md
```diff
@@ -220,6 +220,11 @@ For more information on granting database permissions, see [Getting Started with
 
 ### Create an Azure AD user and create a sample table
 
+> [!NOTE]
+> There are some limitations when the Azure AD user, that is part of the Azure AD group used as the Azure AD admin on SQL Managed Instance, logs in to SQL Managed Instance.
```
rwestMSFT:
```suggestion
> There are some limitations when the Azure AD user, which is part of the Azure AD group used as the Azure AD admin on SQL Managed Instance, signs into SQL Managed Instance.
```

VanMSFT:
@nofield Adding Jordan for review. Thanks!

nofield:
Hi @miljann995, thanks for this suggestion. I wholeheartedly agree that this limitation should be documented- I have a work item to create a limitations section about Azure AD groups, as there are a couple quirks around it. I'm good with this suggestion for now, and will roll it into that more formal section when I get to it.

nofield (approved):

nofield:
#sign-off

PR #9086, opened 2023-06-20
linux-odbc: Simplify instructions for Debian/Ubuntu

orgads:
+ update supported versions.


Court72:
@David-Engel

#label:"aq-pr-triaged"

David-Engel (commented):
@v-ancha This is similar to the change you are working on.

1. We want to standardize on wget, not curl. See the installing SQL Server on Linux instructions for reference.
2. For Debian, include a condition to error if running an unsupported version (like we have for Ubuntu).
3. We don't yet support Ubuntu 23.04 so we shouldn't be changing that if block in the Ubuntu instructions.

CC: @v-johoang @v-chojas

David-Engel:
Similar changes have been incorporated into the linux install instructions now, making this PR obsolete. Thanks for the suggestions.

David

#please-close

PR #9094, opened 2023-06-21
Update examples-of-bulk-import-and-export-of-xml-documents-sql-server.md

DanielAdeniji:
Added a third example to offer a template for inserting data into a multi-column table.

In this case, the column names for the target table and the results of the OPENROWSET are explicitly indicated.

### DIFF
--- a/docs/relational-databases/import-export/examples-of-bulk-import-and-export-of-xml-documents-sql-server.md
+++ b/docs/relational-databases/import-export/examples-of-bulk-import-and-export-of-xml-documents-sql-server.md
@@ -133,6 +133,59 @@ GO
 ```  
   
  [&#91;Top&#93;](#top)  
+
+
+ #### Example C  
+ This example uses the `SINGLE_BLOB` option in an `INSERT ... SELECT * FROM OPENROWSET(BULK...)` statement to import data from a file named `SampleData3.txt`.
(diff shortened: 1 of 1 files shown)


rwestMSFT:
@DanielAdeniji Thank you. I'll include this in an internal PR. The underlying article is in need of refreshing.

DanielAdeniji:
Thank You,

Daniel

rwestMSFT:
Upon review, I don't see the need for an entirely new example, because it is an alternative to Example A, with columns specified.

I will add a variation of your code sample under Example A.

Thanks again for your contribution.

DanielAdeniji:
Randolph:-

I think Example A lacks broad-based application.

It only covers a single-use code.  That of adding data to a table with a single column.

Example b follows behind on how to update that single column.

Example B updates records where IntCol is 1


WHERE IntCol = 1;

But, none of the examples have IntCol equal to 1.

That is:-

  1.  IntCol is not explicitly defined as "not null" column
  2.  IntCol is not defined as an identity column

I think setting IntCol to 1 within the insert statement will further readability.

Thank You,

Daniel

rwestMSFT:
I agree setting the INT column as IDENTITY is a good idea, and have done that.

PR #9099, opened 2023-06-22
Update high-availability-sla.md -- Duplicate text on line 77

wesleyf1984:
Duplicated text on line 77.

Removed duplicated text.

"Each availability zone is a separate physical location with independent power, cooling, and networking."


rwestMSFT:
Thanks, @wesleyf1984! We appreciate your contribution.

PR #9108, opened 2023-06-26
Update sp-add-schedule-transact-sql.md

andreasjordan:
Align with source code of msdb.dbo.sp_verify_schedule.

And here is the relevant code from msdb.dbo.sp_verify_schedule:
```
  IF ((@freq_subday_type <> 0x1) AND (@freq_subday_interval < 1)) -- FREQSUBTYPE_ONCE and less than 1 interval
     OR
     ((@freq_subday_type = 0x2) AND (@freq_subday_interval < 10)) -- FREQSUBTYPE_SECOND and less than 10 seconds (see MIN_SCHEDULE_GRANULARITY in SqlAgent source code)
  BEGIN
    SELECT @reason = FORMATMESSAGE(14200, '@freq_subday_interval')
    RAISERROR(14278, -1, -1, @reason)
    RETURN(1) -- Failure
  END
```

So 10 seconds is allowed, shorter periods raise an error. I'm not a native english speaker, so please adjust the wording as you like. I just think that "should" is not correct as it is just not possible and 10 seconds is a valid option.

### DIFF
--- a/docs/relational-databases/system-stored-procedures/sp-add-schedule-transact-sql.md
+++ b/docs/relational-databases/system-stored-procedures/sp-add-schedule-transact-sql.md
@@ -98,7 +98,7 @@ Specifies the units for *@freq_subday_interval*. *@freq_subday_type* is **int**,
 
 #### [ @freq_subday_interval = ] *freq_subday_interval*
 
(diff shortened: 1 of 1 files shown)


markingmyname (approved):

markingmyname:
#sign-off

PR #9128, opened 2023-07-05
Update xp-cmdshell-server-configuration-option.md

VladDBA:
Hi SQL Server Docs Team,
As a best practice, it is recommended to revert "show advanced options" back to 0 after changing the desired advanced option, so I've figured I can add the command to set it back to false after enabling xp_cmdshell.

Best regards,

Vlad Drumea


rwestMSFT:
@VladDBA Thanks!

rwestMSFT:
#sign-off

PR #9135, opened 2023-07-06
Update sql-server-2022-database-engine-events-and-errors-14000-14999.md

VladDBA:
Hi SQL Server Docs Team,

I've fixed the incomplete Description for message_id 14852 so that it matches the text in sys.messages in SQL Server 2022-2016.

Thank you,

Vlad


Jak-MS:
@MashaMSFT 
- Can you review this PR?


#label:"aq-pr-triaged"

rwestMSFT:
@VladDBA Thank you! This change will not be approved, because the articles are published directly from the source code.

**However**, you have revealed a bug in the generator, so when we do a regular publish from the source again, it will include this change.

PR #9160, opened 2023-07-13
Update stopping-system-versioning-on-a-system-versioned-temporal-tabl…

Libens-bufo:
Description:
Typo fix.

Proposed change: add missing space


Court72:
Thanks, @libens-bufo!

PR #9168, opened 2023-07-17
Correct Alpine's ODBC v17 downloads signature filename

senorflor:
The verification will fail otherwise because the filename is not the same as what is downloaded above.


David-Engel (approved):
Thanks for catching that!

David-Engel:
#sign-off

PR #9191, opened 2023-07-28
Update create-fulltext-index-transact-sql.md

way0utwest:
This statement cannot run inside a user transaction, so adding a note to the remarks to clarify.


rwestMSFT:
Related to #9192.

PR #9192, opened 2023-07-28
Update alter-fulltext-index-transact-sql.md

way0utwest:
Added remarks section to match up with the CREATE FT Index and a note that this cannot run in a user transaction.


rwestMSFT:
Thanks, @way0utwest. We'll incorporate these two PRs into an internal PR that also refreshes these articles.

PR #9200, opened 2023-08-02
Update drop-index-transact-sql.md

ishepherd:
Fix broken reference into the syntax BNF.


WilliamDAssafMSFT:
Thanks @ishepherd for taking the time to submit this PR! Good catch. Your commit will be merged and this article update is in flight. After the updated article is published, GitHub will list you as an article contributor. 

#sign-off

PR #9223, opened 2023-08-08
Update Full Text Search - Contains Function

wolfffelipe:
Adjusting the Contains function with the quotes in the correct places


rwestMSFT:
Thanks for the feedback, @wolfffelipe!

rwestMSFT:
#sign-off

PR #9238, opened 2023-08-11

SQLRockstar:
Current docs state "Four types of masks are available", and then it lists five types (datetime added for sql 2022).


Court72:
Thanks, @sqlrockstar!

PR #9244, opened 2023-08-14
Clarify symmetric key remarks

way0utwest:
The existing remarks reference session, which has a different meaning for users and connections. This remark refers to other keys in this database, which cannot have the same identity_source.


VanMSFT:
Thanks for submitting this change, @way0utwest! This looks good to me. I'll approve it.

FYI - @rwestMSFT 

#sign-off

PR #9254, opened 2023-08-18
Specify optional parameters

way0utwest:
Noting which parameters are required and which are optional

### DIFF
--- a/docs/t-sql/statements/create-symmetric-key-transact-sql.md
+++ b/docs/t-sql/statements/create-symmetric-key-transact-sql.md
@@ -67,7 +67,7 @@ CREATE SYMMETRIC KEY key_name
 
 ## Arguments
  *Key_name*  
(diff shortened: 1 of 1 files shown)


VanMSFT:
@way0utwest - Hi Steve! Thanks for this addition. Although we appreciate your attempt at clarifying the article further, noting whether a parameter is optional is not something we normally do. This is because it's implied in the Syntax section, and the way we specify our syntax convention. Please let us know if you have further feedback, but for now, I will deny this addition. Thanks!

way0utwest:
I get that. The thing that was strange here is you have the various parameters specified the same in the syntax section, but some are not optional. Should the syntax diagram reflect that?

VanMSFT:
> I get that. The thing that was strange here is you have the various parameters specified the same in the syntax section, but some are not optional. Should the syntax diagram reflect that?

Absolutely. If you see that the Syntax is formulated incorrectly according to the convention, that would be where to fix it.

Court72:
#label:"aq-pr-triaged"

rwestMSFT:
@way0utwest This change has been incorporated into an internal PR. When that PR is merged, this one will automatically close, and the article will be published shortly thereafter.

Thank you again for your contribution.

PR #9257, opened 2023-08-19
Minor syntax correction: missing a space

finnigd:
A space is missing: "...URIto..." should be "...URI to..."


Court72:
Thanks @finnigd!

PR #9265, opened 2023-08-23
Update info about TF 8558

lucastanheta:
This edit is to include the indication that TF 8558 is also supported on SQL Server 2019 CU18 and later + SQL Server 2022


rwestMSFT:
Thank you, @lucastanheta. We will include this change in an internal article update, and this PR will automatically close.

PR #9287, opened 2023-08-29
Add docs about sys.dm_xe_database_session_targets:failed_buffer_count

slaymaker1907:
This is a column we recently added to help track failures of individual targets. It is not yet present in SQL MI or box which is why the docs for sys.dm_xe_session_targets have not also been updated.


rwestMSFT:
@slaymaker1907 I'll pull these into an internal PR and refresh the underlying articles at the same time. Thanks for the contribution.

rwestMSFT:
Replaced by https://github.com/MicrosoftDocs/sql-docs-pr/pull/28194.

PR #9315, opened 2023-09-11
Update date-transact-sql.md

salmanarshad2000:
Unseparated string literal - 6 digit literals correspond to yymmdd, not yyyymm

### DIFF
--- a/docs/t-sql/data-types/date-transact-sql.md
+++ b/docs/t-sql/data-types/date-transact-sql.md
@@ -64,7 +64,7 @@ The following tables show the valid string literal formats for the **date** data
   
 |Unseparated|Description|  
 |-----------------|-----------------|  
-|[yy]yymmdd<br /><br /> yyyy[mm][dd]|The **date** data can be specified with four, six, or eight digits. A six- or eight-digit string is always interpreted as **ymd**. The month and day must always be two digits. A four-digit string is interpreted as year.|  
+|[yy]yymmdd<br /><br /> yyyy|The **date** data can be specified with four, six, or eight digits. A six- or eight-digit string is always interpreted as **ymd**. The month and day must always be two digits. A four-digit string is interpreted as year.|  
   
 |ODBC|Description|  
 |----------|-----------------|  


Court72:
@MikeRayMSFT

#label:"aq-pr-triaged"

salmanarshad2000:
Testing if `nnnnnn` is parsed as `yyyymm` or `yymmdd`:

```sql
select datestr, cast(datestr as datetime2(0)) as dateval
from (values
    ('2012')     -- Jan 01 2012
  , ('121002')   -- Oct 02 2012 or Oct 02 1912
  , ('20121002') -- Oct 02 2012
  , ('9999')     -- Jan 01 9999
--, ('999912')   -- exception, parsed as yymmdd instead of yyyymm
) as x(datestr)
```

#sign-off

MikeRayMSFT:
Reviewing

MikeRayMSFT (approved):

MikeRayMSFT:
With thanks to @salmanarshad2000 

#sign-off

rwestMSFT:
#hold-off

rwestMSFT:
@salmanarshad2000 Your edit removes the optional values for month and day, changing the intent and meaning. I don't think this is the correct change to make, but I will refresh the article and consider a better representation of the examples there.

#please-close

rwestMSFT:
@salmanarshad2000 Following up to say the updated article will publish in the next day or so. You'll note that the `yyyy[mm][dd]` value has been changed to `yyyy[mmdd]`, which covers the scenarios more accurately.

salmanarshad2000:
> @salmanarshad2000 Following up to say the updated article will publish in the next day or so. You'll note that the `yyyy[mm][dd]` value has been changed to `yyyy[mmdd]`, which covers the scenarios more accurately.

Choosing between `yyyy | [yy]yymmdd` and `[yy]yymmdd | yyyy[mmdd]` is opinion based, but it is correct either way.

rwestMSFT:
@salmanarshad2000 Sure, but that's not the debate here. Your original edit removed `[mm][dd]` completely, leaving behind just `yyyy`, which is not correct. That's why I replaced it with `yyyy[mmdd]`.

Edited to add: Thank you for your feedback. Without it, this article wouldn't have been reviewed, so we do appreciate it.

salmanarshad2000:
> @salmanarshad2000 Sure, but that's not the debate here. Your original edit removed `[mm][dd]` completely, leaving behind just `yyyy`, which is not correct. That's why I replaced it with `yyyy[mmdd]`.
> 
> Edited to add: Thank you for your feedback. Without it, this article wouldn't have been reviewed, so we do appreciate it.

You're welcome.

My point is that the grammar `[yy]yyMMdd | yyyy[MMdd]` has an overlap for 8 digit string. My proposed example `yyyy | [yy]yyMMdd` is better (opinion based of course). 4 digit strings match `yyyy` and 6/8 digit strings match `[yy]yyMMdd`.

rwestMSFT:
Yes, I understand, and the overlap is acceptable, because of order of precedence.

- The first option allows 6 or 8 digit values: `yyyyMMdd` or `yyMMdd`.
- The second option allows 4 or 8 digit values: `yyyy` or `yyyyMMdd`.

The second option reminds customers that `yyyy` is valid, but shouldn't preclude the option of `yyyyMMdd`. There is ambiguity in just having `yyyy`.

PR #9323, opened 2023-09-12
Update tables-transact-sql.md

DickBaker:
declaring nvarchar(128) and sysname separately suggests they are DIFFERENT but these datatypes are identical. IMHO you should banish such variance in ALL the TSQL docs and just state nvarchar(128). Less clutter/complication and less brain cycles required to understand it!


American-Dipper:

 
#label:"aq-pr-triaged" 

rwestMSFT:
Hi @DickBaker Thanks for your feedback. These data types are considered distinct internally, and we will not be removing **sysname** from the documentation.

PR #9370, opened 2023-10-02
Update configure-the-remote-access-server-configuration-option.md

devsikander:
Explain that example disables remote access

### DIFF
--- a/docs/database-engine/configure-windows/configure-the-remote-access-server-configuration-option.md
+++ b/docs/database-engine/configure-windows/configure-the-remote-access-server-configuration-option.md
@@ -67,10 +67,10 @@ To execute `sp_configure` with both parameters to change a configuration option,
 
 1. From the Standard bar, select **New Query**.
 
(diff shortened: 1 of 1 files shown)


rwestMSFT:
Hi @sikanderiqbal, thank you for your contribution.

I will not be accepting this PR for two reasons:

1. The original article already contains a description of what `0` (*disabled*) means:

   > The default value for the **remote access** option is `1` (enabled). This grants permission to run local stored procedures from remote servers or remote stored procedures from the local server. To prevent local stored procedures from being run from a remote server or remote stored procedures from being run on the local server, set the option to `0` (disabled).

   Adding the proposed change to the example text doesn't improve the meaning of `0` (*disabled*).

1. This feature is deprecated, and the top of the article makes this clear. Since the default is `1` (*enabled*), and customers are discouraged from changing the setting, your proposed change draws unnecessary attention to disabling it.

Thank you for your understanding.

PR #9443, opened 2023-10-23
Avoid suggesting to run curl as superuser

riuri:
In this case, curl doesn’t need to be run as superuser. Superuser is only required for writing to the specified files.

With the suggestions, we also skip the steps to enter and exit superuser mode.


rwestMSFT:
@riuri Thank you for your contribution. These changes look good, so I'm going to check back with the product team to confirm.

riuri:
@rwestMSFT nice, thanks!

@microsoft-github-policy-service agree company="ATS"

rwestMSFT:
We've merged your PR, and the new article should publish in the next day or so.

PR #9491, opened 2023-11-12
Corrected feature name to "parallel index operations"

finnigd:
I believe the feature should be called "parallel index operations" (not "parallel indexed operations") as the documentation refers to the former (https://learn.microsoft.com/en-us/sql/relational-databases/indexes/configure-parallel-index-operations?view=sql-server-ver16), and the latter sounds like operations involving indexes (ie. most queries), rather than operations _on_ indexes.


Court72:
@rwestMSFT

#label:"aq-pr-triaged"

rwestMSFT:
@finnigd Good edit. However, this change is not restricted to a single article, so I will incorporate it into a larger internal PR that fixes it in other places.

When that internal PR is merged, this one will automatically close. Thanks for your contribution.

PR #9496, opened 2023-11-15
added LOCAL/GLOBAL to SET @cursorVar = select ....

OleksiiKovalov:
SQL Server 19 allows both LOCAL and GLOBAL specifiers to set @ var = cursor

```
select @@version
declare @var cursor
set @var = CURSOR LOCAL FOR select * from sys.columns
set @var = CURSOR GLOBAL FOR select * from sys.columns

```

Microsoft SQL Server 2019 (RTM-CU23) (KB5030333) - 15.0.4335.1 (X64) 
	Sep 21 2023 17:28:44 
	Copyright (C) 2019 Microsoft Corporation
	Developer Edition (64-bit) on Windows 10 Pro 10.0 <X64> (Build 22631: ) (Hypervisor)

image

### DIFF
--- a/docs/t-sql/language-elements/set-local-variable-transact-sql.md
+++ b/docs/t-sql/language-elements/set-local-variable-transact-sql.md
@@ -41,7 +41,8 @@ SET
 |   
   { @cursor_variable =   
     { @cursor_variable | cursor_name   
-    | { CURSOR [ FORWARD_ONLY | SCROLL ]   
+    | { CURSOR [ LOCAL | GLOBAL ]
+        [ FORWARD_ONLY | SCROLL ]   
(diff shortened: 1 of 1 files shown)


rwestMSFT:
@OleksiiKovalov Once again we appreciate your close attention to detail.

I need to run this past the product group, and as you may be aware, it's a busy time for us, so it may take a while to confirm and merge your change.

rwestMSFT:
@OleksiiKovalov We have staged an internal PR that includes this change, as well as a refresh of the underlying article. When that is merged, this PR will automatically close.

Please note that we have modified the syntax slightly as follows:

`CURSOR [ [ LOCAL | GLOBAL ] ]`

The way you had it, it implied that either one was required, but with the additional square brackets, it indicates that both can be left out (with `LOCAL` being the default).

PR #9501, opened 2023-11-18
Update select-over-clause-transact-sql.md

HugoKornelis:
CHECKSUM is not an aggregate function. I assume (and a quick test confirms) that the limitation is on CHECKSUM_AGG.


HugoKornelis:
I noticed that there is actually a bigger error in the documentation. I will close this PR and make a new one.

PR #9502, opened 2023-11-18
Update select-over-clause-transact-sql.md

HugoKornelis:
The listed restriction is incorrect. CHECKSUM is not an aggregate function at all. If CHECKSUM_AGG was meant, then it was still incorrect, because CHECKSUM_AGG can be used with OVER without errors. Example (AdventureWorks): SELECT p.ProductID,
       p.Name,
       p.ListPrice,
       p.ProductLine,
       CHECKSUM_AGG (p.DaysToManufacture) OVER (PARTITION BY p.ProductLine
                                                ORDER BY p.ListPrice
                                                ROWS BETWEEN 4 PRECEDING AND 2 FOLLOWING) AS MovingAverage
FROM   Production.Product AS p;

This only fails when you change it to CHECKSUM_AGG(DISTINCT p.DaysToManufacture). But the error you get in that case is the same that you also get for COUNT(DISTINCT p.DaysToManufacture) or even MAX(DISTINCT p.DaysToManufacture). So it is a restriction on DISTINCT, not on CHECKSUM_AGG (or on CHECKSUM)


Court72:
@VanMSFT

#label:"aq-pr-triaged"

VanMSFT:
Thanks, @HugoKornelis! Looks good to me.

#sign-off

PR #9506, opened 2023-11-21
Explicit limitation of master log backups

way0utwest:
The product does not allow the master log to be backed up, though you can change to full mode. This change explicitly states no log backup allowed


rwestMSFT:
Steve, once again we appreciate your contribution, and once again it looks like you've uncovered an article that needs a refresh as well. We'll pull this into an internal PR and you'll get the appropriate credit.

way0utwest:
FWIW, I think these could also use the explicit limitation: 

- https://learn.microsoft.com/en-us/sql/t-sql/statements/backup-transact-sql?view=sql-server-ver16
- https://learn.microsoft.com/en-us/sql/relational-databases/backup-restore/back-up-a-transaction-log-sql-server?view=sql-server-ver16

PR #9513, opened 2023-11-23
Removed spurious optional semicolon

mdruiter:
I don't think this optional semicolon is correct.


American-Dipper:

 
#label:"aq-pr-triaged" 

rwestMSFT:
@mdruiter Thanks for your feedback.

We had some new code automation add the semi-colon for system object syntax, and it looks like it was included here as well. Agreed that functions shouldn't take them, but this might be a bigger issue than a single article. I'll incorporate this PR into an internal fix that will address all of them.

PR #9524, opened 2023-11-29
Added that the IP resource needs to be offline and online to take effect the Probe port settings.

yukitaza:
The settings made with set-clusterparameter will take effect the next time the resource comes online. This was not explicitly stated in the public information, so I will add it.

image
Set-ClusterParameter output the following warning after setting.

WARNING: The properties were stored, but not all changes will take effect until IP Address [resource_name] is taken offline and then online again.


ShannonLeavitt:
@tarynpratt

#label:"aq-pr-triaged"

tarynpratt:
@yukitaza Thank you for making this change. It should be visible to all users later today. 

#sign-off

PR #9547, opened 2023-12-12
Update alter-workload-group-transact-sql.md

kendra-little:
Adding note that to modify resource governor configuration in Azure SQL Managed instance, you must be in the context of the master database. 

If I attempt to run a resource governor command in the context of another database, I receive the message:

> Msg 40510, Level 16, State 2, Line 31
> Statement 'ALTER RESOURCE GOVERNOR' is not supported in this version of SQL Server.

But it works when I use master.

I think probably this note should be integrated into an include file in some way, but I couldn't manage to do this through the web UI and I wanted to submit this before I forgot. If it's easier to close this PR and make the change another way, totally understand. Thanks in advance!

### DIFF
--- a/docs/t-sql/includes/alter-workload-group.md
+++ b/docs/t-sql/includes/alter-workload-group.md
@@ -7,6 +7,9 @@ ms.topic: include
 ---
 Changes an existing Resource Governor workload group configuration, and optionally assigns it to a Resource Governor resource pool.
 
+> [!NOTE]  
+> For Azure SQL Managed Instance, you must be in the context of the `master` database to alter Resource Governor configuration.
+
 :::image type="icon" source="../../includes/media/topic-link-icon.svg" border="false"::: [Transact-SQL syntax conventions](../language-elements/transact-sql-syntax-conventions-transact-sql.md).
 
 ## Syntax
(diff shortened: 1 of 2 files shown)


American-Dipper:

 
#label:"aq-pr-triaged" 

rwestMSFT:
@kendra-little Are you suggesting that this should be noted across all the Resource Governor articles in this section where we mention SQL Managed Instance?

kendra-little:
> @kendra-little Are you suggesting that this should be noted across all the Resource Governor articles in this section where we mention SQL Managed Instance?

Hiya,

I don't have a super strong opinion -- I ran into the issue when altering the default workload group, which I think might be one of the more common operations for resource governor with Managed Instance. 

But I can see how someone might run into it on other pages like https://learn.microsoft.com/en-us/sql/relational-databases/resource-governor/create-and-test-a-classifier-user-defined-function?view=sql-server-ver16, which might also be common for folks to use due to the handy scripts.

Whatever you think best, really :)   Just took me a while to figure it out, especially because I don't see the same behavior on my local SQL Server install, which doesn't seem to care which database I'm using when running these commands. I was convinced for a brief time that it was broken or not allowed until I thought to try using master.

Thank you!

kendra-little:
ps -- i think i just realized why  you asked. I made this PR using the "pencil" and I saw 

> [!INCLUDE [alter-workload-group](../includes/alter-workload-group.md)]

which seemed like where the modification would go, but I didn't know how to edit that directly through the UI :) 

Sorry if that was confusing!

rwestMSFT:
@kendra-little I think this is what you had in mind. Check the files tab now.

PR #9551, opened 2023-12-14
Updated existing VMs grammar

bubbletroubles:
The header currently says "Exist VMs". This is not grammatically correct, it should be "Existing VMs". Additionally, there is HTML anchor tag for this heading as "existing-vms". The tag can now be removed, as the anchor "existing-vms" will be catered for with the new heading.


bluefooted (approved):

rwestMSFT:
Thanks! This probably happened during a regular gerund cleanup process. We appreciate your feedback.

PR #9555, opened 2023-12-16

sommarskog:
I've a number of changes in the table which details what is the default collation given the Windows locale. Some of these changes are based on actually testing running the SQL Server Setup with that locale. But there are also a couple which had the text "not available at server level" where I have made changes based on my results for other locals that had this. Thus, these changes needs to be verified. I have marked these changes with (*).

Permit me also point out that there are quite a few Windows locale entirely missing from the list.

### DIFF
--- a/docs/relational-databases/collations/collation-and-unicode-support.md
+++ b/docs/relational-databases/collations/collation-and-unicode-support.md
@@ -159,7 +159,7 @@ The following table shows the default collation designations, as determined by t
 |Afrikaans (South Africa)|0x0436|0x0409|Latin1_General_CI_AS|
 |Albanian (Albania)|0x041c|0x041c|Albanian_CI_AS|
 |Alsatian (France)|0x0484|0x0409|Latin1_General_CI_AS|
-|Amharic (Ethiopia)|0x045e|0x0409|Latin1_General_CI_AS|
+|Amharic (Ethiopia)|0x045e|0x0409|Latin1_General_CI_AI|
 |Arabic (Algeria)|0x1401|0x0401|Arabic_CI_AS|
 |Arabic (Bahrain)|0x3c01|0x0401|Arabic_CI_AS|
 |Arabic (Egypt)|0x0c01|0x0401|Arabic_CI_AS|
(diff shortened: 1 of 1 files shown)


sommarskog:

Court72:
@WilliamDAssafMSFT

#label:"aq-pr-triaged"

WilliamDAssafMSFT:
Thanks @sommarskog. 

It could take from a few business days to a few weeks to look further into this with the product team and decide on how to address, especially with the holidays. We'll contact you if we need further information, and you'll receive an email message each time we mention you in a comment. 
	
Thanks again for this feedback, as always your suggestions help improve our documentation.

WilliamDAssafMSFT:
@sommarskog It does look like some are missing compared to https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-LCID/%5bMS-LCID%5d.pdf. We will be reviewing, likely responding in 2024.

WilliamDAssafMSFT:
@sommarskog we'll continue to investigate. 
["For backward compatibility reasons, the default collation is set to the oldest available version that's associated with each specific locale. Therefore, this isn't always the recommended collation."](https://learn.microsoft.com/en-us/sql/relational-databases/collations/collation-and-unicode-support?view=sql-server-ver16#:~:text=For%20backward%20compatibility%20reasons%2C%20the%20default%20collation%20is%20set%20to%20the%20oldest%20available%20version%20that%27s%20associated%20with%20each%20specific%20locale.%20Therefore%2C%20this%20isn%27t%20always%20the%20recommended%20collation.)
Are you suggesting that these default collations are used by SQL upon install, or, are you observing the default collations for use in Windows?

sommarskog:
> @sommarskog It does look like some are missing compared to https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-LCID/%5bMS-LCID%5d.pdf. We will be reviewing, likely responding in 2024.

I should add that I am not sure how meaningful it is to add all these locales. The table is already frighteningly long already. And I would not expect there be many servers with these Windows locales as the system locale. Maybe there just should be a note that says that missing locales have one of Latin1_General_CI_AI/AS, which I would expect to be the case for the majority of them. If there are any deviations, they would have to be listed. Then again, it's not really what they call inclusive. Maybe the table should be moved to a separate topic?

sommarskog:
> "For backward compatibility reasons, the default collation is set to the oldest available version that's associated with each specific locale. Therefore, this isn't always the recommended collation.

I had missed that passage. Thanks for making me aware of it. I'm working on an article on collations, and I've just finished the section on default locales (which as a side effect resulted in this PR. There's all reason for to quote this part.

Now to your question:

> Are you suggesting that these default collations are used by SQL upon install, or, are you observing the default collations for use in Windows?

Not sure what you mean here. The way I tested this is that I changed my system locale (on a test VM), rebooted and ran SQL Server setup. I clicked through the wizard until I came to the Server configuration page and observed what I saw in the Collation tab.

setupcollation

WilliamDAssafMSFT:
OK, so you tested the values in each Windows locale by using SQL Setup and seeing what the default collation is listed there. That's useful info. Thanks. Will update.

sommarskog:
And just to repeat, I did not test those I marked with (*), but they are based on my assumptions. I left some work for Microsoft to verify. :--)

WilliamDAssafMSFT:
Hi @sommarskog, just following up on this. It's going to be a bit and requires coordination internally. I'm going to close this, it's not clear yet what should change, and will require some research to come back in. Really appreciate the effort and information here, in time it will pay off.

sommarskog:
OK. It remains to see what state the topic is in when I publish my article. Don't worry. That may not happen very soon.

PR #9560, opened 2023-12-19
Missed negation in the doc?

AntonC9018:
The doc [here](https://learn.microsoft.com/en-us/sql/relational-databases/partitions/create-partitioned-tables-and-indexes?view=sql-server-ver16#create-a-partitioned-table-on-one-filegroup-using-transact-sql) says it's not supported, so I assume the author missed the "not" or else it doesn't make sense. Unless I'm missing something.


VanMSFT:
Thanks for catching this, @AntonC9018! I'll approve.

#sign-off

VanMSFT:
#sign-off

PR #9569, opened 2023-12-23
standardise on UPPERCASE for type0..type2 as per docs

DickBaker:
The specified TYPE should standardise to be uppercase as explained in the earlier text. It is incorrect to assume that all MSSQL instances or individual databases have Case-insensitive collation!

P.S. you will have to merge my earlier PR9568

### DIFF
--- a/docs/relational-databases/system-stored-procedures/sp-addextendedproperty-transact-sql.md
+++ b/docs/relational-databases/system-stored-procedures/sp-addextendedproperty-transact-sql.md
@@ -179,9 +179,9 @@ GO
 
 EXEC sp_addextendedproperty @name = N'Caption',
     @value = 'Postal code is a required column.',
-    @level0type = N'Schema', @level0name = 'Person',
-    @level1type = N'Table', @level1name = 'Address',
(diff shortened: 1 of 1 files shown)


Jak-MS:
@markingmyname 
- Can you review this PR?


#label:"aq-pr-triaged"

rwestMSFT:
Thanks for your contribution.

I'm confused, though. On the one hand, you remove the `N''` prefix from a number of places, and on the other you leave the prefix there and only change the casing. You also spread your changes across two PRs on the same lines in the same file, making this incredibly difficult to parse.

I'll have to speak to a member of the product group, which means these two PRs will have to be merged manually, and only reviewed next year when everyone is back.

rwestMSFT:
After careful review, I figured out what your intention was. There are three articles affected by your suggestions, so we'll publish an internal PR that includes your suggested changes.

Thanks again for your contribution.

DickBaker:
Randolph: OK glad you got your head round it! Will progress on your internally worked up correction (3 articles) notify me via this PR9569 so I can see satisfactory conclusion ?Tks

PR #9583, opened 2024-01-03
Update rank-transact-sql.md

guzmanda:
The result set in Example B is incorrect. It seems the actual query used DENSE_RANK instead of RANK for the result. Corrected the result for fix the RaankBySalary column values.


rwestMSFT:
Thanks for the edit, @guzmanda!

PR #9585, opened 2024-01-03
Update specifying-a-target-namespace-using-the-targetnamespace-attrib…

jtshaver:
…ute-sqlxml-4-0.md

Changed targetNameSpace.xml to targetNamespace.xml to match the code example


rwestMSFT:
@jtshaver Thank you. We'll include your contribution in an internal PR, and when that is merged, this PR will automatically close and you can expect to see the updated article in the next day or so.

PR #9593, opened 2024-01-08
Change order of server roles so they are alphabetical

jpomfret:
The roles in this table, the last two weren't alphabetical - also means they don't line up with the order they are displayed when you connect to an Azure SQL Server


Court72:
@AndreasWolter

#label:"aq-pr-triaged"

rwestMSFT:
@jpomfret Thank you for your contribution! We'll incorporate it into an internal PR that also refreshes the rest of the article. That will publish in the next day or so.

PR #9612, opened 2024-01-18
Update set-ansi-defaults-transact-sql.md

erikdarlingdata:
Add filtered indexes to the list of things with required ANSI SET options. Having any of these set will cause the following index create statement to fail.

--SET QUOTED_IDENTIFIER OFF;
--SET NUMERIC_ROUNDABORT ON;
--SET ANSI_WARNINGS OFF;
--SET ANSI_PADDING OFF;
--SET ANSI_NULLS OFF;
--SET CONCAT_NULL_YIELDS_NULL OFF;
DROP TABLE IF EXISTS #t;
CREATE TABLE #t(id bigint)
CREATE INDEX t ON #t(id) WHERE id = 0;
GO


rwestMSFT:
Thanks, @erikdarlingdata!

DickBaker:
FWIW this PR (like many others) is tagged with "do-not-merge" flag [by the prmerger-automator bot], but has been merged anyway, which smells like a contradiction. Maybe the actual approver of any PR should clear that flag (or a bot should automagically clear it as assumed the approver knows what s/he is doing? Hmm.

rwestMSFT:
@DickBaker Thank you for your feedback.

PR #9631, opened 2024-01-25
Update sys-fn-my-permissions-transact-sql.md

way0utwest:
added clarification. NULL, NULL returns server permissions


rwestMSFT:
Thanks Steve. We'll take this change to an internal PR and refresh the underlying article. You'll still get credit.

PR #9644, opened 2024-01-30
[Ledger] Add nsg and routing rules for ledger to work in SQL MI vnet

taicchoumsft:
While working with SQL MI team, we found that the default NSG and route rules do not allow SQL to communicate with the ledger.  

We document the manual rule changes that a user needs to make to the NSG and route rules.


rwestMSFT:
@taicchoumsft Please use the sql-docs-pr repository to add new content. Thank you.

If you'd like help getting started, please reach out to me internally.

PR #9646, opened 2024-01-31
Update diagnose-resolve-latch-contention.md

chel-from-chel:
I think there was a mistake. Instead of 'spinlocks' should be used 'latches'. This article is about latches.


bluefooted (approved):

chel-from-chel:
Now when I am reading through the article it starting to make sense. Closing the pull request.

PR #9652, opened 2024-02-02
Update sp-prepare-transact-sql.md - Make the sp_prepare handle example better

nethish:
The first example provided in [sp_prepare](https://learn.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-prepare-transact-sql?view=sql-server-ver16#examples) page is little bit confusing. The handle variable name and the param name are same, and it took some time (although only a little), to realize the difference.


Court72:
@markingmyname

#label:"aq-pr-triaged"

nethish:
Made changes to the other example as well to be in line with the first one. 

Problem:-
I was confused to see two @P1 variables declared -- one for handle and one as a parameter. This MR is to rename the handle variable for little bit of easy readability.

rwestMSFT:
Thank you, @nethish.

PR #9655, opened 2024-02-06
Update using-azure-active-directory.md

MaxMeng1985:
When using MSI for authentication in ODBC driver:  For user assigned MSI, the UID should be set to the object id of the user assigned MSI.  For system assigned MSI, we don't have to set the UID, but if we set any value for UID, it also need to be the object id of the system assigned MSI.  The current document didn't make that clear enough, also the screenshot is a bit mis-leading by include a username in the UID field.

### DIFF
--- a/docs/connect/odbc/using-azure-active-directory.md
+++ b/docs/connect/odbc/using-azure-active-directory.md
@@ -29,7 +29,7 @@ The `Authentication` keyword can be used when connecting with a DSN or connectio
 
 |Name|Values|Default|Description|
 |-|-|-|-|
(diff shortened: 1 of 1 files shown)


Court72:
@David-Engel

#label:"aq-pr-triaged"

docs/connect/odbc/using-azure-active-directory.md
```diff
@@ -79,7 +79,7 @@ It's possible to use Microsoft Entra authentication options when creating or edi
 
 `Authentication=ActiveDirectoryMsi` for Microsoft Entra managed identity authentication
 
-![The DSN creation and editing screen with Managed Service Identity authentication selected.](windows/create-dsn-ad-msi.png)
+![The DSN creation and editing screen with Managed Service Identity authentication selected.](windows/create-dsn-ad-sys-msi.png)
```
David-Engel, line 82: I would revert this change and simply update the existing screenshot with one like this:

image

rwestMSFT:
I'll do that, @David-Engel. Thanks for your contribution, @MaxMeng1985.

PR #9660, opened 2024-02-09
Update alter-table-transact-sql.md

mtclifford:
Updated Drop Column examples to include use of IF EXISTS with multiple columns being dropped.

### DIFF
--- a/docs/t-sql/statements/alter-table-transact-sql.md
+++ b/docs/t-sql/statements/alter-table-transact-sql.md
@@ -1503,20 +1503,29 @@ The examples in this section demonstrate dropping columns and constraints.
 
 #### A. Drop a column or columns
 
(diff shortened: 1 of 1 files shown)


Court72:
@markingmyname

#label:"aq-pr-triaged"

rwestMSFT:
@mtclifford Thank you for your suggestion. Unfortunately we cannot accept the changes in their current state.

- `IF EXISTS` syntax is not backward compatible with SQL Server 2014, which is still under extended support.

- Instructions in comments aren't translated by our internationalization process.

- The example requires conditional execution based on the state. If there is indeed an error caused by the existing script, then please provide feedback around that, which is backward compatible for SQL Server 2014 as well.

If you can produce an updated example that takes these issues into account, then we will be more likely to accept your contribution.

Thanks again for taking the time to create this PR.

PR #9661, opened 2024-02-09
Update add-sensitivity-classification-transact-sql.md

MaxMeng1985:
"SQL Server does not maintain a list of the mapping between LABEL and LABEL_ID.
Also, it seems most of the parameter of this T-SQL command are optional, e.g., when we create a classification using this T-SQL command, we can provide both LABEL and LABEL_ID. Or we can provide only one of them.   Also, we can set LABEL A with LABEL_ID 1 for object A, but also set LABEL A with LABEL_ID 2 for object B. 
if you want to ensure a one-to-one relationship between the LABEL and LABEL_ID, you need to specify both the LABEL and LABEL_ID correctly when you create a classification. "
My change is to trying to clarify the above.

### DIFF
--- a/docs/t-sql/statements/add-sensitivity-classification-transact-sql.md
+++ b/docs/t-sql/statements/add-sensitivity-classification-transact-sql.md
@@ -93,6 +93,7 @@ Is an identifier based on a predefined set of values which define sensitivity ra
 - Only one classification can be added to a single object. Adding a classification to an object that is already classified will overwrite the existing classification.
 - Multiple objects can be classified using a single `ADD SENSITIVITY CLASSIFICATION` statement.
 - The system view [sys.sensitivity_classifications](../../relational-databases/system-catalog-views/sys-sensitivity-classifications-transact-sql.md) can be used to retrieve the sensitivity classification information for a database.
+- SQL Server does not maintain a list of the mapping between LABEL and LABEL_ID, if you want to ensure a one-to-one relationship between the LABEL and LABEL_ID, you need to specify both the LABEL and LABEL_ID correctly when you create a classification. 
 
 ## Permissions
 


Court72:
@Madhumitatripathy

#label:"aq-pr-triaged"

Madhumitatripathy:
Thanks @MaxMeng1985. At present the ADD CLASSIFICATION command works if any one of the parameters are supplied. If user supplies only LABEL or only INFORMATION_TYPE without corresponding IDs, the command will succeed. However, it is recommended to have one to one mapping between LABEL_ID and LABEL name. Additionally RANK parameter is not used anywhere in SQL today, so just having a rank value against a column without a LABEL or INFORMATION_TYPE will not add any value and should be avoided. TSQL will succeed if only RANK parameter is provided which has been identified as a bug and this will be addressed in future.

MaxMeng1985:
> Thanks @MaxMeng1985. At present the ADD CLASSIFICATION command works if any one of the parameters are supplied. If user supplies only LABEL or only INFORMATION_TYPE without corresponding IDs, the command will succeed. However, it is recommended to have one to one mapping between LABEL_ID and LABEL name. Additionally RANK parameter is not used anywhere in SQL today, so just having a rank value against a column without a LABEL or INFORMATION_TYPE will not add any value and should be avoided. TSQL will succeed if only RANK parameter is provided which has been identified as a bug and this will be addressed in future.

@Madhumitatripathy, thanks for your comments and confirmation.  That is the reason I proposed the changes. Could you please help to review it? Thanks.

rwestMSFT:
@MaxMeng1985 We're going to do an internal PR which takes these suggestions into account, but we will not be approving this PR directly.

When the internal PR is merged, this PR will automatically close, and then the updated article will be published in the next business day.

MaxMeng1985:
> @MaxMeng1985 We're going to do an internal PR which takes these suggestions into account, but we will not be approving this PR directly.
> 
> When the internal PR is merged, this PR will automatically close, and then the updated article will be published in the next business day.

@rwestMSFT , thanks for the update, that works for me.

PR #9664, opened 2024-02-11
Update json-path-expressions-sql-server.md

jasonhorner:
Adding a note about escaping the dot operator (.) I ran into this scenario where there was a . in the column name but it wasn't obvious that it needed to be escaped as the column just returned null. if I used:` '$.content_access.event_date' `and it gave me a bit of trouble working out the correct syntax:  `'$."content_access.event_date"'`


Court72:
@jovanpop-msft

#label:"aq-pr-triaged"

jovanpop-msft:
#sign-off

PR #9686, opened 2024-02-22
Remove CAST in example in sys-fn-xe-file-target-read-file-transact-sql.md

LarnuUK:
Remove `CAST` on `timestamp_utc` on the last example. The column is already a `datetime2(7)`, so the `CAST` is not required, and could be seen to imply that the column is a different data type.


rwestMSFT:
Thank you! We'll incorporate this into an internal PR that refreshes the underlying article, and you will retain credit.

Jak-MS:
housekeeping ... 

#label:"aq-pr-triaged"

PR #9697, opened 2024-03-02

VladDBA:
Hello SQL Server Docs Team,

I think this sentence might have gotten a bit mixed up:
"instant file initialization can benefit from transaction log autogrowth events up to 64 MB." 
Since it's not IFI that benefits from 64MB autogrowth, but the other way around.
So, I've reordered it to this:
"transaction log autogrowth events up to 64 MB can benefit from instant file initialization."

Best regards,

Vlad


WilliamDAssafMSFT:
Nice catch, @VladDBA!  The word `from` was throwing that sentence off. This is a nice re-phrasing.

Thanks for taking the time to submit this PR! Your commit will be merged and this article update is in flight. After the updated article is published, GitHub will list you as an article contributor. 

#sign-off

PR #9699, opened 2024-03-04
Update sql-machine-learning-services-windows-install-sql-2022.md

greglow-sdu:
Further updates to installing python. The existing link to the downloads now only goes to a page where it says the files aren't available. The link that I've provided is to the same site but gets you to a page with the release that's required. Also, earlier today, I suggested a change to the command for installing R. Similar changes are needed for Python when using named instances. The instructions as shown would not work. And it's not obvious how to put the name of the named instance into the commands. I find most people are trying ".\SQLEXPRESS" or "MACHINENAME\SQLEXPRESS", neither of which will work.

### DIFF
--- a/docs/machine-learning/install/sql-machine-learning-services-windows-install-sql-2022.md
+++ b/docs/machine-learning/install/sql-machine-learning-services-windows-install-sql-2022.md
@@ -159,7 +159,7 @@ Beginning with [!INCLUDE [sssql22-md](../../includes/sssql22-md.md)], runtimes f
     icacls "C:\Program Files\Python310\Lib\site-packages" /grant "NT Service\MSSQLLAUNCHPAD":(OI)(CI)RX /T
     icacls "C:\Program Files\Python310\Lib\site-packages" /grant *S-1-15-2-1:(OI)(CI)RX /T
     ```
-Note: if you have installed SQL Server as a named service, you will need to check the name of hte launchpad service account. You can do that by running services.msc and checking the account that the SQL Server Launchpad Service is running as. For example, if you have a default installation of SQL Server Express edition, you would need to use these commands instead:
+Note: if you have installed SQL Server as a named service, you will need to check the name of the Launchpad service account. You can do that by running `services.msc` and checking the account that the SQL Server Launchpad service is running as. For example, if you have a default installation of SQL Server Express edition, you need to use these commands instead:
 
     ```cmd
     icacls "C:\Program Files\Python310\Lib\site-packages" /grant "NT Service\MSSQLLAUNCHPAD$SQLEXPRESS":(OI)(CI)RX /T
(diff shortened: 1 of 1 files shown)


Jak-MS:
@WilliamDAssafMSFT 
- Can you review this PR?

NOTE: The 'Notes" should be reformatted to use the `> [!NOTE]` formatting

#label:"aq-pr-triaged"

WilliamDAssafMSFT:
Thanks @greglow-sdu, we're reviewing internally and will treat these two PRs as one. We may need to bring your commits into an internal PR. Will update.

docs/machine-learning/install/sql-machine-learning-services-windows-install-sql-2022.md
```diff
@@ -159,6 +159,12 @@ Beginning with [!INCLUDE [sssql22-md](../../includes/sssql22-md.md)], runtimes f
     icacls "C:\Program Files\Python310\Lib\site-packages" /grant "NT Service\MSSQLLAUNCHPAD":(OI)(CI)RX /T
     icacls "C:\Program Files\Python310\Lib\site-packages" /grant *S-1-15-2-1:(OI)(CI)RX /T
     ```
+Note: if you have installed SQL Server as a named service, you will need to check the name of hte launchpad service account. You can do that by running services.msc and checking the account that the SQL Server Launchpad Service is running as. For example, if you have a default installation of SQL Server Express edition, you would need to use these commands instead:
```
seantleonard:
```suggestion
Note: if you have installed SQL Server as a named service, you will need to check the name of the Launchpad service account. You can do that by running `services.msc` and checking the account that the SQL Server Launchpad service is running as. For example, if you have a default installation of SQL Server Express edition, you need to use these commands instead:
```

WilliamDAssafMSFT:
@greglow-sdu I'm confused, `https://www.python.org/downloads/` doesn't appear to be broken. Perhaps they had a temporary outage? We'd rather not link to a specific version, for maintenance.

WilliamDAssafMSFT:
Thanks @greglow-sdu for taking the time to submit this PR! Good catch, this is an important addition to document behavior that is unintuitive. There are a few other edits needed for your changes and this article as a whole, so I'll bring this commit into a PR from the internal repo. Your commit will be merged and this article update is in flight. After the updated article is published, GitHub will list you as an article contributor. 

greglow-sdu:
ISite is not broken but it linked to there to get the 3.1 binaries but you can't get them there. And you must have 3.1.

Get Outlook for iOS<https://aka.ms/o0ukef>

WilliamDAssafMSFT:
Yeah, it looks like the Python site doesn't publish a full windows installer for security releases. Full releases, like https://www.python.org/downloads/release/python-31011/, have everything you need.

PR #9700, opened 2024-03-04
Update dbcc-cleantable-transact-sql.md

brettshearer:
Added additional information about the '0' is not allowed claim. I think the document should actually describe what really happens with '0' as there are internet posts around this but the MS docs are incomplete,


rwestMSFT:
Thank you for your contribution. Our policy is not to document undefined or unsupported behavior.

rwestMSFT:
@brettshearer We're going to take the contribution anyway. I spoke to a colleague about the article's original wording, and we're going to specifically say that `0` is undocumented and unsupported. You'll still get contributor credit.

PR #9702, opened 2024-03-04
Update perform-index-operations-online.md

kendra-little:
Adding note about longstanding issue where online index rebuilds take out exclusive (x) locks after a LOB column is dropped from the table.

See this blog post for more details: https://littlekendra.com/2024/02/07/sql-server-online-index-rebuild-offline-blocking-outage/


rwestMSFT:
@kendra-little We can't stack info boxes, so I'll refresh this article and figure out a better way to inline this note. Thanks for your contribution.

PR #9706, opened 2024-03-06
Update sqlpackage-for-azure-synapse-analytics.md

MaxMeng1985:
update the sqlpackage command example, because the current example will fail with Windows authentication not supported error. Since this document is written for Synapse, I think we should avoid that.


dzsquared:
thanks @MaxMeng1985!
#sign-off

PR #9718, opened 2024-03-15
Document has been updated with grammar correction

AzureMentor:
This document has been updated with the following grammar correction:

"lets assume" should be "let's assume"


WilliamDAssafMSFT:
Thanks @AzureMentor for taking the time to submit this PR! Good catch. Your commit will be merged and this article update is in flight. After the updated article is published, GitHub will list you as an article contributor. 

#sign-off

PR #9749, opened 2024-04-05
Update truncate-table-transact-sql.md to include a truncate rollback example

davepcallan:
There is a lot of misconceptions in the community regarding the ability to rollback a truncate in SQL Server. Many devs incorrectly think it is not possible. I think a simple SQL example for them to see it for themselves would be helpful.


davepcallan:

rwestMSFT:
@davepcallan Thanks for your contribution. Our contributor and style guide has Examples at the end of the article, so I will review and test your code, then refresh the article in an internal PR, which also incorporates your suggestion. When that PR is merged, this one will automatically close, and you will still get contributor credit.

PR #9750, opened 2024-04-05
Update sp-helplogreader-agent-transact-sql.md

0x7FFFFFFFFFFFFFFF:
Add important information


rwestMSFT:
Thanks @0x7FFFFFFFFFFFFFFF, we'll incorporate your change into an internal PR that also refreshes this article.

PR #9757, opened 2024-04-12
Update sys-extended-procedures-transact-sql.md

emad-almousa:
Hi,

The correct table should be  sys.all_objects not sys.objects. the table sys.all_objects contains a column called "type" where "X" is associated with "EXTENDED_STORED_PROCEDURE".

Also, worth stating when you execute: 
SELECT * 
FROM master.sys.extended_procedures;
GO

NO rows are being returned back...I am using SQL Server 2022 release. I think this table is not being populated anymore.

Regards,
Emad


Court72:
@rwestMSFT

#label:"aq-pr-triaged"

rwestMSFT:
I can't comment on whether the `sys.extended_procedures` is being populated, but we will incorporate your change into an internal PR that also refreshes this article.

Thanks for your contribution.

PR #9758, opened 2024-04-16
Remove References to AT TIME ZONE in FOR SYSTEM_TIME AS OF

LarnuUK:
The documentation states that `AT TIME ZONE` can be used in conjunction with `FOR SYSTEM_TIME AS FOR`. The example gives, however, gives an error:
> Incorrect syntax near 'AT'.

This is true if you use the clause against a `VIEW` or a `TABLE`, the same error occurs. [db<>fiddle](https://dbfiddle.uk/6Kv6-9o6)

It appears, therefore, that the documentation is incorrect, and the section should be removed, which this PR does.

*If wanted, however, I am happy to add an example that uses a variable of `datetimeoffset` with `AT TIME ZONE` used against it, but I would want to check out how this works before doing so, to make sure the documentation is informative.*

### DIFF
--- a/docs/relational-databases/tables/querying-data-in-a-system-versioned-temporal-table.md
+++ b/docs/relational-databases/tables/querying-data-in-a-system-versioned-temporal-table.md
@@ -99,16 +99,6 @@ SELECT * FROM [vw_GetOrgChart]
 FOR SYSTEM_TIME AS OF'2021-09-01 T10:00:00.7230011';
 ```
 
-Alternatively, you can query the view using the `AS OF` subclause with a local time zone and `AT TIME ZONE`:
-
-```sql
-/* Querying view AS OF with local time*/
-DECLARE @LocalTime DATETIMEOFFSET = '2021-09-01 10:00:00.7230011 -07:00';
-
-SELECT * FROM [vw_GetOrgChart]
-FOR SYSTEM_TIME AS OF @LocalTime AT TIME ZONE 'UTC';
-```
-
 ## Query for changes to specific rows over time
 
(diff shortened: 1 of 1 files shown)


Court72:
@rwestMSFT

#label:"aq-pr-triaged"

rwestMSFT:
Hi @LarnuUK

I originally authorized the PR last year that added this change, and didn't do due diligence to test it properly beforehand.

On the `WideWorldImporters` database, you *can* do this:

```sql
DECLARE @CustomerTimeZone NVARCHAR(128) = 'Central European Standard Time';

SELECT StockItemID, ValidFrom,
    --Assign the known offset only
    ValidFrom AT TIME ZONE 'Pacific Standard Time' AS OrderDate_TimeZonePST,
    --Assign the known offset, then convert to another time zone
    ValidFrom AT TIME ZONE @CustomerTimeZone AS OrderDate_TimeZoneCustomer
FROM Warehouse.StockItems;
```

However, you are correct, and `AT TIME ZONE` is not compatible with `AS OF`.

Instead of approving your PR directly, which removes the example entirely, I'm going to roll back the change from last year, but you will still get contributor credit.

Thanks for your contribution.

LarnuUK:
Thanks @rwestMSFT . I'll close this PR then as well, as a roll back does solve the problem as well.

rwestMSFT:
@LarnuUK Thanks - I'll comment here once I've refreshed all the articles in this area, because some of them haven't been looked at in a while.

rwestMSFT:
@LarnuUK We're merging just two article updates this week, while the rest of the temporal content is still to be refreshed in a second phase. Thank you for your patience.

PR #9763, opened 2024-04-22
Update create-check-constraints.md

brettshearer:
Remove code example that highlights a misunderstanding of how constraints work.

### DIFF
--- a/docs/relational-databases/tables/create-check-constraints.md
+++ b/docs/relational-databases/tables/create-check-constraints.md
@@ -40,9 +40,10 @@ monikerRange: "=azuresqldb-current||>=sql-server-2016||>=sql-server-linux-2017||
 3.  In the **Check Constraint Expression** dialog box, type the SQL expressions for the check constraint. For example, to limit the entries in the `SellEndDate` column of the `Product` table to a value that is either greater than or equal to the date in the `SellStartDate` column or is a NULL value, type:  
   
     ```sql  
-    SellEndDate >= SellStartDate OR SellEndDate IS NULL  
+    SellEndDate >= SellStartDate  
     ```  
-  
+    Note that SellEndDate does not need to be checked for NULL as the check constraint must evaluate to anything but false, and a null comparison results in neither true nor false
+    
      Or, to require entries in the `zip` column to be five digits, type:  
   
     ```sql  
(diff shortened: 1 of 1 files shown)


Court72:
@WilliamDAssafMSFT

#label:"aq-pr-triaged"

docs/relational-databases/tables/create-check-constraints.md
```diff
@@ -40,9 +40,10 @@ monikerRange: "=azuresqldb-current||>=sql-server-2016||>=sql-server-linux-2017||
 3.  In the **Check Constraint Expression** dialog box, type the SQL expressions for the check constraint. For example, to limit the entries in the `SellEndDate` column of the `Product` table to a value that is either greater than or equal to the date in the `SellStartDate` column or is a NULL value, type:  
   
     ```sql  
-    SellEndDate >= SellStartDate OR SellEndDate IS NULL  
+    SellEndDate >= SellStartDate  
     ```  
-  
+    Note that SellEndDate does not need to be checked for NULL as the check constraint must evaluate to anything but false, and a null comparison results in neither true nor false
```
rwestMSFT, line 45: @brettshearer I don't agree with this assertion. `IS NULL` evaluates to a Boolean.

WilliamDAssafMSFT:
#assign:rwestMSFT

docs/relational-databases/tables/create-check-constraints.md
brettshearer: Hi Randal,

Try the following code. It highlights that the 'is null' check is redundant (as the comparison of nullable fields results in null which is neither true or false.

```sql
create table DateSample
(
	DateFrom datetime,
	DateTo datetime
)

alter table DateSample
	Add constraint DateFromExample check (DateTo > DateFrom)

insert into DateSample values (null, null)
insert into DateSample values (null, getutcdate())
insert into dateSample values (getutcdate(), null)

insert into DateSample values (getutcdate(), getutcdate()) -- will fail
insert into DateSample values (getutcdate(), dateadd(minute, -1, getutcdate())) -- will fail
```

rwestMSFT: @brettshearer My name is Randolph.

I tried this on SQL Server 2022 with the latest Cumulative Update.

In both cases, whether with your constraint, or one based on the article, the second to last command succeeds, and the last one fails.

```sql
alter table DateSample
   Add constraint DateFromExample check (DateTo >= DateFrom OR DateTo IS NULL)
alter table DateSample
   Add constraint DateFromExample2 check (DateTo >= DateFrom)

-- this succeeds (I used >= instead of >)
insert into DateSample values (getutcdate(), getutcdate())

-- this succeeds
insert into DateSample values (getutcdate(), NULL)

-- this fails
insert into DateSample values (getutcdate(), dateadd(minute, -1, getutcdate()))
```

As I noted previously, `IS NULL` evaluates to a Boolean, so that explains the identical behaviour between your constraint and mine.

(**Edited** to add new `NULL` example)

rwestMSFT: @brettshearer Does the article contain a silly example? Yes. The `OR` section is redundant in this particular scenario because the columns are both nullable.

However, it is not correct to say that `IS NULL` doesn't resolve to true or false, because it explicitly returns a Boolean.

In my estimation, the article should be modified to change the wording of the example and preceding paragraph to exclude the `OR` subclause, but your assertion about `IS NULL` isn't correct.

I will create an internal PR to review this article and correct the example.

brettshearer:
Certainly my wording can be improved. Thanks for the prompt attention.

PR #9794, opened 2024-05-21
Removing duplicate info box

siggemannen:
Seems some merge gone wrong so the info box is displayed twice. Not sure which to keep, but "cached" sounds better in my ears


Court72:
@markingmyname

Note that one version uses "caches" and the other uses "stores," if that makes a difference in which note you decide to keep.

#label:"aq-pr-triaged"

rwestMSFT:
@siggemannen Agreed. We're incorporating your change into an article refresh, which will publish in the next day or so. When that PR is merged, this one will close automatically.

PR #9801, opened 2024-05-29
Update tune-performance-with-the-query-store.md

lightbulbdev:
1. Number of executions for each query - Edited to consolidate duplicates where the query id, plan id, and SQL text are the same and sums the execution counts
2. Queries with multiple plans - A section has been added to identify queries with the highest number of plans


lightbulbdev:

PR #9808, opened 2024-05-31
Update create-external-table-transact-sql.md - No support for nested …

nojetlag:
…buckets on S3

The documentation gives the reader the impression that subfolders / nested buckets are generally supported by polybase, however it seems that is not true for buckets that reside on s3. Therefore the documentation should be corrected.


Court72:
@WilliamDAssafMSFT

#label:"aq-pr-triaged"

WilliamDAssafMSFT:
Thanks @nojetlag for contacting Microsoft with feedback about our product documentation.  

Keep in mind that **ListBucket** is required for external tables or OPENROWSET queries that point to an S3 folder location, instead of a single file. Without ListBucket permissions, you will receive the error `Msg 4860, Level 16, State 7, Line 15 Cannot bulk load. The file "s3://<ip address>:9000/bucket/*.*" does not exist or you don't have file access rights.`  Keep in mind too the total URL length limit of 259 characters for Basic Auth. Any chance those two factors explain it? Can you provide any error associated with this?

rwestMSFT:
@nojetlag We haven't had a response from William's question to you from three weeks ago. If we don't hear back within a few days, I will close this PR.

PR #9811, opened 2024-06-05
Correct FCI instance for Enterprise Edition

SeanGallardy:
Correcting the number of FCIs available in clustering for Enterprise Edition which is still 16 and has not changed.


SeanGallardy:
This will need to be changed in 2016/2017/2019/and 2022 documentation.

MikeRayMSFT:
Replaced by https://github.com/MicrosoftDocs/sql-docs-pr/pull/30794

PR #9817, opened 2024-06-12
Update sys-dm-database-backups-azure-sql-database.md

marigbede:
After some testing and investigations, the sys.dm_database_backups does exist in the Hyperscale Tier. 

hs_001

hs_002


Court72:
@SudhirRaparla

#label:"aq-pr-triaged"

SudhirRaparla:
@Court72  - Can you please reach out to Dinakar Nethi as he is the new PM for Backup Restore.

SudhirRaparla:
DMV is a SQL Server entity so it will exist for Hyperscale Databases as well. However you wont see correct content being populated for Hyperscale database as it is snapshot based. And in the snapshot you can see only Transaction log backups. We should keep the preview banner and say that it is not applicable for Hyperscale atleast.

marigbede:
Then I would suggest that it be modified to especially reflect your comment. It explicitly communicates the position that it would not work at all.

PR #9819, opened 2024-06-17

andreasjordan:
Add the "Perform volume maintenance tasks" right to the service account, as the setup.exe is doing it. Downside when using an AD account: Configuration manager will not show the setting, as it just looks for the StartName of the service. We wrote about the topic here: https://blog.ordix.de/instant-file-initialization-microsoft-sql-server-set-up-check


Court72:
@WilliamDAssafMSFT

#label:"aq-pr-triaged"

rwestMSFT:
@andreasjordan We appreciate your feedback. I have reached out to the security team. It may take weeks or months to get a response and update the content, and we'll let you know here when it happens.

rwestMSFT:
#sign-off

PR #9821, opened 2024-06-17
Incorrect memory values

ProfiseeAdmin:
HS premium memory optimized tier has double the amount of ram available per vCore. Only 24, 64 and 80 core columns had the right values.


ProfiseeAdmin:
Messed up

PR #9822, opened 2024-06-17
Fix HS Memory Optimized RAM values

ProfiseeAdmin:
HS Memory Optimized has twice the amount of RAM per vCore. Only the 24c, 64c and 80c columns had the right values.


dimitri-furman (approved):

arvindshmicrosoft (approved):
Thanks for identifying and correcting this, @ProfiseeAdmin !

dimitri-furman:
#sign-off

PR #9823, opened 2024-06-20
Update sp-executesql-transact-sql.md

bpd0018:
There are way more than 12 statements generated. You get one statement for each year and month combination for each language where the first 3 characters of the month name are different.

Personally, I think you would be better off using a table naming scheme that is not locale dependent like 'Sales_' +
cast(datepart(yy, @PrmOrderDate) as char(4)) +
'_' +
cast(datepart(mm, @PrmOrderDate) as char(2))

To see the table names that you get now, run the following ...

DECLARE @PrmOrderDate DATETIME;
SET LANGUAGE N'us_english';
SET @PrmOrderDate = '01/31/2021'; -- set date here because the date literal format also depends on LANGUAGE SELECT SUBSTRING(DATENAME(mm, @PrmOrderDate), 1, 3) + CAST(DATEPART(yy, @PrmOrderDate) AS CHAR(4)) + 'Sales'; SET LANGUAGE N'Español';
SELECT SUBSTRING(DATENAME(mm, @PrmOrderDate), 1, 3) + CAST(DATEPART(yy, @PrmOrderDate) AS CHAR(4)) + 'Sales'; SET LANGUAGE N'ελληνικά';
SELECT SUBSTRING(DATENAME(mm, @PrmOrderDate), 1, 3) + CAST(DATEPART(yy, @PrmOrderDate) AS CHAR(4)) + 'Sales';


Court72:
@markingmyname

#label:"aq-pr-triaged"

docs/relational-databases/system-stored-procedures/sp-executesql-transact-sql.md
```diff
@@ -221,7 +221,7 @@ EXEC sp_executesql @InsertString,
 GO
 ```
 
-Using `sp_executesql` in this procedure is more efficient than using `EXECUTE` to execute a string. When `sp_executesql` is used, there are only 12 versions of the `INSERT` string that are generated, one for each monthly table. With `EXECUTE`, each `INSERT` string is unique because the parameter values are different. Although both methods generate the same number of batches, the similarity of the `INSERT` strings generated by `sp_executesql` makes it more likely that the query optimizer reuses execution plans.
+Using `sp_executesql` in this procedure is more efficient than using `EXECUTE` to execute a string. With `EXECUTE`, each `INSERT` string is unique because the parameter values are different. When `sp_executesql` is used, there is only 1 `INSERT` string generated for each language, month and year combination. (DATENAME() returns different values depending on the value of LANGUAGE in the session that is executing the InsertSales procedure.) Although both methods generate the same number of batches, the similarity of the `INSERT` strings generated by `sp_executesql` makes it more likely that the query optimizer reuses execution plans.
```
rwestMSFT:
```suggestion
Using `sp_executesql` in this procedure is more efficient than using `EXECUTE` to execute a string. With `EXECUTE`, each `INSERT` string is unique because the parameter values are different. When `sp_executesql` is used, there is only one `INSERT` string generated for each language, month and year combination. (`DATENAME()` returns different values depending on the value of `LANGUAGE` in the session that is executing the `InsertSales` procedure.) Although both methods generate the same number of batches, the similarity of the `INSERT` strings generated by `sp_executesql` makes it more likely that the query optimizer reuses execution plans.
```

markingmyname (changes requested):
Please review @rwestMSFT comments before continuing to read my comments.


Using sp_executesql in this procedure is more efficient than using EXECUTE to execute a string. When using EXECUTE, each INSERT statement is unique because the parameter values are embedded directly in the string. In contrast, with sp_executesql, there is only one INSERT statement generated for each combination of language, month, and year. (Note that DATENAME() returns different values depending on the session's LANGUAGE setting when executing the InsertSales procedure.) Although both methods generate the same number of batches, the consistency of the INSERT statements generated by sp_executesql increases the likelihood that the query optimizer will reuse execution plans.

rwestMSFT:
After discussion with the product group, we are revising that paragraph further in an internal PR, which will be published in the next day or so. Thank you for your feedback.

PR #9832, opened 2024-06-28
Some default values for advanced option is different from actual SSMS…

taroarai:
… GUI

Some default values for advanced option is different from actual SSMS GUI.

ANSI PADDING
True > False
* this is showed as the screen capture on this page.

Generate Script for Dependent Objects
Script Indexes
True > True for entire database and False for specific object.

### DIFF
--- a/docs/ssms/scripting/generate-and-publish-scripts-wizard.md
+++ b/docs/ssms/scripting/generate-and-publish-scripts-wizard.md
@@ -146,7 +146,7 @@ Use this page to specify how you want this wizard to generate scripts. Many diff
 
 **General** - The following options apply to the entire script.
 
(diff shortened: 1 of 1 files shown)


Court72:
@markingmyname

#label:"aq-pr-triaged"

rwestMSFT:
@taroarai Thanks for your contribution. It may take some time to merge your suggestion, as it relates to SSMS, and we've publicly announced that a new version will be released in the near future. We'll incorporate your change into any content updates for that release.

/cc @markingmyname

koatsush:
@rwestMSFT , Thanks for your comment.
A very nervous client is pointing out the difference between the Learn and the SSMS defaults.
We would prefer that this be changed as soon as possible, but when exactly do you plan to change the document?

PR #9834, opened 2024-07-01
Table Variables can be used in INSERT..EXEC statements

LarnuUK:
The documentation advised that Table variables can't be used in an `INSERT...EXEC` statement. This *might* have been true before ([Stack Overflow](https://stackoverflow.com/q/14622940/2029983) suggests it was back in SQL Server 2000), however, it's not true in SQL Server 2014+. I have removed that statement, as it not true.

I've also "tidied" the same statement to make it clear that it's the `INTO` clause a table variable can't be used in; `SELECT ... INTO... FROM @TableVariable` is acceptable syntax.

### DIFF
--- a/docs/t-sql/data-types/table-transact-sql.md
+++ b/docs/t-sql/data-types/table-transact-sql.md
@@ -116,7 +116,7 @@ Assignment operation between **table** variables isn't supported.
 Because **table** variables have limited scope and aren't part of the persistent database, transaction rollbacks don't affect them.
 Table variables can't be altered after creation.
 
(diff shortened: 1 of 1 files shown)


Court72:
@MikeRayMSFT

#label:"aq-pr-triaged"

MikeRayMSFT:
@uc-msft for review

uc-msft:
Yes, this is correct. Table variables can be target of INSERT..EXEC statement in newer versions of SQL Server.

MikeRayMSFT (approved):

MikeRayMSFT:
#sign-off

With thanks to @LarnuUK

PR #9842, opened 2024-07-09
Fix spacing of "-V" command

emil-eklund:
Fixes spacing used in examples for the "-V" (severity) command.
Incorrect usage will cause the command to fail. 

See earlier discussion: https://github.com/Azure/sql-action/issues/235

### DIFF
--- a/docs/tools/sqlcmd/sqlcmd-utility.md
+++ b/docs/tools/sqlcmd/sqlcmd-utility.md
@@ -1352,11 +1352,11 @@ Use the following practices to help maximize security and efficiency.
 
 Use the following practices to help maximize correctness:
 
-- Use `-V16` to log any [severity 16 level messages](../../relational-databases/errors-events/database-engine-error-severities.md#levels-of-severity).  Severity 16 messages indicate general errors that can be corrected by the user.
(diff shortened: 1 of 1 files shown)


emil-eklund:
I think this passage is incorrect too, but I've not had the time to verify for all parameters:

>Currently, sqlcmd doesn't require a space between the command-line option and the value. However, in a future release, a space may be required between the command-line option and the value.

Court72:
@dlevy-msft

#label:"aq-pr-triaged"

dlevy-msft:
This appears to be an issue with the tool. Will need some time to investigate.

emil-eklund:
> This appears to be an issue with the tool. Will need some time to investigate.

I agree. But I do also think the documentation should be updated if the following statement is true: 

>Currently, sqlcmd doesn't require a space between the command-line option and the value. **However, in a future release, a space may be required between the command-line option and the value.**

rwestMSFT:
@emil-eklund Thanks for your patience. We are updating the **sqlcmd** article in the next business day or so, with the following note.

> [!NOTE]  
> If you use the `-i` option followed by one or more additional parameters, you must use a space between the parameter and the value. This is a known issue in **sqlcmd** (Go).

You will still receive contributor credit.

PR #9844, opened 2024-07-10
Update tcp-ip-properties-ip-addresses-tab.md

katrushenkov:
fix typo


Court72:
Thanks for the typo fix @katrushenkov

PR #9854, opened 2024-07-24
Note about Automatic Backups Taken by Azure Service Are Only Available for Restore Operations and Cannot Be Downloaded or Accessed Directly by End Customers

rarandiv:
There is no note in the document which says those automatic backups taken by the Azure service are not available for end customers to download or access directly; they can only be used for restore operations. Which I have append in the document.

### DIFF
--- a/azure-sql/database/automated-backups-overview.md
+++ b/azure-sql/database/automated-backups-overview.md
@@ -48,6 +48,8 @@ Azure SQL Database creates:
 
 The exact frequency of transaction log backups is based on the compute size and the amount of database activity. When you restore a database, the service determines which full, differential, and transaction log backups need to be restored.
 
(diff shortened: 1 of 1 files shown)


Court72:
@WilliamDAssafMSFT

#label:"aq-pr-triaged"

WilliamDAssafMSFT:
Hi @rarandiv! This is covered in other places, but could be called out here, in addition to listing the export database alternatives. Thanks for taking the time to submit this PR! There are a few other edits needed for your changes and this article as a whole, so I'll bring this commit into a PR from the internal repo. Your commit will be merged and this article update is in flight. After the updated article is published, GitHub will list you as an article contributor. 

rarandiv:
Hi @WilliamDAssafMSFT , Awesome, thank you for considering the feedback and taking swift action

PR #9857, opened 2024-08-07
Update from-using-pivot-and-unpivot.md to add UNPIVOT syntax

Rob-S:
I was looking for the UNPIVOT syntax, but the page only has PIVOT syntax. The most significant differences for UNPIVOT include no &lt;aggregate function&gt;, no &lt;alias for the pivot table&gt; and the pivot columns from the IN clause are not repeated in the SELECT clause. Also, there were "[" and "]" rather than "&lt;" and "&gt;" identifying placeholders to be replaced by column names.


Court72:
@VanMSFT

#label:"aq-pr-triaged"

docs/t-sql/queries/from-using-pivot-and-unpivot.md
```diff
@@ -38,22 +38,43 @@ The syntax for `PIVOT` is simpler and more readable than the syntax that may oth
...
-    [first pivoted column] AS <column name>,  
-    [second pivoted column] AS <column name>,  
-    ...  
-    [last pivoted column] AS <column name>  
+SELECT [<non-pivoted column> AS <column name>,]
+    ...
+    <first pivoted column> AS <column name>,
+    <second pivoted column> AS <column name>,
+    ...
+    <last pivoted column> AS <column name>
```
VanMSFT: The original syntax here is correct. The first column is not optional and hence encased in `< >` to denote the non-pivoted column. The rest of the columns are optional, and should be encased by brackets `[ ]`

docs/t-sql/queries/from-using-pivot-and-unpivot.md
```diff
@@ -38,22 +38,43 @@ The syntax for `PIVOT` is simpler and more readable than the syntax that may oth
...
+    (<SELECT query that produces the data>)   
+    AS <alias for the source query>  
+UNPIVOT
+(  
+    <new output column created for values in result of the source query>
+FOR   
+<output column for names of the pivot columns>
+    IN ( [first pivoted column], [second pivoted column],  
+    ... [last pivoted column])  
+)
```
VanMSFT: This syntax probably better suited in the [UNPIVOT Example](https://learn.microsoft.com/en-us/sql/t-sql/queries/from-using-pivot-and-unpivot?view=sql-server-ver16#unpivot-example) section.

VanMSFT (commented):
Thanks, @Rob-S! Please see my comments and let me know your thoughts.

docs/t-sql/queries/from-using-pivot-and-unpivot.md
Rob-S: > The original syntax here is correct. The first column is not optional and hence encased in < > to denote the non-pivoted column.

First, the first, non-pivoted column is optional. I can remove that from the first PIVOT example and when I run it, it works fine:

```-- Pivot table with one row and five columns  
SELECT [0], [1], [2], [3], [4]  
FROM  
(
  SELECT DaysToManufacture, StandardCost   
  FROM Production.Product
) AS SourceTable  
PIVOT  
(  
  AVG(StandardCost)  
  FOR DaysToManufacture IN ([0], [1], [2], [3], [4])  
) AS PivotTable;
```

So I believe using the "< >" to denote the column and "[ ]" to denote the whole thing, including "AS" and the alias, is correct. Technically, the AS clause is also optional, so it possibly should be:

`SELECT [<non-pivoted column> [AS <column name>], ]`

> The rest of the columns are optional, and should be encased by brackets [ ]

If that's the case, then the whole line, including "AS" and the alias should be encased in the square brackets "[ ]". The characters around the column name, ex: currently "[...]" as in "[second pivoted column]", denoting the column only, should be "&lt;...&gt;", as in "&lt;second pivoted column&gt;". Therefore, the syntax for that line would be:

`[<second pivoted column> [AS <column name>], ]`

The other lines under the SELECT clause would be similar.

docs/t-sql/queries/from-using-pivot-and-unpivot.md
Rob-S: The Remarks directly below the Syntax section directly reference the UNPIVOT clause. So if the Remarks are there, then I would say the UNPIVOT should remain above those Remarks.

The other option is:
- Change "Syntax" to "PIVOT Syntax"
- Add a "UNPIVOT Syntax" Section above the "UNPIVOT Example"
- Move the first paragraph of the Remarks, concerning the UNPIVOT syntax, below the "UNPIVOT Syntax" section.

docs/t-sql/queries/from-using-pivot-and-unpivot.md
VanMSFT: Hi @Rob-S - Thanks! You're right, you can run the script without a non-pivoted column. But I don't see that as very useful if that's the case. I agree with your changes though.

docs/t-sql/queries/from-using-pivot-and-unpivot.md
VanMSFT: I think it's probably best if we add some more context before the syntax to include `UNPIVOT`. Thanks for the suggestions!

docs/t-sql/queries/from-using-pivot-and-unpivot.md
```diff
@@ -38,23 +38,43 @@ The syntax for `PIVOT` is simpler and more readable than the syntax that may oth
 The following syntax summarizes how to use the `PIVOT` operator.
```
VanMSFT:
```suggestion
The following syntax summarizes how to use the `PIVOT` and `UNPIVOT` operator.  
```

rwestMSFT:
@Rob-S Thank you for your contributions. We're going to create an internal PR from this, to allow us freshness pass over the existing content, and include your changes as discussed with @VanMSFT.

On a personal note, I want to also thank you for another contribution you made some time ago regarding `QUOTED_IDENTIFIER`, which was merged last week.

Rob-S:
@rwestMSFT Just in summary for @VanMSFT, it originally looked like the optional clauses were not being highlighted with "[ ... ]" (as evidenced by the word "optional" in the "optional ORDER BY clause" part). Hence my original commit was simply to surround the column name placeholders properly with "< ... >". That's the simplest fix. As we later got into the discussion about optional clauses, that's what resulted in the more elaborate meta-syntax to surround the full-length (ex: including the AS <alias> part) optional clauses with "[ ... ]". Thanks!

PR #9858, opened 2024-08-09
Update manage-certificates.md to reflect a process that exists.

cslthawkinsv:
There is no version of SQL that has an Import button or wizard in the instance Properties dialogue.  There is also no such button or wizard in the Availability Group properties dialogue.  Please update the instructions for all versions of SQL to reflect a process that is possible.  Selecting the certificate from the dropdown if it happens to be visible.

### DIFF
--- a/docs/database-engine/configure-windows/manage-certificates.md
+++ b/docs/database-engine/configure-windows/manage-certificates.md
@@ -57,11 +57,9 @@ You can use certificate management in [!INCLUDE [ssnoversion-md](../../includes/
 
 1. Right-click **Protocols for** *&lt;instance Name&gt;*, and then select **Properties**.
 
-1. Choose the **Certificate** tab, and then select **Import**.
+1. Choose the **Certificate** tab, and then select the certificate from the dropdown list, then select **Apply**. > [!NOTE]  
+> There is no version of SQL that has an Import button or wizard in the instance Properties dialogue.  Please correct the page to reflect a feature that exists.
 
-1. Select **Browse** and then select the certificate file.
-
-1. Select **Next** to validate the certificate. If there are no errors, select **Next** to import the certificate to the local instance.
 ::: moniker-end
 
 ::: moniker range="<= sql-server-2017"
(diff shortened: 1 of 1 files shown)


cslthawkinsv:
@microsoft-github-policy-service agree company="Pentagon Federal Credit Union"

cslthawkinsv:
The image shows the properties dialogue without any Import button.

cslthawkinsv:
***@***.***

cslthawkinsv:
I have tried to paste/email images but they don't show up.  It's a very easy process to replicate.  Simply open the named properties dialogues and observe there is no Import option but there is a dropdown menu.

rwestMSFT:
@cslthawkinsv This is a screenshot from SQL Server Configuration Manager that I just took on 2022. It appears on 2019 as well.

I am closing your PR, as I cannot reproduce your issue.

Screenshot_2024-08-09_11h04m27

cslthawkinsv:
Interesting,
Here is a screenshot showing that the import button does not exist:

cslthawkinsv:
Here is another taken from SQL Server 2017 showing there is no import button even before TLS is configured the first time.

rwestMSFT:
@cslthawkinsv Correct, it's not available in SQL Server 2017.

Are you visiting the properties pane of the **Protocols**, as the article describes?

cslthawkinsv:
Randolph,
I see you are correct about the 2019 and 2022 versions.
I believe I attempted to edit the instruction page for SQL 2017.
That page now shows the edit I made (or at least the same text) in the single SQL Server Instance section but shows the Import instructions in both the failover cluster and availability group sections of the page.

Certificate management (SQL Server Configuration Manager) - SQL Server | Microsoft Learn<https://learn.microsoft.com/en-us/sql/database-engine/configure-windows/manage-certificates?view=sql-server-2017&preserve-view=true>
https://learn.microsoft.com/en-us/sql/database-engine/configure-windows/manage-certificates?view=sql-server-2017&preserve-view=true

This properties pane for protocols is indeed the one I’m showing.

rwestMSFT:
Unfortunately, no images are coming through. They aren't supported if you use email.

I understand what you're saying, and the issue is related to how the underlying Markdown works. I will work with the product group to update the article to remove references to the Import button in SQL Server 2017. This may take several weeks due to the size of our current backlog.

cslthawkinsv:
Thank you for hearing me and taking action.
The feedback process is a bit of a hassle but still better than almost every other company’s document feedback process, where they even exist.

PR #9864, opened 2024-08-23
Update register-a-service-principal-name-for-kerberos-connections.md

krishnamsft20:
need to update the steps. Original steps do not work


rwestMSFT:
@krishnamsft20 If you're an employee, please close this and use `sql-docs-pr` instead.

PR #9870, opened 2024-09-03
Fix column order of hobt_id

bangjiehan:
There is a different column order between docs and implementation.
Only checked in SQL Server 2022 and 2019.


Court72:
@rwestMSFT

#label:"aq-pr-triaged"

rwestMSFT:
Thank you for your contribution. However, this is not a significant enough change at this time. We may take this into account in a future update.

PR #9880, opened 2024-09-14
Update always-encrypted-database-engine.md

emad-almousa:
Hello,
I am recommending adding the following statement highlighted in bold:

This provides a separation between those who own the data and can view it, and those who manage the data but should have no access: on-premises database administrators, cloud database operators, or other high-privileged unauthorized users. As a result, Always Encrypted allows customers to securely store their sensitive data in the cloud, reducing the risk of data theft by malicious insiders. **In addition, its worth noting that Always Encrypted provides extended data protection by not exposing any histogram data which is a data exposure area for system admins, cloud operators, and highly privileged accounts in traditional database deployments.**

This something Microsoft is not highlighting as a security protection area being implemented when always encrypted feature is used. To clarify more, when a sensitive table is being audited against SELECT statements for example...an admin can partially by pass this by viewing histogram data instead of directly running SELECT statement. with always encrypted you are _**preventing**_ totally the admin from viewing sensitive data.

I hope my point is clear.

Thanks and have a good day
Emad


Pietervanhove (commented):
Hi Emad,

I don't this is a good place to call this out. We're already saying that high-privileged users cannot see the data, that includes histogram data.

Regards,
Pieter

emad-almousa:
Hi,

It's your call at the end, thanks for your response.

VanMSFT:
#please-close

PR #9906, opened 2024-10-15
Update sys-dm-broker-queue-monitors-transact-sql.md

MisterZeus:
Clean up overcomplicated T-SQL script, to make it more readable and understandable.


rwestMSFT:
The underlying article needs a refresh, so we’ll take your contribution into account when doing that. When the internal PR is merged, this PR will automatically close. Thank you for your contribution.

PR #9910, opened 2024-10-18
Update configure-the-user-connections-server-configuration-option.md

VladDBA:
Hello SQL Server Docs Team!

Long time no PR 😄 

### Inaccurate/confusing wording 
 1. "For example, if only 10 users are logged in, 10 user connection objects are allocated." 
 This doesn't seem accurate, or at least the wording doesn't help. 
 Since it makes it sound like the limit controls the number of users connected to the instance, not the total number of allowed connections, regardless of the users they originate from.
For example, if the limit would be set to 10, a single user with 10 connections can exhaust the number of allowed connections and subsequent connection attempts from other users would fail.
So, I've changed it to this:
 "For example, if only 10 users are logged in, each with one connection, 10 user connection objects are allocated. The same would happen in the case of a single user establishing 10 connections."

2. "to configure the value of the `user connections` option to `325` users."
Changed to: 
"to configure the value of the `user connections` option to `325` connections." 
Because the option controls the total number of connections that come from users.

### Adjustments to the example code
`USE AdventureWorks2022;` doesn't make much sense since it errors out if someone doesn't have AdventureWorks2022 on their instance (and not everyone does), so I replaced it with `USE master;` instead.

### Leaving `show advanced options` how we found it
Added command to set `show advanced options` back to 0 for security (by obscurity) purposes.

### Added step 4
Because the `user connections` configuration option isn't dynamic (as per the `is_dynamic` column in `sys.configurations`). This is confirmed also by setting the value to a low number (4) and then opening enough query editor tabs in SSMS, to go over the new limit, which doesn't lead to an error until after the instance is restarted.

P.S. This might not be my only PR tonight 😄 

Have a great weekend!

### DIFF
--- a/docs/database-engine/configure-windows/configure-the-user-connections-server-configuration-option.md
+++ b/docs/database-engine/configure-windows/configure-the-user-connections-server-configuration-option.md
@@ -19,7 +19,7 @@ helpviewer_keywords:
 
 [!INCLUDE [SQL Server](../../includes/applies-to-version/sqlserver.md)]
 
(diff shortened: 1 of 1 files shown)


rwestMSFT:
@VladDBA Since this is an old article, we need to refresh it to follow a new server configuration article style, so your PR will be incorporated into that. Thanks as always for your contribution.

PR #9911, opened 2024-10-18
Update configure-the-user-options-server-configuration-option.md

VladDBA:
Hello SQL Server Docs Team!

This was another reference to AdventureWorks2022 which doesn't really make much sense in the case of `sp_configure`.

Best regards,

Vlad


rwestMSFT:
Way ahead of you.

#please-close

PR #9925, opened 2024-11-06
Removing incorrect owner info

AndreasWolter:
All SQL Agent procs except for the new "managed backup" and "smart admin" stuff are owned by the dbo. Not db_owner.  Also, I don't see why this needs to be documented at all, since this has no effect on the behavior/permissions. Hence, I suggest removing to simplify this

### DIFF
--- a/docs/includes/msdb-execute-permissions.md
+++ b/docs/includes/msdb-execute-permissions.md
@@ -6,4 +6,4 @@ ms.service: sql
 ms.topic: include
 ms.custom: include file
 ---
-You can grant the `EXECUTE` permissions on this procedure, but these permissions may be overridden during a SQL Server upgrade.
+You can grant `EXECUTE` permissions on this procedure, but these permissions can be overridden during a SQL Server upgrade.
--- a/docs/includes/msdb-execute-permissions.md
+++ b/docs/includes/msdb-execute-permissions.md
@@ -6,4 +6,4 @@ ms.service: sql
 ms.topic: include
 ms.custom: include file
 ---
-This stored procedure is owned by the **db_owner** role. You can grant `EXECUTE` permissions for any user, but these permissions may be overridden during a SQL Server upgrade.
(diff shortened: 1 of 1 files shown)


Court72:
@rwestMSFT

#label:"aq-pr-triaged"

rwestMSFT:
Great feedback, thanks!

rwestMSFT:
Since this file is used in a lot of places, I'm checking your edit with the product group.

AndreasWolter:
Nothing can change the fact that all the agent procs are in the dbo schema which by design is owned by dbo. 
Also the new schemas managedbackup and smart_admin are owned by dbo.
This query will tell you that:
select p.name, p.principal_id as objectowner, s.principal_id as schemaowner, p.schema_id, s.name as schemaname from msdb.sys.procedures as p
inner join msdb.sys.schemas as s
on p.schema_id = s.schema_id

rwestMSFT:
Not arguing the technicality. I need to confirm with the PG that it's OK, because the edit affects a number of files. Thanks for your patience.

AndreasWolter:
sure thing. just giving your some more data

docs/includes/msdb-execute-permissions.md
```diff
@@ -6,4 +6,4 @@ ms.service: sql
 ms.topic: include
 ms.custom: include file
 ---
-This stored procedure is owned by the **db_owner** role. You can grant `EXECUTE` permissions for any user, but these permissions may be overridden during a SQL Server upgrade.
+You can grant the `EXECUTE` permissions on this procedure, but these permissions may be overridden during a SQL Server upgrade.
```
rwestMSFT:
```suggestion
You can grant `EXECUTE` permissions on this procedure, but these permissions can be overridden during a SQL Server upgrade.
```

rwestMSFT:
Owing to content freeze for Ignite, pulling this into an internal PR. Thanks again, Andreas.

PR #9931, opened 2024-11-13
Update sql-machine-learning-services-windows-install-sql-2022.md

joshts0:
Updating Python 3.10 link to link directly to the latest Python 3.10.11 release, instead of the python.org front page, as it's difficult for customers to find the correct downloadable and installable Python 3.10.


Jak-MS:
@VanMSFT 

#label:"aq-pr-triaged"

rwestMSFT:
@joshts0 Unfortunately we won't accept this PR, because it adds an additional maintenance burden for us to keep the URL up to date.

However, I do want to thank you for your contribution, and hope that you will do so again in the near future.

PR #9937, opened 2024-11-19
removing unrelated articles to avoid confusion

AndreasWolter:


Court72:
@rwestMSFT

#label:"aq-pr-triaged"

rwestMSFT:
@AndreasWolter Not sure of the motivation here. The links you removed are still related to certificates.

Also, our content guidelines have a requirement of 3 links in the related content. I know we don't always stick to that, but we can't remove these and not replace them.

AndreasWolter:
The 2 links refer to certificates within the SQL Engine, exposed in master and other databases.
This is not related to certificates that are used by the OS for encrypting the connections. Those you cannot grant permissions for using TSQL.  To avoid people mixing up topics I thought its better to delete these.

Here are some more useful references if you need to add something instead: https://learn.microsoft.com/en-us/sql/database-engine/configure-windows/certificate-requirements?view=sql-server-ver16 
https://learn.microsoft.com/en-us/sql/database-engine/configure-windows/certificate-overview?view=sql-server-ver16

rwestMSFT:
That first one in the related content is the same as your first example, so we'll just take the second one and add it.

AndreasWolter:
oups, pardon the confusion

AndreasWolter:
@rwestMSFT  I just found an interesting documentation that actually helps someone who wants to learn more about TLS in SQL Server :  https://learn.microsoft.com/en-us/windows-server/security/tls/tls-ssl-schannel-ssp-overview

PR #9947, opened 2024-11-25
Added VIEW SERVER STATE being included in SERVER_OPERATION_GROUP as this is a very noisy event

AndreasWolter:


rwestMSFT:
Thanks, Andreas. Another article in need of refresh!

PR #9948, opened 2024-11-26
Spelling Fix Wave 3

ArieHein:
Adding all findings across the repo that i managed to find, obviously there might be more but for this year, i think that  s plenty :)

@rwestMSFT I know its a lot of files, but look on the bright side, i have over 4k for the azure docs, and that's before ignite docs ;)


rwestMSFT:
Thankfully I don't work in the Azure Docs content area as much 😄.

Give this 2 - 3 weeks, and I'll take care of it.

rwestMSFT:
@ArieHein I'm back at work this week, so I'm looking into this now.

PR #9969, opened 2024-12-13
Update resource-limits.md

cbenkler:
Added clarification to the IO latency duration range to make clearer that IO requests exceeding the range are possible. (Although not as clear as I'd like - using "compromise language" based on comment from NNikolic intended to balance clarity with pragmatism).


Jak-MS:
@vladai78 

#label:"aq-pr-triaged"

kendra-little:
I don't think expecting storage performance for a database transaction log to respond in faster than 60 seconds is unreasonable. 
Sincerely,
Reasonable customer who can read these public comments

cbenkler:
Hi Kendra,

I can understand how you would think that I was referring to you given that you don't have context of all the conversations regarding this change. I was not referring to you nor was I referring to any customer in particular, but rather the general notion of "if we use language _____ that opens the door for an unreasonable customer..." which was a part of some of those discussions. Your support ticket was certainly a catalyst for the change, but once decided to make the change there must be very broad-scope consideration. It wasn't about you.

Regardless, thank you for raising - I agree that I should have chosen my words more carefully.

rwestMSFT:
This PR will be closed automatically when the internal PR is merged.

PR #9970, opened 2024-12-16

AV25242:
Changed the verbiage for inclusiveness


rwestMSFT:
Thank you for your contribution!

rwestMSFT:
@AV25242 Following up to say we took this opportunity to look for other articles with language that could be updated, so thanks again for your feedback.

PR #9976, opened 2024-12-25
Update vector-normalize-transact-sql.md

sommarskog:
I don't know too much about vectors, but the sentence as given does not make much sense. But I can see that if I use norm1, the sum of the absolute value of the elements is 1. If I use norm2, the sum of the squares is 1. And with norminf, the biggest absolute value is 1.

Someone who knows more about vectors maybe could elaborate more on the length. (Which I first read as "number of elements", which is the more computeresque interpretation of what is a mathematical term here.)

### DIFF
--- a/docs/t-sql/functions/vector-normalize-transact-sql.md
+++ b/docs/t-sql/functions/vector-normalize-transact-sql.md
@@ -66,7 +66,7 @@ A string with the name of the norm type to use to calculate the norm of the give
 
 ## Return value
 
-The result is a vector with the same direction as the input vector but with a length of 1.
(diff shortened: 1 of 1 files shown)


ShannonLeavitt:
@WilliamDAssafMSFT

#label:"aq-pr-triaged"

WilliamDAssafMSFT:
Thanks @sommarskog for taking the time to submit this PR! Good catch. The PG reviewed and agree. Your commit will be merged and this article update is in flight. After the updated article is published, GitHub will list you as an article contributor. 

Thanks again Erland for this feedback, your suggestions help improve our documentation.

#sign-off

PR #9986, opened 2025-01-12
Spelling Fixes

ArieHein:
Last wave of spell fixes.

@rwestMSFT thanks for adding all the previous ones, hope you had a good Christmas and new year break :)
I added it all in one PR like you requested for the first ones.

Cheers !


rwestMSFT:
Thanks again, @ArieHein. We'll do the same as before: close this PR and check every contribution independently. We may decide to not take specific corrections for various reasons. An example is where our metadata has `RAISEERROR_TSQL`, and you've corrected it to `RAISERROR_TSQL`. While the function is called `RAISERROR`, the metadata is used to help guide search engines, for customers who aren't aware of the correct spelling. Just one example, but it should give you a little insight into how we do things.

ArieHein:
@rwestMSFT, thats perfectly ok, Thanks !
Interesting comment to do with search engines.  Wouldn't that slightly influence AI training though ?
When spelling will be done by AI instead of humans, i wonder would it allow deliberate issues for its own training. Kinds of self "unit-test" :)

rwestMSFT:
> @rwestMSFT, thats perfectly ok, Thanks ! Interesting comment to do with search engines. Wouldn't that slightly influence AI training though?

No. In that file's metadata, we already had a correctly spelled `RAISERROR_TSQL` above the correction you made. The file is meant to have both.

rwestMSFT:
All approved edits will be live on Tuesday a short time after our morning publish time, which is around 10:30am Pacific.

PR #9992, opened 2025-01-22
move the block comment inside the stored proc.

codandn:
It is very tough to troubleshoot when the block comment is outside the stored procedure, especially when running it with DMV. We want to make sure that st.text gets the stored procedure name first before the comment block. If the comment is not placed inside, you have to expand the column to get to the stored procedure name, which complicates troubleshooting. Typically, you add a substring on st.text to get only a few characters of the text, and if the comment is placed before creating the stored procedure, it results in only showing the comments. Moving all comments inside the stored procedure ensures that the stored procedure name is retrieved first, making troubleshooting more efficient.

example:

```sql
SELECT SUBSTRING(st.text,1, 200)
FROM sys.dm_exec_query_stats wq
OUTER APPLY sys.dm_exec_query_plan(wq.plan_handle) AS qp
OUTER APPLY sys.dm_exec_sql_text(wq.plan_handle) AS st;
```

### DIFF
--- a/docs/relational-databases/stored-procedures/create-a-stored-procedure.md
+++ b/docs/relational-databases/stored-procedures/create-a-stored-procedure.md
@@ -3,7 +3,7 @@ title: Create a stored procedure
 description: Learn how to create a Transact-SQL stored procedure by using SQL Server Management Studio and by using the Transact-SQL CREATE PROCEDURE statement.
 author: WilliamDAssafMSFT
 ms.author: wiassaf
-ms.date: 01/25/2024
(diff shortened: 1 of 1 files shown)


v-dirichards:
@WilliamDAssafMSFT 

#label:"aq-pr-triaged"

WilliamDAssafMSFT:
@codandn Great feedback, nice pull request. Should be merged soon. When published, you'll receive contributor credit in GitHub and on the public docs site. Thanks for contributing to Docs. 

#sign-off

PR #9996, opened 2025-01-25
Update between-transact-sql.md

asos-martinsmith:
The statement that if any part is `NULL` the overall expression evaluates to `UNKNOWN` is not correct.

Counter example 

```
SELECT 1
WHERE NOT 2 BETWEEN NULL AND 1
```

Because `2 BETWEEN NULL AND 1` is equivalent to 

```
2  >= NULL  AND 2  <= 1
```

or 

```
UNKNOWN  AND FALSE
```

Which is `FALSE`

(so `TRUE` when negated with `NOT`)


Court72:
@rwestMSFT

#label:"aq-pr-triaged"

rwestMSFT:
Thanks, Martin. I will incorporate this change into an internal PR to refresh this article.

PR #10006, opened 2025-02-03
Update quickstart-install-connect-docker.md warning

LockTar:
The data is also removed when you just stop or restart the container. You don't have to remove the container.


LockTar:
Sorry this PR is not correct... My data is stored after a restart. I was confused with `tempdb` database...

PR #10013, opened 2025-02-12
Update quickstart-install-connect-suse.md

petersatsuse:
Adapt latest SLES versions and some remark to Azure Marketplace images


rwestMSFT:
@petersatsuse Thanks, Peter. Amit and I will pull these changes into an internal PR and work with you as appropriate.

PR #10029, opened 2025-02-28
Update base64-encode-transact-sql.md

sommarskog:
Corrected the section on Return types.


rwestMSFT:
Thanks Erland, we've got an internal PR that takes your contributions (this one and #10030) into account, that will publish in the next business day or two.

PR #10037, opened 2025-03-10
Clarity Counter Signatures ProcSelectT1

mattcargile:
Description text of `EXECUTE` permissions doesn't match the images.


rwestMSFT:
@mattcargile Looks like this article needs some closer attention. Thanks for letting us know. We'll do an internal PR to refresh it, and you'll still get contributor credit.

PR #10044, opened 2025-03-20
Fix PowerShell Boolean casting issue in Execute method

kaczynskim:
This PR corrects a misleading example in the PowerShell example for the PackageInfo.Execute method usage.

**Issue:**
In PowerShell, passing "false" (a string) instead of $false (a Boolean) incorrectly evaluates to $true.

**Fix:**
Updated documentation to clarify that false should be passed as $false (Boolean) instead of "false" (string).
Ensured that examples use the correct Boolean syntax to prevent execution mode misconfigurations.

**Impact:**
This fix ensures that users do not unintentionally enable 32-bit execution mode due to incorrect PowerShell type casting.

**Reference Issue:**
PowerShell automatically evaluates non-empty strings as $true, causing Execute("false", $envRef) to behave incorrectly.
Expected usage should be Execute($false, $envRef).

Please review and approve.


v-dirichards:
@chugugrace Can you review the proposed changes?
 
#label:"aq-pr-triaged"

rwestMSFT:
Thanks for your contribution. I'm merging this PR as it looks correct.

PR #10051, opened 2025-03-31
Update editions-and-components-of-sql-server-2022.md

moematsuo:
>2 Tuning enabled only on Standard edition features.

This statement has been present since SQL Server 2012 and seems to have stayed without any specific verification. I compared the DTA functionality between Enterprise and Standard editions and found no differences. 
I suggest removing this annotation as it causes confusion.


v-dirichards:
@rwestMSFT 

#label:"aq-pr-triaged"

rwestMSFT:
@moematsuo Thank you for your contribution. We're looking into this.

rwestMSFT:
@moematsuo I spoke with a PM in the product group who has noted that there is some subtlety you're not aware of, to do with distributed partitioned views that aren't supported in Standard edition.

We appreciate your contribution, but at this time, we will not be able to accept it. Thank you for taking the time.

#please-close

moematsuo:
Thanks for checking. So, can you add the detail about the distributed partitioned views? This is so vague.

PR #10058, opened 2025-04-04
Update sys-dm-exec-requests-transact-sql.md

jasonhorner:
Provided a more real-world example of showing active requests as this is the most likely use case for this dmv.


Court72:
@WilliamDAssafMSFT

#label:"aq-pr-triaged"

rwestMSFT:
Hi Jason, I like this example, but I'm going to use it as an additional example. I think there's still some value in the existing one.

PR #10067, opened 2025-04-09
Update query-store-hints.md

emad-almousa:
Hello,

I am recommending to add a warning note that if someone purges the query store data through "Purge Query Data" option all hints will be consequently removed. The reason I am suggesting this is that Query Store is regarded as data collector and with this new feature....its the first time we can use it for "**_setting_**" a configuration in a sense.  So, system admins will forget that their newly introduced hints will be removed. Not sure if Microsoft can protect these hints stored information in system table **sys.query_store_query_hints** from deletion when "Purge Query Data" is pressed.

Regards,
Emad

### DIFF
--- a/docs/relational-databases/performance/query-store-hints.md
+++ b/docs/relational-databases/performance/query-store-hints.md
@@ -197,6 +197,10 @@ Finally, remove the hint from `query_id` 39, using [sp_query_store_clear_hints](
 ```sql
 EXEC sys.sp_query_store_clear_hints @query_id = 39;
 ```
+Important
+
+Please note all Query Store data will be removed including "hints" information if Query Store data is reset through "Purge Query Data" option.
+
 
 ## Related content
 


emad-almousa:
Hello,

I am recommending to add a warning note that if someone purges the query store data through "Purge Query Data" option all hints will be consequently removed. The reason I am suggesting this is that Query Store is regarded as data collector and with this new feature....its the first time we can use it for "setting" a configuration in a sense. So, system admins will forget that their newly introduced hints will be removed. Not sure if Microsoft can protect these hints stored information in system table sys.query_store_query_hints from deletion when "Purge Query Data" is pressed.

Regards,
Emad

rwestMSFT:
@emad-almousa While I agree that an article discussing the purge feature in Query Store should have a warning about hints, this is not the correct place for it. I would suggest identifying the article that discusses purging data, and adding a note there about hints being removed.

Additionally, I remind you that we have a [contributor guide](https://learn.microsoft.com/en-us/contribute/content/how-to-write-overview), which contains the requirements for submitting PRs to SQL Docs. Given the number of contributions you've made in the last few years, I would hope you'd be more familiar with our Markdown standards, including the addition of info boxes in the correct format. Please review the contributor guide, and other examples of our content, to get a better idea of formatting and style.

#please-close

emad-almousa:
Hello @rwestMSFT  thanks for the feedback and nice words. I will do my best in the future (its not easy for me to be honest....i read the document you have shared). have a good day.

rwestMSFT:
@emad-almousa We always consider your GitHub contributions, and review them in the larger scope of tens of thousands of articles. That means we have to weigh all changes against the existing content. In this particular case, it's better suited to a different article. In other cases (like your contribution earlier today about security), it might be too specific and doesn't support the *intent* of the article. That being said, we do review everything, so please keep contributing.

PR #10069, opened 2025-04-13
Add impact of TEXTSIZE on sp_executesql and EXEC statement

rrg92:
Hi,

I recently ran into an issue related to this and had never noticed that SQL Agent changes the `TEXTSIZE` setting by default.  

I think it would be really helpful if the documentation explicitly mentioned this behavior—especially to improve indexing and discoverability for others facing similar issues (e.g., the query works fine in SSMS but fails when executed via SQL Agent).

In my case, the problem happened while using `EXEC` in combination with a linked server to MySQL, so I imagine many others around the world might encounter the same situation.

### DIFF
--- a/docs/t-sql/statements/set-textsize-transact-sql.md
+++ b/docs/t-sql/statements/set-textsize-transact-sql.md
@@ -50,6 +50,40 @@ SET TEXTSIZE { number }
  Setting SET TEXTSIZE affects the @@TEXTSIZE function.  
   
  The setting of set TEXTSIZE is set at execute or run time and not at parse time.  
+
+ ### Impacts on `sp_executesql` and `EXECUTE` Statements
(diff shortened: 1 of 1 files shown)


v-dirichards:
@WilliamDAssafMSFT 

#label:"aq-pr-triaged"

rwestMSFT:
#reassign:rwestMSFT

rwestMSFT:
The underlying article needs a refresh, so I'll look into this and incorporate your changes as needed. Thanks for your contribution. This PR will automatically close when the internal PR is merged, which may take a few days.

rwestMSFT:
@rrg92 FYI, we're condensing this and moving it to the Manage Job Steps article, in the SSMS content area. You'll retain contributor credit here (even though it'll end up being a link), but not there, due to how GitHub works. We expect the article to publish in the next couple of business days.

rrg92:
No problem @rwestMSFT! The most important thing is that the question about TEXTSIZE makes sense and is now being addressed officially in some documentation by you. I'm happy about that — thanks for the attention!

rwestMSFT:
Thank you for bringing this to our attention, and forcing me to review older content 😄

PR #10070, opened 2025-04-14
Update create-function-transact-sql.md

DanielAdeniji:
Immaterial Revision

Schema Binding Section

The binding of the function to the objects it references is removed only when one of the following actions occurs:

The binding of the function to the objects it references is removed only when ne of the following actions occurs:


rwestMSFT:
Thank you.

PR #10075, opened 2025-04-23
Update failover-cluster-instance-azure-shared-disks-manually-configur…

shigeooooooooooo:
Deploy a managed with the shared disk feature enabled
--->
Deploy a managed disk with the shared disk feature enabled

It seems "disk" is missing.


v-regandowner:
Looks good; thanks @shigeooooooooooo

PR #10081, opened 2025-05-04
Update set-trace-definition-defaults-sql-server-profiler.md

sqlworldwide:
Added space between words in four different places.


ttorble:
Thanks @sqlworldwide, looks good.

PR #10096, opened 2025-05-20
Denote that OPENJSON does not support json data type (yet)

LarnuUK:
The documentation incorrectly states that all JSON functions support the `json` data type in the **Function support** section, however, the **Limitations** section denotes that `OPENJSON` does *not*, which is true. Use of the `json` data type with `OPENJSON` returns an error:
> JSON data type cannot be used in OpenJson function.

This commit denotes that `OPENJSON` doesn't support the `json` data type in the **Function support** section. It should be updated when it does.

### DIFF
--- a/docs/t-sql/data-types/json-data-type.md
+++ b/docs/t-sql/data-types/json-data-type.md
@@ -91,7 +91,7 @@ WHERE id = 1;
 
 ## Function support
 
-All JSON functions support the **json** type with no code changes or usage difference necessary.
+All JSON functions support the **json** type with no code changes or usage difference necessary, with the exception **OPENJSON**; see [limitations](#limitations).
 
 For a complete list of JSON functions, see [JSON functions](../functions/json-functions-transact-sql.md).
 
--- a/docs/t-sql/data-types/json-data-type.md
+++ b/docs/t-sql/data-types/json-data-type.md
@@ -3,8 +3,8 @@ title: JSON data type (preview)
 description: The native JSON data type provides advantages for storing JSON data over varchar or nvarchar. Learn more about the JSON data type.
 author: WilliamDAssafMSFT
 ms.author: wiassaf
-ms.reviewer: randolphwest, jovanpop
-ms.date: 02/07/2025
+ms.reviewer: randolphwest, jovanpop, umajay
+ms.date: 05/21/2025
 ms.service: sql
 ms.topic: reference
 ms.custom:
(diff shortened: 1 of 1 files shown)


Court72:
@WilliamDAssafMSFT

#label:"aq-pr-triaged"

WilliamDAssafMSFT:
@LarnuUK thanks for the PR. We're investigating. What platform(s) are you finding this on?

LarnuUK:
> @LarnuUK thanks for the PR. We're investigating. What platform(s) are you finding this on?

Hi @WilliamDAssafMSFT ,

This is on Azure SQL Database. It can be repro'd with a script as simple as:
```sql
DECLARE @JSON json = N'{"Sample":"Json"}';
SELECT *
FROM OPENJSON(@JSON)
```
> Msg 13658, Level 16, State 1, Line 2  
> JSON data type cannot be used in OpenJson function. 

As stated in the PR, the documentation actually conflicts with itself. As it states:

[](https://learn.microsoft.com/en-us/sql/t-sql/data-types/json-data-type?view=azuresqldb-current&preserve-view=true#function-support)

> ## Function support
> 
> All [JSON functions](https://learn.microsoft.com/en-us/sql/t-sql/functions/json-functions-transact-sql?view=azuresqldb-current) support the **json** type with no code changes or usage difference necessary.
>
> For a complete list of JSON functions, see [JSON functions](https://learn.microsoft.com/en-us/sql/t-sql/functions/json-functions-transact-sql?view=azuresqldb-current).

But then, in the limitations states:
> ## Limitations
> - ...
> - Currently, the `OPENJSON()` function doesn't accept the **json** type, currently that is an implicit conversion. Explicitly convert to **nvarchar(max)** first.

`OPENJSON` is explicitly listed as a [JSON function](https://learn.microsoft.com/en-us/sql/t-sql/functions/json-functions-transact-sql?view=azuresqldb-current), so the `json` data type *can't* be supported by all the functions if `OPENJSON` doesn't support it, due to a documented limitation.

I *assume* that the plan is for `json` to be supported, but the support should denote that `OPENJSON` doesn't support it yet.

LarnuUK:
Alternatively, if `OPENJSON` should now support `json`, but doesn't (perhaps due to a regression?) then when/if `OPENJSON` is fixed, the note about `OPENJSON` not supporting `json` should be removed from the limitations.

Based on the fact that, at the moment, `OPENJSON` errors on a `json` data type (in Azure SQL Database), it appears the Limitations section is correct, and the Function support section not.

LarnuUK:
Interestingly, I gave the same query a go on SQL Server 2025 (CTP2.0) - 17.0.700.9, and it *works*, suggesting a regression in Azure SQL Database, and that the limitation should be removed (once it's fixed?).

WilliamDAssafMSFT:
@LarnuUK Nice catch. There's a note here to be added for SQL Server 2025. Currently, that is the only platform where this fix is in. Thanks for taking the time to submit this PR! Your commit will be merged, as well as mine, and this article update is in flight. After the updated article is published, GitHub will list you as an article contributor. 

#sign-off

WilliamDAssafMSFT:
#sign-off

PR #10102, opened 2025-05-27
Fix interchanged descriptions between Agent and Profile procedures

arthur-sm:
The descriptions for the agent and profile procedures were swapped

image


VanMSFT:
Thanks for catching this, @arthur-sm! I approve.

#sign-off

PR #10105, opened 2025-05-29
Fix the syntax of WITH clause

rrg92:
The previous syntax suggests that this works:

```sql
ALTER RESOURCE GOVERNOR RECONFIGURE WITH (CLASSIFIER_FUNCTION = NULL);
```

However, after running some tests I found that `ALTER` only supports the `WITH` clause when it is *not* preceded by `RECONFIGURE`. Therefore, that part should be moved so it is separate from the `RECONFIGURE` option.

Test on SQL 22 (also in sql 25 public preview):
image

This is the correct syntax:
image


dimitri-furman (approved):
Thanks for finding the error in the syntax. We appreciate your contribution!

v-regandowner:
@dimitri-furman - I'll merge this PR based on your approval. If this article requires further changes, please open a separate pull request.

PR #10106, opened 2025-05-30
Adds a section on unsupported keys to troubleshoot-tde.md

iamjustsam:
Adds a section on unsupported keys. Using an unsupported key will throw an InternalServerError without further information.

### DIFF
--- a/docs/relational-databases/security/encryption/troubleshoot-tde.md
+++ b/docs/relational-databases/security/encryption/troubleshoot-tde.md
@@ -185,6 +185,22 @@ Confirm that the server has permissions to the key vault and the correct permiss
 - If the server identity is present, ensure that it has the following key permissions: Get, WrapKey, and UnwrapKey.
 - If the server identity isn't present, add it by using the **Add New** button.
 
+### Wrong key format
+
+**Error message**
+
+_500 InternalServerError - An unexpected error occured while processing the request._
+
+**Detection**
+
+To identify the unsupported key in the key vault:
+
+- Use the Azure portal, go to the **Key vault** service menu > **Objects** > **Keys**, and check the key type and length.
+
+**Mitigation**
+
+Make sure that the key uses a [supported key length](/azure/azure-sql/database/transparent-data-encryption-byok-overview#requirements-for-configuring-tde-protector).
+
 ## Getting TDE status from the Activity log
 
 To allow for monitoring of the database status due to Azure Key Vault key access issues, the following events will be logged to the [Activity Log](/azure/service-health/alerts-activity-log-service-notifications) for the resource ID based on the Azure Resource Manager URL.


iamjustsam:

v-regandowner:
@rwestMSFT - Can you review the proposed changes? 
	
#label:"aq-pr-triaged"

VanMSFT:
Hi @Pietervanhove - Can you please review to see if this is an appropriate addition the doc? Thanks!

#assign:Pietervanhove

Pietervanhove:
It might be a good idea to add this, but I'm not sure if the error message "_500 InternalServerError - An unexpected error occured while processing the request._" is only related to a wrong key format. I'll need to double check with engineering.

v-dirichards:
> It might be a good idea to add this, but I'm not sure if the error message "_500 InternalServerError - An unexpected error occured while processing the request._" is only related to a wrong key format. I'll need to double check with engineering.

@Pietervanhove  - Would you add a comment to indicate what action to take on this PR?
Thanks!

Pietervanhove:
> > It might be a good idea to add this, but I'm not sure if the error message "_500 InternalServerError - An unexpected error occured while processing the request._" is only related to a wrong key format. I'll need to double check with engineering.
> 
> @Pietervanhove - Would you add a comment to indicate what action to take on this PR? Thanks!

I'm waiting for engineering on their feedback. Once I received their feedback, I can comment on what action we should take

Pietervanhove:
> Adds a section on unsupported keys. Using an unsupported key will throw an InternalServerError without further information.

Hi,

We have verified our code and as expected we do have an error message that captures the wrong key size. Can you please let me know what action you did and how we can reproduce this?

Thanks,
Pieter

iamjustsam:
Hi Pieter

You can reproduce this in either of the following ways:
- Deploy a key to an existing SQL Managed Instance using an ARM resource, referencing a key with an unsupported length
  ```json
  {
    "type": "Microsoft.Sql/managedInstances/keys",
    "apiVersion": "2024-05-01-preview",
    "name": "[concat(parameters('name'), '/', variables('encyptionKeyName'))]",
    "properties": {
      "serverKeyType": "AzureKeyVault",
      "uri": "[parameters('cmkKeyId')]"
    }
  }
  ```

  The ARM deployment will fail with the following error:
  ```json
  {
    "code": "InternalServerError",
    "message": "An unexpected error occured while processing the request. Tracking ID: '...'"
  }
  ```

- Deploy a key to an existing SQL Managed Instance using Azure CLI, referencing a key with an unsupported length
  `az sql mi key create -kid <KeyId> --managed-instance <ManagedInstance> --resource-group <ManagedInstanceResourceGroup>`
  The command will fail with the following output:
  ```
  (InternalServerError) An unexpected error occured while processing the request. Tracking ID: '...'
  Code: InternalServerError

  Message: An unexpected error occured while processing the request. Tracking ID: '...'
  ```

Pietervanhove:
Hi Sam,

We are aware of that issue on the MI side - we do not handle this error gracefully in Management Service, so it only appears as an internal server error. We created a repair item to address it. So, for now, I wouldn't add this to the documentation because the internal server error is not unique for the key size issue.

Regards,
Pieter

v-dirichards:
#please-close

> Hi Sam,
> 
> We are aware of that issue on the MI side - we do not handle this error gracefully in Management Service, so it only appears as an internal server error. We created a repair item to address it. So, for now, I wouldn't add this to the documentation because the internal server error is not unique for the key size issue.
> 
> Regards, Pieter

PR #10113, opened 2025-06-10
-f parameter

emad-almousa:
Hi,

**-f** parameter will disable all database engine audit rules configured, i think its worth stating it in the documentation.

thank you
emad


rwestMSFT:
Thank you for your contribution. The phrase "minimal configuration" already covers this scenario.

#please-close

emad-almousa:
Thanks for the feedback and usual cooperation, I honestly disagree. The term minimal configuration is broad and the security aspect should be clearly written to distinguish it from -m parameter where security audit will function (some might get confused that both startup parameters will disable audit).

I just blogged about these two startup parameters (-f,-m) and clearly certain different configurations are disabled: https://sqlserverwarrior.wordpress.com/2025/06/10/sql-server-startup-parameters-f-m-and-what-trace-flags-gets-enabled-by-default/

At the end it's your call of course.

Have a great day.

rwestMSFT:
We'll consider adding something to the auditing section of the content. Thanks again.

rwestMSFT:
Following up:

- This article is getting refreshed.

- There is already an article about minimal configuration, which specifies that Audit is disabled. I am adding a link to this article from here.

- The minimal configuration article is getting refreshed.

- The SQL Server Audit overview article is getting refreshed.

You will see these changes in the next day or so.

emad-almousa:
@rwestMSFT thank you

PR #10119, opened 2025-06-20
Clarified not to delete or disable a previous TDE protector key after a rotation.

sanjayd:
A user I was supporting in the past disabled their previous TDE Protector key in keyvault after rotating it. This caused their database to become unavailable following the rotation. Earlier in the doc it is advised that old backups and log files may still require access to the old key and to keep it active until sys.dm_db_log_info shows it's no longer in use, but this !important text box only mentions not to delete it. Adding a clarification to keep it active would help avoid similar situations.


v-regandowner:
@Pietervanhove - Can you review the proposed changes? 
	
#label:"aq-pr-triaged"

VanMSFT:
Hi @sanjayd! Thanks for submitting this change. We appreciate the feedback. However, keeping the key around is already implied in the article, and it should follow that the previous key is used to access the previous protected data. I don't think the extra verbiage is necessary. I will close this for now, but we'll be happy to accept more feedback.

#assign:VanMSFT

PR #10123, opened 2025-06-24
Update with-common-table-expression-transact-sql.md

erikdarlingdata:
@rwestMSFT might could use some tidying but I'm sure it's in good hands now.


rwestMSFT:
Thanks, Erik. This article needs some additional updates, so we'll incorporate these changes into an internal PR. You should see the updated article in the next day or so.

PR #10127, opened 2025-06-28
Update tempdb-space-resource-governance-walkthrough.md

emad-almousa:
Hi,

updating SSMS section as the provided description covers SSMS 20,19...but doesn't cover SSMS 21 as the interface has drastically changed in 21 release.

Thanks
Emad


dimitri-furman:
@emad-almousa thank you for your contribution! We do plan to make this change, however at present there is a [known issue](https://developercommunity.visualstudio.com/t/App-name-cannot-be-set-in-the-modern-con/10918029) in SSMS 21 that makes the new connection dialog not work for setting the app name. Once that is resolved, we'll update this article similarly to what you suggested.

We'll close this PR for now, but rest assured that this change is planned.

dimitri-furman:
#please-close

PR #10130, opened 2025-07-04
explicit nvarchar limitation in binary_checksum

sgjsy-axt:
binary_checksum is limited to 255 characters for any nvarchar type, not just nvarchar(max)
i have added "n | " in an effort to make this clearer in a concise way


rwestMSFT:
Thanks for the contribution. The underlying article needs refreshing, so we will incorporate your changes into an internal PR. When that PR is merged, this PR will automatically close.

v-dirichards:
#label:"aq-pr-triaged"

uc-msft (commented):
This looks good to me. And we can also change varbinary(max) to say the same since that also means you can use varbinary(n).

PR #10131, opened 2025-07-06
Update from-transact-sql.md to re-add incorrectly-removed OUTER keyword

binki:
Two versions of the same query are present to demonstrate that the `OUTER` keyword is optional in the `FULL OUTER JOIN` syntax. However, currently, the two versions of the same query are actually the identical query.

This change makes the two examples different again to be consistent with the point being demonstrated as inferred from the context.

Fixes a regression in 87245a27588bf78e636c2322f4b96c199db74760.


rwestMSFT:
This is what happens when you use a T-SQL formatter when refreshing articles. Thanks for the fix. This was entirely my fault 😄

PR #10140, opened 2025-07-16
Changed Non-cluster behavior when Clustered index is dropped.

ramanathnayak:
Currently the documentations says that Non Clustered Index remains disabled, when Clustered Index is dropped, however that is not what actually happens in the product. So have updated the documentation accordingly.


rwestMSFT:
We'll look into this. Thank you for your contribution.

#reassign:rwestMSFT

rwestMSFT:
Behaviour is confirmed. We are refreshing the underlying article in an internal PR. When that merges, this PR will automatically close and you can expect to see the updated article in the next day or so.

PR #10180, opened 2025-09-17
Remove duplicated entries in time zone mapping table - sql-server-linux-configure-time-zone.md

egbakou:
This PR removes redundant entries from the time zone mapping table in `sql-server-linux-configure-time-zone.md`.


rwestMSFT:
Thank you for your contribution. We'll address this in an internal PR.

PR #10187, opened 2025-10-01
Update create-vector-index-transact-sql.md

josh-simar:
Changed the example. The example text stated the usage of the "dot" product metric but the code was using the cosine metric.

Alternatively if you want you could change the example text but I don't think that was the intended so I changed the code.


yorek:
thanks for catching this :)

yorek:
@MikeRayMSFT LGTM

yorek (approved):

v-dirichards:
@yorek  I'll merge this PR based on your approval. If this article requires further changes, please open a separate pull request.

PR #10199, opened 2025-10-09
Update custom roles for VM migrations-2

abhims14:
Updated the date for the document and added a data action for blob read permissions.


rwestMSFT:
Please open your PR against the private repository.

PR #10219, opened 2025-11-14
Fix Syntax ORDER BY not required

way0utwest:
This isnt' required. Pinged Conor Cunningham as observed behavior was different than the doc


ktoliver:
It looks like this update is being pushed through a new PR in the private repo. We'll close this PR. Thanks.

PR #10220, opened 2025-11-14
Clarify SUM function syntax and arguments

way0utwest:
Updated syntax for the SUM function to include optional brackets for order_by_clause. Clarified the explanation of the partition_by_clause and order_by_clause.


ktoliver:
It looks like this update is being pushed in a new PR in the private repo. We'll close this PR. Thanks.

PR #10224, opened 2025-11-19
Update from-transact-sql.md

Bretttt:


rwestMSFT:
Good edit! Thanks @Bretttt.

PR #10229, opened 2025-11-21
Improving "BEGIN...END in T-SQL"

daiplusplus:
Previous versions of the SQL Server's documentation for `BEGIN...END` featured more remarks which are lacking from the current version which I felt were helpful to mention; I also added new remarks about some common gotchas I've encountered (such as an example workaround for the prohibition on empty blocks, and other clarifying remarks.

I also replaced the main T-SQL example because I felt the use of `IF @TRANCOUNT...` in an article about `BEGIN...END` _without_ an associated control-flow statement was inappropriate, while the remarks about `ROLLBACK` were irrelevant.

### DIFF
--- a/docs/t-sql/language-elements/begin-end-transact-sql.md
+++ b/docs/t-sql/language-elements/begin-end-transact-sql.md
@@ -28,17 +28,24 @@ monikerRange: ">=aps-pdw-2016 || =azuresqldb-current || =azure-sqldw-latest || >
 
 [!INCLUDE [sql-asdb-asdbmi-asa-pdw-fabricse-fabricdw-fabricsqldb](../../includes/applies-to-version/sql-asdb-asdbmi-asa-pdw-fabricse-fabricdw-fabricsqldb.md)]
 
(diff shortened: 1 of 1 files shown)


daiplusplus:
I'm unsure how to fix the bookmark-not-found error because [it's a `#:~:text=` fragment link](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment/Text_fragments), rather than an element identifier.

Incidentally, [the link to the error's documentation](https://learn.microsoft.com/help/platform/validation-ref/bookmark-not-found) gives me a HTTP 404 error.

rwestMSFT:
#review-pr

rwestMSFT:
@daiplusplus Thank you for this effort. We'll incorporate this contribution into an internal PR, and you will retain credit.

daiplusplus:
BTW, @rwestMSFT - while researching older docs for this article I waded through the SQL Server 7.0, 2000, and 2005 documentation I saw references to claimed support for `BEGIN/END` within a `CASE` expression - though this wasn't corroborated by anything else; I posted it to Dba.StackExchange here: https://dba.stackexchange.com/questions/349284/could-begin-end-really-be-used-inside-case-in-sql-server-2005#comment678645_349284 

I don't suppose you could weigh-in on this - at least to confirm it was a mistake, or if this was something that was actually supported in some ancient version?

Thanks!

PR #10233, opened 2025-11-28
sys.columns: Corrected data types on vector_base_type*

dhmacher:


dhmacher:
There are a couple more DMVs with vector columns. Will create a new PR.

PR #10243, opened 2025-12-07
Update sp-help-spatial-geography-histogram-transact-sql.md

sommarskog:
Corrected the required permission based on the actual behaviour, although this has to be construed as a bug. That is, you don't get an explicit permission error, but only "A severe error occurred on the current command. The results if any should be discarded". Given how long this behaviour has been there, fixing the actual bug is unlikely to pass the triage bar. Thus, better to adopt the documentation.


rwestMSFT:
Replaced by https://github.com/MicrosoftDocs/sql-docs/pull/10244.

PR #10245, opened 2025-12-08
docs: Add updated docs for BJS in Arc SQL Extension

mdrakiburrahman:
Adds some FAQs from @hdharshana into the agent docs.

### DIFF
--- a/docs/sql-server/azure-arc/agent-extension-files.md
+++ b/docs/sql-server/azure-arc/agent-extension-files.md
@@ -74,6 +74,33 @@ These tables store background job definition and execution history. Background j
 
 Additionally, the table `dbo.SQLServerAzureArcProperties` contains the resource identity for the SQL Server instance in Azure Resource Manager.  This table can be used to detect if the SQL Server instance is Arc-enabled, and if so, what the identity of the resource is in Azure.
 
+### FAQs
+
+1. Where are these background jobs?
+   
+   The background jobs are used to perform long-running tasks that can persist state in case of computer restarts. The logic of the jobs are in the extension, while state state is stored in `msdb`. For example, a Migration Assessment Job can take a long time to execute, the state is stored in `msdb`.
+   
+2. What security context do they use to execute?
+
+   They are run in the Service `C:\Program Files\Sql Server Extension\SqlServerExtension.Service.exe`. The service connects to SQL Server `msdb` as the low privelge `NT Service\SQLServerExtension` user. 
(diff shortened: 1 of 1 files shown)


docs/sql-server/azure-arc/agent-extension-files.md
```diff
@@ -74,6 +74,33 @@ These tables store background job definition and execution history. Background j
...
 
+### FAQs
+
+1. Where are these background jobs?
+   
+   The background jobs are used to perform long-running tasks that can persist state in case of computer restarts. The logic of the jobs are in the extension, while state state is stored in `msdb`. For example, a Migration Assessment Job can take a long time to execute, the state is stored in `msdb`.
+   
+2. What security context do they use to execute?
+
+   They are run in the Service `C:\Program Files\Sql Server Extension\SqlServerExtension.Service.exe`. The service connects to SQL Server `msdb` as the low privelge `NT Service\SQLServerExtension` user.
```
sfeitler, line 85:
```suggestion
   They are run in the Service `C:\Program Files\Sql Server Extension\SqlServerExtension.Service.exe`. The service connects to SQL Server `msdb` as the low privilege `NT Service\SQLServerExtension` user. 
```

sfeitler, line 85: Aside from the typo... :)

The user depends on whether the extension is installed in Least Privilege mode or not. In non-LP mode, we are connecting as the LocalSystem user, which is generally not a low-privilege user.

We call out these account differences in other docs. I don't know what wording we would want to use here.

hdharshana: Hi @mdrakiburrahman - thank you so much for the fast turnaround on this , much appreciated! Sorry - I missed a few questions on the email, that were on the ADO. The addnl three questions are below:

- Which features use these tables and jobs and why? 
- How to deal with scenarios where these tables are dropped by DBA inadvertently? Or corruption occurs in these tables? etc.
- How to create these tables manually if they are missing?

Thanks so much for your help with this!

sfeitler: @hdharshana any thoughts on wording? In essence, the user is the same user we use for all SQL Server access. So, in non-least-privilege, that is "AUTHORITY\SYSTEM", in LP it is as Raki described.

rwestMSFT:
Please move this PR to `sql-docs-pr`.

PR #10251, opened 2025-12-15
Fix system message schema URLs

nivinj:
Schema URLs are http, not https. The URL is correct in the XML snippet, but incorrectly using https in the document text.


rwestMSFT:
Thank you for your contribution!

PR #10265, opened 2026-01-14
Add missing limitation information for the JSON data type

jing8956:
```tsql
DECLARE @true JSON = 'true'; -- invalid
DECLARE @false JSON = 'false'; -- invalid 
DECLARE @number JSON = '1234.56'; -- invalid 
DECLARE @string JSON = '"foo"'; -- invalid 
DECLARE @null JSON = 'null' -- invalid

DECLARE @object JSON = '{}' -- valid 
DECLARE @array JSON = '[]' -- valid
```

### DIFF
--- a/docs/t-sql/data-types/json-data-type.md
+++ b/docs/t-sql/data-types/json-data-type.md
@@ -147,6 +147,8 @@ Currently, the `OPENJSON()` function doesn't accept the **json** data type in so
 
 - In [!INCLUDE [sssql25-md](../../includes/sssql25-md.md)], the `OPENJSON()` function does support **json**. For more information, see [Key JSON capabilities in SQL Server 2025](../../relational-databases/json/json-data-sql-server.md#key-json-capabilities).
 
+The top-level type must be a JSON Object or a JSON Array. Scalars, booleans, and null values are not supported.
+
 ### Size limitations
 
 Size limits of JSON data in storage:


WilliamDAssafMSFT:
@jing8956 Thanks for creating a pull request on Microsoft documentation. This appears to be pretty straightforward, can you explain further why this distinction would be necessary about the **json** data type? Also, `top-level data type` is not a Microsoft docs concept, can you explain further what you mean by "top-level"?

Court72:
@WilliamDAssafMSFT
	
#label:"aq-pr-triaged"

jing8956:
@WilliamDAssafMSFT Sorry for being pretty straightforward.
I can't speak English, I can only communicate with the help of a translator. Therefore, I try to avoid writing long texts.

According to the definition in [RFC 8259](https://www.rfc-editor.org/rfc/rfc8259#section-13), Text that contains only a value is also a valid JSON.

This is very useful for dynamic forms; some forms store complex objects, while others only need to store a single scalar.

Ever since MS-SQL introduced a dedicated JSON data type, I have always thought it would be a new solution to completely replace the previous use of varchar or nvarchar for storing JSON data. That was until today, when I failed to store a JSON boolean value in a column with the JSON data type.

Based on my actual testing on SQL Server 2025 and Azure SQL, the JSON data type is only a subset of the JSON.

Regarding the explanation of 'top-level,', [ISJSON (Transact-SQL)](https://learn.microsoft.com/sql/t-sql/functions/isjson-transact-sql) has a longer explanation:
> IETF RFC 8259 conformant JSON document that contains only a JSON scalar value at top level.

If you feel that a longer explanation is needed to clarify this, please feel free to modify my pull request.

WilliamDAssafMSFT:
Thanks @Jing8956 this is a good contribution to the doc. Will bring this commit internally and add to it. This doc flight is in progress.

PR #10283, opened 2026-02-12
Enhance sqlpackage-extract.md with PowerShell example

jasonhorner:
The example provided won't work as is because it depends on the $Credential variable which isn't set. this example should work


dzsquared (approved):

ttorble:
@dzsquared I'll merge this PR based on your approval. If this article requires further changes, please open a separate pull request.

PR #10287, opened 2026-02-21
READPAST cannot be used in snapshot-isolation even with other hints

runekock:
I get an error if I try WITH(UPDLOCK, READPAST) in snapshot isolation. Remove sentence that indicates that extra hints are sufficient to make it work.


ttorble:
@VanMSFT 

#label:"aq-pr-triaged" 

VanMSFT:
Thanks for catching this, @runekock — you're absolutely right.

I verified the behavior by running four tests against SQL Server with `ALLOW_SNAPSHOT_ISOLATION ON`:

| Test | Hint Combination | Isolation Level | Result |
|------|-----------------|-----------------|--------|
| 1 | `READPAST` | SNAPSHOT | ❌ Error 650 |
| 2 | `UPDLOCK, READPAST` | SNAPSHOT | ❌ Error 650 |
| 3 | `HOLDLOCK, READPAST` | SNAPSHOT | ❌ Error 650 |
| 4 | `READPAST` (control) | READ COMMITTED | ✅ Success |

All three SNAPSHOT tests produce the same **Error 650**: *"You can only specify the READPAST lock in the READ COMMITTED or REPEATABLE READ isolation levels."*

The engine rejects `READPAST` before it even evaluates companion hints like `UPDLOCK` or `HOLDLOCK`. The second sentence contradicts the first and describes behavior that SQL Server doesn't implement. This makes sense semantically too — `SNAPSHOT` isolation uses row versioning rather than row locks, so there are no locked rows for `READPAST` to skip.

Your fix is correct. Approving.

#sign-off

PR #10299, opened 2026-03-13
Fix date format pattern in format-transact-sql.md

leighghunt:
Corrected the date format pattern in the documentation.

The existing format would result in "March DD, 2026 (Friday)". Updated so results in "March 13, 2026 (Friday)"


rwestMSFT:
Good catch! Thanks, @leighghunt

PR #10333, opened 2026-05-18
Enhance Hyperscale conversion documentation

ericmattingly:
Added details about manual cutover and synchronous copy during database conversion to Hyperscale. This behavior was unexpected and caused an outage for us so want to make sure it's well documented.


WilliamDAssafMSFT:
@ericmattingly Will pull this internal.

PR #10349, opened 2026-06-15
Fix formatting and terminology in licensing documentation

allanhirt:
Fixed HA terms to make them accurate


allanhirt:
Need to do this in the -pr Branch.

PR #10364, opened 2026-07-18
Update set-arithabort-transact-sql.md

sommarskog:
A thorough review of the topic:
- Removed "Always set ARITHABORT to ON in your logon sessions. Setting ARITHABORT to OFF can negatively impact query optimization, leading to performance issues." which is bad and incorrect advice.
- Clearly stating that as long as ANSI_WARNINGS is ON, this setting has no effect. 
- Removed all references to compat level 80, since it is only supported in SQL 2008 and earlier.
- Removed text and passages that seemed to just rehash what had been said already.
- Changed "those errors" and similar to "arithmetic errors".
- Reworked the example so that it actually shows something interesting.


v-regandowner:
@WilliamDAssafMSFT 

#label:"aq-pr-triaged"

WilliamDAssafMSFT:
@sommarskog Hi Erland, thanks for the careful read and the PR. I've reviewed with the PG, should be close to ready. Review in flight. Thanks again for this feedback. Your suggestions help improve our documentation.

WilliamDAssafMSFT:
In flight with edits.

PR #10391, opened 2026-09-11
Update SUM function syntax and OVER clause details

donican:
This article doesn't mention that the OVER clause can include the ROWS or RANGE clause. I believe this is particularly important because when the ORDER BY clause is specified, ROWS or RANGE assumes its default value, which can then be explicitly changed by the user. My suggested change clarifies this point.


v-regandowner:
@rwestMSFT - Can you review the proposed changes? 

#label:"aq-pr-triaged"

